PR #286, opened 2014-08-19
RFC: Support checksum offload

talex5:
This pull request is just to track this feature. Possible improvements include:
- Replace `needs_checksum` with a more generic `flags` argument.
- Consider modifying `writev` rather than adding a new function (makes us incompatible with V1 though).
- Provide a way to tell users whether offload is supported.

### DIFF
--- a/types/V2.mli
+++ b/types/V2.mli
@@ -292,6 +292,9 @@ module type NETWORK = sig
   (** [writev nf bufs] output a list of buffers to netfront [nf] as a
       single packet. *)
 
+  val writev2 : needs_checksum:bool -> t -> buffer list -> unit io
+  (** Extends [writev] with the ability to request TCP checksum offload. *)
+
   val listen : t -> (buffer -> unit io) -> unit io
   (** [listen nf fn] is a blocking operation that calls [fn buf] with
       every packet that is read from the interface.  It returns as soon
@@ -349,6 +352,9 @@ module type ETHIF = sig
   val writev : t -> buffer list -> unit io
   (** [writev nf bufs] output a list of buffers to netfront [nf] as a
       single packet. *)
+
+  val writev2 : ?needs_checksum:bool -> t -> buffer list -> unit io
+  (** Extends [writev] with the ability to request TCP checksum offload. *)
 
   val mac : t -> macaddr
   (** [mac nf] is the MAC address of [nf]. *)


samoht:
could we have a better name that `write2`? :-)

dbuenzli:
> Consider modifying writev rather than adding a new function

That seems a better option no ?

samoht:
yes, modifying is better

samoht:
ping ?

talex5:
If we're breaking the API, I'd also like to add some way to provide back-pressure. Currently, listen spawns a new async thread for each frame received. I think it should wait for these in series (the caller does an async call once they're ready for the next frame, if needed). That would also ensure frames got processed in the order they arrived, for cases where that's useful.

In general though making API changes in Mirage is far more painful than it should be, since we need new releases of `mirage-types`, `mirage-net-*`, `tcpip` and `mirage` all at once.

I think @djs55 suggested something like replacing `V1.NETWORK` with `Mirage_net.V1`. Then I could add a `V2` with the new interface, then make `mirage-net-xen` provide an implementation of `V2` alongside `V1`. That would allow releasing `mirage-types` and `mirage-net-*` without waiting for anything else.

In wonder if we could get the Mirage tool to pick the right versions automatically? e.g. `mirage-net-xen` can provide `V1` and `V2`, `mirage-net-unix` only provides `V1`, and `tcpip` can accept `V1` or `V2`. Should be simple to solve for... and even if it isn't, we have a SAT solver...

hannesm:
I'm strongly against versioning in module names.  Can we leave versioning to packages, please?  It is not too bad: improve mirage-types (by making a backwards-incompatible change), release a new mirage-types, mark conflicts in opam packages, adjust your favorite package to the new mirage-types version, release, rinse, repeat.  We actually have the mirage-dev repository to stage releases...

Versions in module names would only be needed if we want to use multiple signature versions at the same time in an application, and I've the feeling this is a bad idea.  In case someone wants to achieve a stable branch, it is easy as well: branch off mirage-types in your favourite stable version, make another minor bugfix release, and branch, improve, release dependencies.

I'd expect for an upcoming (3.0?) release to rename `V1` into `M` (for `MirageOS`), and `V1_LWT` into `M_LWT`, or use `S` instead of `M` if you don't want to confuse all the people (or use `MIRAGE`) :)

yomimono:
@talex5, can you refresh me about the status of the implementation of checksum offloading?  I think we've come to a bit of community consensus around how to handle some of the API questions raised above, but I seem to have lost the thread on the implementation work that I thought I remembered you having done.

talex5:
Well, the previous comment starts "I'm strongly against ..." so I don't think there is consensus. The actual implementation is fairly trivial (just passing flags through to Xen and back out again).

The flags are defined here:

https://github.com/mirage/mirage-net-xen/blob/master/lib/flags.mli

yomimono:
Yes, much of the discussion did occur in other places (as I'd argue that it's rather off-topic here): http://canopy.mirage.io/irclogs/29-06-2016 for example, previous IRC logs, and in the megathread at https://lists.xenproject.org/archives/html/mirageos-devel/2016-05/msg00019.html .

yomimono:
A fuller discussion of this issue should incorporate the `solo5` backends.  I'd like to close this PR in favor of a fuller discussion on the right implementation interface for this post-mirage3.

avsm:
Agreed; I would also like to have per-interface flags rather than per-function flags if possible, for this sort of functionality.  Also we need to make sure we support the full gamut of offload (e.g. TSO) as well as just checksum offload.

yomimono:
> I would also like to have per-interface flags rather than per-function flags if possible

This could conceivably be part of the `connect` function then, which isn't constrained by the overall `NETWORK` module type, and wouldn't necessarily be API-breaking anyway.

talex5:
But you need to set different flags for each frame. e.g. you can't set TCP offload for ARP messages (Linux dom0 will silently drop them!).

yomimono:
> But you need to set different flags for each frame. e.g. you can't set TCP offload for ARP messages (Linux dom0 will silently drop them!).

...if you need me, I'll be over here being horrified.

talex5:
Yes, well, I did waste a few hours recompiling Linux kernels to find that out...

But even ignoring that, the flags do need to be per frame. It's complicated because Xen wants to avoid calculating or verifying checksums when packets go directly between VMs on the same host.

So, packets have a lifecycle that goes something like this:

1. Data valid, checksum partial. This is the state for a newly generated packet. You can consume it, but must not try to validate it. VMs on the same host can exchange these packets and communicate without ever doing any checksum work.

2. Data valid, checksum complete. The state of a packet about to be sent over an unreliable network, or one which has been received and checked. You can consume it without validating it.

3. Data unknown, checksum complete. The packet has arrived over an unreliable connection and hasn't yet been checked. You must validate it before consuming it.

If you get the state mixed up it will never work because trying to complete an already-complete checksum will corrupt it.

Because mirage-firewall can't see the state, it always assumes the data is valid and overwrites the checksum with a complete fresh one, which isn't ideal. See https://github.com/talex5/qubes-mirage-firewall/issues/1

balrajsingh:
A related thread from a while ago:
https://lists.xen.org/archives/html/xen-devel/2013-12/msg00884.html

talex5:
Closing as long-obsolete.

PR #341, opened 2014-12-19
Don't reverse order of C libraries

talex5:
Looks like they were sorted correctly in the first place. Needed to link zarith, whose C stubs depend on those from gmp.


avsm:
This seems to work in the release 2.2.0 that didn't have this change though...

talex5:
What did it work with? It's only a problem when there's a dependency between two libraries using `xen_linkopts` (e.g. `gmp` and `zarith`). Otherwise, the order doesn't matter.

avsm:
Woops yes -- so this is indeed necessary to compile the full TLS/Xen server.

PR #346, opened 2015-01-13
Added error_message function to FLOW

talex5:
This allows for generic handling of errors without losing detail. For example, a user of a FLOW can use this to log errors. Before, the user would need to know the specific implementation of FLOW
being used.

This is required for TLS support in Conduit. It requires patches to console, vchan and tcpip to be applied first to add the extra method.

Longer term, we may want to add this function to `DEVICE` too.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -148,6 +148,9 @@ module type FLOW = sig
 
   type error
 
+  val error_message : error -> string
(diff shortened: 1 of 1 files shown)


djs55:
I think this is reasonable. We probably shouldn't release it until the implementations of the FLOW signature have this function added.

PR #347, opened 2015-01-14
[RFC] Add TLS support in Conduit

talex5:
Notes:
* Entropy is now initialised automatically, since we don't want to expose the application to this. Is this the correct way to do it?
* Conduit now *requires* a TLS implementation. Should this be optional?
* You can't configure an https web server directly because there's no way to pass in the configuration. Instead, your unikernel needs to take a conduit argument, load the certificate itself, and create the web-server from that.
* This depends on https://github.com/mirage/ocaml-conduit/pull/37

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,5 @@
+* Conduit now takes an optional TLS argument, allowing servers to support encryption.
+
 2.3.0 (2015-03-10):
 * Remove the `IO_PAGE` module type from `V1`. This has now moved into the
   `io-page` pacakge (#356)
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -447,9 +447,9 @@ module Entropy = struct
   let name _ =
     "entropy"
 
-  let module_name _ = "Entropy"
+  let module_name () = "Entropy"
 
-  let construction _ =
+  let construction () =
     match !mode with
     | `Unix | `MacOSX -> "Entropy_unix"
     | `Xen  -> "Entropy_xen.Make(OS.Time)"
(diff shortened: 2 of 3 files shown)


djs55:
Re: initialising entropy: it's not totally clear (to me) what we should do. The current 'entropy' source on Xen is just "Random.self_init ();" (!) which seems silently broken. Someone who really cares about their crypto keys is probably going to wait to insist the entropy comes from somewhere reasonable (e.g. [mirage/mirage-entropy#8] or [RDSEED](https://software.intel.com/en-us/blogs/2012/11/17/the-difference-between-rdrand-and-rdseed)). So although I hate exposing complexity to application (authors) it might be justified in this case.

talex5:
The actual entropy source used is still specified by the user in `config.ml` (typically `default_entropy`), but it is not passed to the application (unless it specifically asks for it). 
This isn't about what the entropy source is, only about when (and how) we initialise that source. I want to do it in `main.ml` because an application that asks for an abstract high level object (e.g. a conduit) shouldn't need to know that on Xen it has to initialise the entropy.

samoht:
I'm not very happy with the current design pattern for adding new stuff to conduit: this doesn't scale and it is not user friendly (you don't really want to change your library every time someone adds a new conduit implementation).

Anyway, this is not really the point of this PR, so I'm fine merging as is. We should definitely modernise conduit next ...

avsm:
Yes, dynamic registration of conduit modes was always the longer-term plan.  I think that the move to `DYNAMIC_FLOW` is right on target for making this happen, so this changeset is good.

samoht:
This needs unreleased packages, see https://github.com/mirage/mirage-dev/pull/52

PR #350, opened 2015-01-26
[Tracking] Remove DEVICE.connect

talex5:
When a module is functorised over a DEVICE it should only have the
ability to *use* devices it is given, not to connect to new ones.

This change also lets us get rid of the 'id' function from DEVICE.

The 'id' and 'error' types are still present to avoid breaking
existing code that uses 'with id = ...', but can be removed later.

This depends on some other patches being applied first:

- [x] https://github.com/mirage/ocaml-mbr/pull/7
- [x] https://github.com/mirage/mirage-console/pull/34
- [x] https://github.com/mirage/mirage-net-unix/pull/10
- [x] https://github.com/mirage/mirage-net-xen/pull/19
- [x] https://github.com/mirage/mirage-block-xen/pull/35
- [x] https://github.com/mirage/mirage-block-unix/pull/20
- [x] https://github.com/djs55/ocaml-vhd/pull/27
- [x] https://github.com/mirage/xen-disk/pull/18
- [x] https://github.com/mirage/mirage-tcpip/pull/100
- [x] https://github.com/mirage/mirage/pull/351
- [x] https://github.com/mirage/mirage-net-macosx/pull/1
- [x] https://github.com/mirage/mirage-entropy/pull/13
- [x] https://github.com/mirage/ocaml-crunch/pull/13
- [x] https://github.com/mirage/ocaml-fat/pull/39
- [x] https://github.com/mirage/mirage-fs-unix/pull/8

And probably some other things...

PR #351, opened 2015-01-26
Fix STACKV4 code gen for DEVICE sig w/o 'connect'

nojb:
See #350 and mirage/mirage-tcpip#100.


talex5:
I think this needs to do something similar for `STACKV4_socket`.

nojb:
Done.  Tested with `mirage-skeleton/stackv4` in socket mode and it worked fine.

PR #353, opened 2015-01-27
Test the opam version

samoht:
Just a proof of concept, not tested at all so might not be working as expected.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -2332,12 +2332,23 @@ let configure_makefile t =
 let clean_makefile t =
   remove (t.root / "Makefile")
 
+let no_opam_version_check_ = ref false
+let no_opam_version_check b = no_opam_version_check_ := b
+
 let configure_opam t =
   info "Installing OPAM packages.";
   match packages t with
   | [] -> ()
   | ps ->
-    if command_exists "opam" then opam "install" ps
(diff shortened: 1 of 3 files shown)


samoht:
I've updated the patch to add `--bypass-opam-version-check` as discussed on one the last mirage calls.

samoht:
Note: this patch will allow us to drop support for opam 1.1 in mirage.

samoht:
Pushed a new version with a simpler command-line name.

lib/mirage.ml
```diff
@@ -2332,12 +2332,23 @@ let configure_makefile t =
...
   | [] -> ()
   | ps ->
-    if command_exists "opam" then opam "install" ps
+    if command_exists "opam" then
+      if !no_opam_version_check_ then ()
+      else (
+        let opam_version = read_command "opam --version" in
+        match split opam_version '.' with
+        | "1"::"2"::_ -> opam "install" ps
+        | _ -> error "Your version of opam: %s is not up-to-date. \
```
talex5: Err... isn't this also going to reject opam 1.3 and all later versions?

samoht: indeed. Not vey urgent because opam 1.3 is not ready yet but I'll fix that

PR #355, opened 2015-01-29
Fix missing log error

talex5:
When not redirecting output, don't try to read the log file on error.
Error was:

    mirage: internal error, uncaught exception:
    	Sys_error("log: No such file or directory")

(is-mirage-broken is suffering from this: https://github.com/mirage/is-mirage-broken/blob/2cc7c6f691f84fb47144d5d049d2d1ace5c38e41/logs/mirage-www-ubuntu-14.04-4.01.0)


avsm:
that's been bugging me for ages...

PR #356, opened 2015-01-29
[RFC] Remove unused IO_PAGE module type

### LINKED ISSUE #343: IO_PAGE definition is duplicated and inconsistent
`types/V1.mli` contains an `IO_PAGE` signature that is similar to the type of Io_page, but not quite the same. e.g.

`io_page.mli`:

    val get : int -> t
    (** [get n] allocates and returns a memory block of [n] pages. If
        there is not enough memory, an [Out_of_memory] exception is
        raised.  Note that this may be a recoverable situation, since
        this function allocates memory from a page-aligned pool, whereas
        the OCaml garbage collector will be running in its own heap that
        may have spare memory despite the [Out_of_memory] being raised
        from this function call. *)

but `V1.mli`:

    module type IO_PAGE = sig
      val get : int -> t
      (** [get n] allocates and returns a memory block of [n] pages. If
          there is not enough memory, the unikernel will terminate. *)

As far as I can tell, no one is using `IO_PAGE` except for a unit-test in `Io_page` that checks that Io_page implements the type. Should it be fixed or deleted?

I'd like to clean up the types a bit. e.g. make `t` private to avoid treating all buffers as IO pages, and perhaps separate out single pages from ranges of pages, since some APIs care (e.g. `Netif`). That would be easier if the definitions weren't duplicated (or, actually, triplicated, as `io_page.mli` declares several functions twice).

talex5:
Closes #343.

Note: io-page 1.0.0 uses this for its unit-tests and may need a constraint added. 1.3.0 and later don't need it. Don't know what 1.1 and 1.2 are doing.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -806,52 +806,6 @@ module type FS = sig
 
 end
 
-module type IO_PAGE = sig
-  (** Memory allocation interface. *)
-
-  type buf
-  (** Type of a C buffer (usually Cstruct) *)
-
-  type t = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
-  (** Type of memory blocks. *)
-
-  val get : int -> t
-  (** [get n] allocates and returns a memory block of [n] pages. If
-      there is not enough memory, the unikernel will terminate. *)
-
-  val get_order : int -> t
-  (** [get_order i] is [get (1 lsl i)]. *)
-
-  val pages : int -> t list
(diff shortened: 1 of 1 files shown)


talex5:
Hmm, so `fat-filesystem` uses `IO_PAGE` too.

If we want to keep this, I suggest moving it into io-page, since it isn't mirage-specific. That would still avoid the need to duplicate it.

avsm:
Moving it into `Io_page.S` sounds fine to me.

PR #359, opened 2015-01-31
Support the 2 Xen entropy sources

djs55:
In a `config.ml` one can use

- `default_entropy`: this is documented as "Pick the strongest entropy source available.". On Xen the only strong source we currently have is via `xentropyd`, although this could change with [mirage/mirage-entropy#9]
- `weak_entropy`: using this means that we will accept bad entropy, typically for dev/test/experimentation

Note that the Unix entropy source has been defunctorised while the Xen source needs to be functorised.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -442,19 +442,27 @@ let default_random: random impl =
 
 module Entropy = struct
 
-  type t = unit
+  type t = [ `Weak | `FromHost ]
 
   let name _ =
     "entropy"
 
-  let module_name () = "Entropy"
+  let module_name _ = "Entropy"
 
-  let construction () =
+  let construction _ =
     match !mode with
     | `Unix | `MacOSX -> "Entropy_unix.Make (OS.Time)"
-    | `Xen  -> "Entropy_xen"
+    | `Xen  -> "Entropy_xen.Make(OS.Time)"
 
-  let packages () =
+  let id t =
+    match !mode, t with
+    (* default on Unix is the system entropy source *)
+    | (`Unix | `MacOSX), _ -> "()"
+    (* Xen has multiple methods, with different system requirements *)
+    | `Xen, `Weak -> "`Weak"
+    | `Xen, `FromHost -> "`FromHost"
+
+  let packages _ =
     match !mode with
     | `Unix | `MacOSX -> [ "mirage-entropy-unix" ]
     | `Xen  -> [ "mirage-entropy-xen" ]
@@ -465,10 +473,10 @@ module Entropy = struct
     append_main "module %s = %s" (module_name t) (construction t) ;
     newline_main () ;
     append_main "let %s () =" (name t);
-    append_main "  %s.connect ()" (module_name t);
+    append_main "  %s.connect %s" (module_name t) (id t);
     newline_main ()
 
-  let clean () = ()
+  let clean _ = ()
 
   let update_path t _ = t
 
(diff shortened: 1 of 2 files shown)


hannesm:
hmm, I'm not sure about this.. should this really be exposed to `config.ml` at all? I do appreciate the logic to select the entropy source based on mode (and a target - if we have sth within mirage - should be the other case).

I wouldn't want to enable users to shoot into their feet to easily, and rather select entropy sources within mirage based on the available information.

djs55:
I'd be keen to remove `weak_entropy` completely once we have something like `RDSEED` or perhaps the entropy gathering through interrupt timings thing (as long as someone other than me writes it -- I'm sure I would mess it up!) My thinking was: I'd like us to stay compatible with AWS (no `xentropyd`, currently no other good option) and I would prefer if the user shot themselves in the foot (by typing 'w e a k') to us shooting them in the foot by default, if you see what I mean :-)

hannesm:
understood. I prefer to have users doing sth more advanced (such as manually patching the mirage-entropy device (which unfortunately is released in its current broken state (weak entropy) on xen)) than writing `weak_entropy` in their config...

avsm:
`weak` entropy is an odd name, since it can eventually be as good as current virtualised Linux is (e.g. via `RDSEED` or other harvesting).  It just happens to be weak right now...but isn't `builtin` or similar better?

djs55:
I chose 'weak' because it describes the interface... or the promise the
system makes. Perhaps a better name would be 'no_entropy' :-) My idea was:
you say what you need in config.ml, and mirage tries to supply it. If it
can't then the unikernel prints a message and hangs. (Actually Im not sure
what happens with entropy users if no entropy is fed in)

I think 'builtin' is a better name for the 'type id' poly variant in the
mirage-entropy implementation though since it does describe the
implementation. (I used 'weak' there too, probably a mistake)

I'm still mulling over @hannesm's feedback. My best idea so far is to make
the mirage tool quiz the user about the implications of bad entropy: no
truly secure connections, must promise not to process anyone's private
data. A bit like a security/privacy captcha :-) perhaps all 'no_entropy'
binaries should expire a few days after they're built? Dunno. Or we could
require users patch / pin the code to get the unsafe behaviour.

We ought to figure something out soon, since the shipping version is bad
(on Xen).

(It was a mistake to type this on a phone! Sorry for inevitable typos and
autocorrect disasters)

hannesm:
ok - taking a step back - we had some discussion of this sort already last year https://github.com/mirage/mirage-entropy/commit/a220ebf7d2bb5ee1a0bfdd43ae988cdb077b0f8f ... also recap on http://lists.xenproject.org/archives/html/mirageos-devel/2014-11/msg00146.html

My point is: either there is entropy or not. The current released mirage-entropy on Xen should not be out there (I tried here https://github.com/mirage/mirage-entropy/pull/6).

The upcoming question is who (and where) is entropy configured (and having the possibility to configure no entropy will lead to people configuring no entropy without thinking about it..):
- we do have a good story if there's control over dom0 (xentropyd)
- rdrand/rdseed shouldn't be too hard
- lwt timer loop as well

Does the mirage tool know which sources will be available when generating a main.ml? Certainly not (how should it know whether we'll deploy on x86 with control over dom0 or no control over dom0?).
But: mirage tool could take a guess (e.g. no need to check for rdrand/rdseed on arm) and generate some runtime code (or mirage-entropy should) - where it tests for xenstored availabality, rdseed/rdrand availability - and bails if it can't find any entropy (this would make mirage-entropy-xen always depend on mirage-console-xen, which I don't think is too bad)!?

(related questions are: do we agree this solution is good? if so, who's going to implement it [runtime config, rdseed/rdrand, lwt timings]? is it worth to look into more than those 3 entropy sources (obviously, additional static strings etc can be integrated fairly easily once the config thingy is in place)?)

The user experience will be: either have entropy, or fail at boot/init time. no configuration required.

pqwy:
What problem does this solve, exactly?

The problem of running crypto on systems where we don't yet have or are even unable to supply entropy?

Or the problem of easy development and testing without setting up `xentropyd`?

avsm:
I'm going to merge a version of this patch without the `weak` option.  I agree that `mirage-entropy-xen` doing dynamic probing to (e.g.) find xentropyd makes more sense than doing this at compile time.

PR #360, opened 2015-01-31
[RFC] New system for passing parameters to unikernels

nojb:
A modest improvement to the current `mirage` tool that makes it easy to configure unikernels parameters (such as IP addresses).  These parameters can even be set at run-time (without recompiling) and they are automatically listed in the man page of `mirage configure`.  See https://gist.github.com/nojb/f9c262d5fd691e6171ca for more details (NB the signature of `Mirage.Key` has changed a little) or https://github.com/nojb/mirage-better-params-ex for an example.

**Remarks**

1. All `config.ml` files need to be revised to work with the new system.
2. It would be nice to be able to set configuration parameters that are used by the actual unikernel, however, this is not yet supported in the current code.  Only the configuration combinators can access the configuration keys.
3. I am not sure if configuration choices such as *DHCP vs Static* can be expressed in the new system.
4. The implementation is hacky in parts.
5. Not all combinators have been ported to use the new system.
6. For some `CONFIGURABLE`s (e.g. `Crunch`), the notion of run-time parameter does not make sense because its configuration is purely a compile-time operation.  This type of restriction is not yet accounted for in the current code.

Since this will break all existing `config.ml`'s, it makes sense to try to come up with a solution that is as flexible as possible.  I don't think  that this system is sufficiently so in its current form.  However I wanted to put this out to stimulate discussion on this subject.

Any and all comments welcome.

**How to test**

````
opam pin add mirage https://github.com/nojb/mirage#better-params
````

**Example**

See https://github.com/nojb/mirage-better-params-ex.

### CODE REVIEW

**Comment by samoht** [comment]
We had a quick chat about that patchset with Nicolas. I think the best way to go forward with this is to let the current combinators as there are but add new ones which allow dynamic parameters (under a `Dynamic` sub-module for instance). This will let us:

- Keep the current `config.ml` files unchanged. Not need to rewrite all the tutorials. No need to add version checks.

- Allow to iterate on the proper design of dynamic parameters. It is still a bit unclear what we really need, but we need something soon to start to experiment with. It is however not a good idea to ask user to switch their `config.ml` every time we fix the design.

Also, we should support simple named runtime parameters as well in this first proposal, as it was what Magnus needed at the beginning. Should be quite easy to add and to improve later on.

**Comment by samoht** [comment]
See https://github.com/MagnusS/mirage-bootvar for simple k/v runtime parameters that we can generate combinators for.

**Comment by nojb** [comment]
Thanks!  This is useful.  I'll try to get the changes we discussed done in
the next couple of days so that people can start playing with it.

**Comment by amirmc** [comment]
Hey folks, just curious how this one is going. Have you had a chance to look at this @nojb?

**Comment by nojb** [comment]
Yes @amirmc, I hope to have the revised version ready sometime tomorrow.  Sorry for the delay - been ill last couple of weeks.

**Comment by amirmc** [comment]
No problem, I was just browsing open issues to see what discussions are still outstanding. 
I hope you feel better!

**Comment by nojb** [comment]
I just pushed a new version of this patch (apologies for the endless delays).  This new version is not meant to be a final design, but rather an implementation of enough functionality so that users can play around with it and hopefully help us figure out what *should* the right design be.

Notable differences with the previous version:

- 100% backwards compatible* - absolutely no changes to the `config.ml` files is needed.  Rather, for those combinators that can be customized via the command line (mainly the ones related to TCP/IP settings), those values specified in the `config.ml` file are taken as defaults, but can be overriden by passing command-line parameters.

*: `mirage configure` now takes the explicit config filename as an optional parameter introduced by `-f` or `--file` instead of using a positional parameter.  This simplifies the logic in `Main.configure`.

- There is a new submodule, `Configvar`, in `Mirage_runtime` that is used both by the `mirage` tool and by the generated `main.ml` files to handle command-line parameters.  It depends on `Cmdliner`.  Another internal module `Mirage_key` contains all the logic that is only used only in the `mirage` tool.

- The configuration phase of the mirage tool (`mirage configure`) now generates a file called `bootvar.ml` that contains all the data related to configuration keys.  In general configuration keys are identified by their names and they are mapped to global variables in the `Bootvar` module by *ocamlification* (removing all characters not on `['a'-'z''A'-'Z''0'-'9''-''_']` and translating '-' into '_').

- The unikernel developer can declare configuration keys to be used by using the Mirage.add_to_bootvars combinator. The signature is:
```
  val add_to_bootvars: ?doc:string -> string -> unit
```
  so that one can specify a docstring and the key `name`. In the unikernel itself the value of this configuration key can be accessed as `Bootvar.name'`, where `name'` is the ocamlification of `name` (see above).

- One can set up dhcp vs. static stacks at *runtime* by using the `--key-stackv4` optional parameter (do `mirage configure --help` for the format, see example below).

Example:

Pin `https://github.com/nojb/mirage#better-params`, go to `mirage-skeleton/static_website`, and do:
```
mirage configure
make
mir-www --help
mir-www --key-stackv4=dhcp
mir-www --key-stackv4=static:0.1.2.3:1.1.1.1:1.2.3.4,3.4.6.7
```

So, how to test it?

First,

0. `opam pin add mirage https://github.com/nojb/mirage#better-params`

And then,

1. Navigate to your favourite unikernel source,
2. Type `mirage configure --help` and look under the heading `UNIKERNEL PARAMETERS`
3. Configure and make, maybe modifying some of the unikernel parameters via the command line.

Alternatively,

2'. Build your unikernel on Unix (with `mirage configure; make`), producing output `mir-<whatever>`.
3'. Type `mir-<whatever> --help` and look under the heading `UNIKERNEL PARAMETERS`

Any and all comments welcome.

**Comment by nojb** [comment]
One more comment: the patch depends on having a functioning `OS.Env.argv` function.  This has only a dummy implementation on Xen right now so run-time parameters will not work on Xen.  One possibility would be to port the code from `mirage-bootvar` (https://github.com/MagnusS/mirage-bootvar) to obtain a simple implementation of `OS.Env.argv` on Xen.

**Comment by samoht** [comment]
That seem pretty cool. So you finally decided to not use `typed` keys?

**Comment by nojb** [comment]
Yes, using typed keys does not add much since the rest of the code generation machinery is completely untyped. Also if one keeps the types then you need to wrap them into existentials each time you want to put them inside lists, sets, etc, which is too cumbersome...

**Comment by hannesm** [comment]
I do not really understand the goal here: you want to pass key/value pairs from the command line to a unikernel? and all the keys life in a flat namespace?
how would I pass static IPv4 configuration to a unikernel with multiple ethernet interfaces? Or multiple certificate and private key pairs to a service which listens on multiple ports?

**Comment by nojb** [comment]
The present goal is very modest: simply to make it a little more convenient to pass the usual configuration information to unikernels on the command line, and allow to do this at run-time.  Right now it is simply impossible to configure multiple IPv4 configurations or other similar setups using this system.

As you point out, the flat namespace dictated by the command-line does not match the space of all possible configuration parameters (which is more like a tree or a forest rather than a list).  I would be very interested in a solution to the problem of designing an easy interface to the full space of configuration parameters.

Any ideas?

**Comment by samoht** [comment]
> The present goal is very modest: simply to make it a little more convenient to pass the usual configuration information to unikernels on the command line, and allow to do this at run-time.

This looks a good goal in the short term: We really need to 1) pass dhcp/static ip configuration on the command-line and 2) pass arbitrary key/value pairs to the runtime, which both seem to be achieved by the patch. Can you rebase to master? I'll have a proper look at it tomorrow.

**Comment by nojb** [comment]
I rebased and squashed everything to a single commit as the history was pretty worthless anyway (it was full of rewrites and re-rewrites, etc...).

File: lib/main.ml [hunk 1]
```diff
@@ -87,8 +87,8 @@ let file =
     ~doc:"Configuration file for Mirage. If not specified, the current directory will be scanned. \
           If one file named $(b,config.ml) is found, that file will be used. If no files \
           or multiple configuration files are found, this will result in an error unless one \
-          is explicitly specified on the command line." [] in
-  Arg.(value & pos 0 (some string) None & doc)
+          is explicitly specified on the command line." ["f"; "file"] in
```
samoht: do you really need this? It will break `mirage-skeleton` and needs some updates to `mirage-www` (both are fine if they are really needed, though).

nojb: I seem to remember that it was just to simplify some of the logic below - let me look into it again and I'll see if I can restore the previous behaviour.

nojb: Ok, so the situation is as follows:

- We need to compile the `config.ml` (or similar) in order to be able to construct the cmdliner term.
- We can't use `Cmdliner.eval` to extract the config file name because we haven't constructed the full cmdliner term yet.
- Right now I am using `Term.eval_peek_opts` to extract the config file name but that only works with optional parameters.
Indeed, there is no way to know if something is a positional parameter when one does not know what the optional parameters are: i.e., `cmd -X Y` could be parsed as either an optional parameter `X` with value `Y` or a flag `X` with positional parameter `Y`.

I don't see an easy way to work around this.  Any ideas?

samoht: I see, that's indeed a good reason to use named argument. I'm fine to modify `mirage-www`.

samoht:
You'll also need to modify the test scripts to use `-f config.ml`.

samoht:
@nojb any update on this? `mirage-bootvar-xen` is now released so parameters can be passed to unikernels via xenstore.

nojb:
I rebased to master and make use of `mirage-bootvar-xen` to get `argv` on Xen (depends on mirage/mirage-bootvar-xen#4).

The current code is hardly ideal, but it is backwards compatible* and I feel that feedback from actual users is crucial to improve it further.  I find it very hard to focus on this without such feedback.

\* The only incompatibility is that custom `config.ml` files used to be specified with a positional argument, but now are specified with an optional argument `-f` (see the discussion above for the reason).

#### Micro-tutorial:

##### Passing key-value parameters to the unikernel

To specify key-value parameters to the unikernel, pass a list of `Mirage.key`'s to `Mirage.register` as folows:
```ocaml
open Mirage
let _ =
  register ~bootvars:[bootvar "my-bootvar"] [ ... ]
```
The `bootvar` function is defined in `Mirage` and has signature:
```ocaml
val bootvar: ?doc:string -> ?default:string -> string -> key
```
allowing to specify an optional docstring and a default value (only string-valued user arguments are supported currently).

This parameter will be listed in the `mirage configure --help` man page and can be set
- at compile-time:
```bash
mirage configure --key-my-bootvar="my-bootvar-value"
```
- at run-time, on unix:
```bash
./unikernel --key-my-bootvar="my-bootvar-value"
```
or on Xen:
```bash
xl create unikernel.xl 'extra="key-my-bootvar=my-bootvar-value"'
```
From inside the unikernel, one can fetch the argument by doing:
```ocaml
Printf.printf "%s\n" (Bootvar_gen.my_bootvar ()) (* will print "my-bootvar-value" *)
```
(here `bootvar_gen.ml` is a file generated during `mirage configure`).

##### Specifying device combinator arguments

One can also specify some of the arguments used by the device combinators in `Mirage` via the command line.  The only visible manifestation of this new feature is the UI: the man page of `mirage configure --help` and the new command line parameters (at compile-time and run-time as above):
```bash
cd mirage-skeleton/static_website
mirage configure --help # will list --key-stackv4 as an optional argument
mirage configure --key-stackv4=dhcp # use dhcp
mirage configure --key-stackv4=static:10.0.0.2:255.255.255.0:10.0.0.1 # use static ip, netmask, gateways
```
*Limitations:* many.  More than one instantiation of the same device will share the configuration parameter (so if two `stackv4`'s are present both would be configured with the same static ip, *if* the command line argument was specified).  The argument format is ad-hoc and contrived, due to the need of mapping the configuration space (which is naturally a tree, two-dimensional) to a command-line string (1-dimensional).  Etc.

Drup:
I rebased and adapted for the whole https revolution here: https://github.com/Drup/mirage/tree/better-params

Nasty bits that I'm not fond of:
- You can't write something like that:
```ocaml
let x = f @@ Bootvar_gen.my_var ()
```
Since it will be executed before `Bootvar_gen.parse_argv`. You end up with either the default value or a very delicate `assert false`.
- The handling of format (ipv4,ipv6,dhcp, ...) is at the moment hardcoded in `mirage_runtime.ml`. It's not very nice. I'm wondering how we could delegate this to a library (the format handling is done both at configuration and at runtime ...).

nojb:
Yes, all of that needs further work.  I don't have time to work on this right now, but here is one idea I have been thinking about.

The basic problem is as follows: for each device `D : CONFIGURABLE`, the **type** `D.t` needs to be made into a value (a tree-like data structure describing the possible configuration choices), say of type `D.t ty`.

Just to be concrete in the case of `S = STACKV4_direct`, the type `S.t` is
```ocaml
type t = {
    clock  : clock impl;
    time   : time impl;
    console: console impl;
    network: network impl;
    random : random impl;
    config : [`DHCP | `IPV4 of ipv4_config];
  }
```
and the hypothetical value describing the configuration would look something like
```ocaml
`Record [
  `Field ("console", `String);
  `Field ("network", `Choice [`Tap0; `Custom `String]);
  `Field ("config", `Choice [
    `DHCP;
    `IPV4 (`Record [
      `Field ("address", `IPv4);
      `Field ("netmask", `IPv4);
      `Field ("gateways", `List `IPv4)
     ])
   ])
] : S.t ty
```
This value would be enough to define serializers/deserializers from any textual format (such as command-line arguments or JSON) with signatures `'a ty -> 'a -> string` and `string -> 'a ty -> 'a option` as well as produce documentation such as `man` pages.  Further, this value could be marshalled between the `config.ml` stage and the `main.ml` stage to implement run-time parameters.

This general approach is hard to realize without some compiler support ([runtime types](https://www.lexifi.com/files/resources/runtime_types.pdf) anyone?), and the current code amounts to a fragile, hand-coded version of this general idea.

Of course one doesn't really need full runtime types to do this… maybe a suitably chosen datatype would be enough to describe the types `D.t` for `D : CONFIGURABLE`…

samoht:
Subsumed by #441

PR #364, opened 2015-02-16
[RFC] Don't put '-classic-display' in the generated Makefiles

talex5:
Classic mode makes it hard to see warnings among the verbose output. If some users want this, I think it would be better to make it a general ocamlbuild configuration setting rather than special-casing it for Mirage.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,5 @@
+* Don't put '-classic-display' in the generated Makefiles.
+
 2.2.1 (2015-01-29):
 * Fix logging errors when `mirage` output is not redirected. (#355)
 * Do not reverse the order of C libraries when linking.  This fixes Zarith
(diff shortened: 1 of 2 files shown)


avsm:
Fine by me, although I'd prefer a `VERBOSE` variable in the Makefile that activates `-classic-display` (which is a ridiculously obscure flag name)

samoht:
good to me as well. The only advantage of `-classic-display` is that it's easier to debug in case something goes wrong. This was pretty useful at the beginning, but now the mirage tool is mature enough to not need it anymore I think.

talex5:
Does `cat _build/_log` work instead of a `VERBOSE` flag?

avsm:
The `VERBOSE` flag is interactive, whereas the cat happens after the build.  There are two things that a VERBOSE flag could set:

- the `classic-display` flag in the build
- the `OPAMVERBSE` env variable to see what the dependency step is doing.

PR #369, opened 2015-03-01
modify signature of checksum:

hannesm:
instead of requiring the full stack of frames (available only when sending),
 pass source, destination address and protocol

this makes it useful for both sending and receiving of frames


samoht:
do we still need this?

hannesm:
certainly this is part of an unfinished abandoned branch, and there should rather be a rethinking of how to clip together a network flow graph (maybe looking at click os etc), instead of hardwiring ip over ethernet.. (which is fine for now at least)

PR #370, opened 2015-03-02
Fix typo on manual page

jvirtanen:


avsm:
Thanks! The `setup.ml` typo will get overwritten the next time the OASIS files are regenerated, but the manual page commit is correct.

PR #378, opened 2015-03-16
Call `opam depext`, add opam files

samoht:


lib/mirage.ml
```diff
@@ -2380,7 +2380,11 @@ let configure_opam t =
         | major::minor::_ ->
           let major = try int_of_string major with Failure _ -> 0 in
           let minor = try int_of_string minor with Failure _ -> 0 in
-          if (major, minor) >= (1, 2) then opam "install" ps else version_error ()
+          if (major, minor) >= (1, 2) then (
```
avsm: Why is this check necessary now that OPAM 1.2 is the minimal version?

samoht: To still warn the user if he is using an older opam version I guess

lib/mirage.ml
```diff
@@ -2380,7 +2380,11 @@ let configure_opam t =
...
+            opam "install" ["depext"];
```
avsm: I like this feature, but I don't like that it attempts to install depext on every `mirage` invocation.  Isn't there a more lightweight way to check if `depext` is installed than calling the solver every time? (`opam install` is slow on ARM).  I'd also like a flag or Makefile variable that skips DEPEXT installation for systems where it's not relevant.

lib/mirage.ml
avsm: Oh I'm sorry -- I thought the version check was elsewhere, but this also does that.  I just didnt want them to be duplicated.

samoht:
> Isn't there a more lightweight way to check if depext is installed than calling the solver every time? 

 I'm not sure. Maybe @altgr will know better. Agreed for the command-line flag, though.

avsm:
I was thinking of `which opam-depext` :-)

samoht:
See https://github.com/mirage/mirage/pull/385

AltGr:
`opam list --external` should help if you don't want to call the solver. It can be combined with `--required-by` and `--recursive` if needed.
Just `opam list -s --external` will list all depexts flags for the currently installed packages, while `opam depext -ls` will detect the system's flags and list the corresponding system packages.

Issue #381, opened 2015-03-18
Better error handling


samoht:
See http://lists.xenproject.org/archives/html/mirageos-devel/2015-01/msg00143.html

Tasks:

- [ ] add `val pp_error: Format.formatter -> error -> unit` in both `V1.FLOW` and `V1.DEVICE`
- [ ] mirage-console
- [ ] mirage-entropy-unix
- [ ] mirage-net-macosx
- [ ] mirage-net-unix
- [ ] vchan

hannesm:
`pp_error` should return a `string` (as described [here](https://github.com/mirage/mirage-www/pull/274/files#diff-abb8c0c60c75065f86ff29942a483ec4R254))!?

samoht:
You want to use it with `Format.printf "error is %a" pp_error e` so the should be `unit` (there's a typo on the page)

hannesm:
yes, discovered, thought, and deleted my comment..

Drup:
Bumping that one, it would make the connect code nicer and I would prefer to do it while I'm shuffling everything.

talex5:
Some related fixes for the error reporting are tracked in https://github.com/mirage/mirage-dev/pull/107

hannesm:
for connect we have nicer errors now (and code even got simpler!!!) -- for normal operations i suspect we still could use some normal result types and pretty printers...

hannesm:
#615 is in, more to come... closing this outdated issue

PR #383, opened 2015-03-18
Add `pp_error` to FLOW and use a result type

samoht:
Related to #381

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -139,7 +139,7 @@ module type FLOW = sig
   type buffer
   (** The type for memory buffer. *)
 
-  type flow
+  type t
   (** The type for flows. A flow represents the state of a single
       stream that is connected to an endpoint. *)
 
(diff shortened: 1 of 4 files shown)


types/V1.mli
```diff
@@ -29,6 +29,9 @@ module type DEVICE = sig
   type error
   (** An error signalled by the device *)
 
+  val pp_error: Format.formatter -> error -> string
```
hannesm: this should now be `unit` here!?

samoht: it was just to check if someone was following :-)

samoht:
These patches should be revisited now that we have a `Rresult` and `Fmt`. I let that issue open to remember (until we got a Mirage's Cuekeeper instance to manage the project)

hannesm:
I'm very much in favour of reviving this PR, and since this breaks the API, we should also rename `V1` and `V1_LWT` to `M` and `M_LWT` (or some other name which does not include versioning information).  Versioning is handled by opam :)

yomimono:
I'm closing this PR as we probably won't incorporate these exact commits, and we've merged quite a few other things that conflict with it.  But the spirit lives on in https://github.com/mirage/mirage-dev/pull/166 and associated commits and discussions, and we'll likely end up merging something resembling this.  Thanks very much, @samoht, for leading the charge here!

PR #386, opened 2015-03-19
Improve docs

samoht:
No semantics change, just cut long lines, add section headers and rewrite few wordings. Could be still improved, though.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -48,7 +48,7 @@ module type TIME = sig
   (** The type for potentially blocking I/O operation *)
 
   val sleep: float -> unit io
-  (** [sleep nsec] Block the current thread for. {b FIMXE:} remove float. *)
+  (** [sleep nsec] Block the current thread for. {b FIXME:} remove float. *)
 end
 
 module type RANDOM = sig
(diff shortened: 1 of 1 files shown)


types/V1.mli
```diff
@@ -1,6 +1,6 @@
 (*
- * Copyright (c) 2011-2014 Anil Madhavapeddy <anil@recoil.org>
- * Copyright (c) 2013      Thomas Gazagnaire <thomas@gazagnaire.org>
+ * Copyright (c) 2011-2015 Anil Madhavapeddy <anil@recoil.org>
+ * Copyright (c) 2013-2015 Thomas Gazagnaire <thomas@gazagnaire.org>
```
dbuenzli: IANAL but I personally stopped updating copyrights date in my sources. IIUC basically the first date tells when you claim the copyright for the first time and that should be enough, it will be protected, at least in [Switzerland](http://www.admin.ch/opc/en/classified-compilation/19920251/index.html#a29), for 50 years after your death (fun that it's shorter by 20 years for computer programs...). So unless you care about a few years more for your descendants I don't think it's very useful. I would love to get some confirmation by a legal person though.

samoht: to be honest, I'm just doing this (as a bad way) to keep track of the active projects.

avsm: Yeah, I just do it as a marker for my own copyrights, not for a particular legal reason.

PR #387, opened 2015-03-19
Dead types

samoht:
Kill all these types!

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -33,13 +33,6 @@ module type DEVICE = sig
   type t
   (** The type representing the internal state of the device *)
 
-  type error
-  (** The type for errors signalled by the device *)
-
-  type id
-  (** This type is no longer used and will be removed once other
-   * modules stop using it in their type signatures. *)
-
   val disconnect: t -> unit io
   (** Disconnect from the device.  While this might take some time to
       complete, it can never result in an error. *)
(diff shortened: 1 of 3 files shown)


samoht:
Side-effect of this commit: `mirage-types` does not need `io-page` anymore.

types/V1.mli
```diff
@@ -16,194 +16,195 @@
...
-  (** Time operations for cooperative threads. *)
+  (** {1 Time operations for cooperative threads} *)
 
   type +'a io
-  (** A potentially blocking I/O operation *)
+  (** The type for potentially blocking I/O operation *)
 
   val sleep: float -> unit io
-  (** [sleep nsec] Block the current thread for [TODO remove float] *)
+  (** [sleep nsec] Block the current thread for. {b FIMXE:} remove float. *)
```
mor1: FIMXE -> FIXME :)

samoht: right, all add a `fixme` note to remember to fix it :p

types/V1.mli
```diff
@@ -456,72 +438,72 @@ module type IP = sig
...
+  (** The type for callback functions that adds the UDP metadata for
+      [src] and [dst] IP addresses, the [src_port] of the connection
+      and the [buffer] payload of the datagram. *)
 
   val input: listeners:(dst_port:int -> callback option) -> t -> ipinput
-  (** [input listeners t] demultiplexes incoming datagrams based on their destination
-      port.  The [listeners] callback is will either return a concrete handler or
-      a [None], which results in the datagram being dropped. *)
+  (** [input listeners t] demultiplexes incoming datagrams based on
+      their destination port.  The [listeners] callback is will either
```
mor1: "is will" -> "will" ?

samoht:
Actually we still need `DEVICE.error` as we still rely on the fact that implementation should have a proper `connect` function. That's annoying that we can't really enforce in the signature though.

PR #390, opened 2015-03-24
Support for latest TLS

samoht:


lib/mirage.ml
```diff
@@ -471,10 +472,14 @@ module Entropy = struct
     append_main "module %s = %s" (module_name t) (construction t) ;
     newline_main () ;
     append_main "let %s () =" (name t);
-    append_main "  %s.connect %s" (module_name t) (id t);
+    append_main "  %s.connect %s >>= function" (module_name t) (id t);
+    append_main "  | `Error e -> %s" (driver_initialisation_error "entropy");
+    append_main "  | `Ok entropy ->";
+    append_main "  %s.handler entropy Nocrypto.Rng.Accumulator.add_rr >>= fun () ->" (module_name t);
```
hannesm: @pqwy is this the intended call semantics for `Nocrypto.Rng` feeding? shouldn't there be at least the named `source` parameter (as in [lwt](https://github.com/mirleft/ocaml-tls/blob/master/lwt/tls_lwt.ml#L277), [nocrypto.mli](https://github.com/mirleft/ocaml-nocrypto/blob/master/src/nocrypto.mli#L429), [rng.ml](https://github.com/mirleft/ocaml-nocrypto/blob/master/src/rng.ml#L139), [fortuna.ml](https://github.com/mirleft/ocaml-nocrypto/blob/master/src/fortuna.ml#L113))?
Could we please, before releasing `2.4.0` with TLS support, sit down and convince ourselves that the code is in a good shape (entropy is not completely broken, have a story for authenticators, ...)?

samoht: @hannesm I'm happy to delay the release until you are happy enough for an initial release (but I would like to unlock the co-instability tls and mirage as soon as possible)

hannesm: I'm eager to have a release asap - but really don't want to mess this up. better safe (and release slightly later) than sorry (and introduce various kinds of security issues).
Supporting TLS out of the box in MirageOS would be a great time to have a security contact mail address and release security advisories if there are security relevant issues found (and describe our practices somewhere on http://openmirage.org).

avsm: Agreed.  Let's spool up all the changes in mirage-dev on trunk branches, and do an audit before a release.  I'd also like to do a careful read through of all the changes. The call tomorrow is a good time to discuss the security process we'd like to initiate.

samoht: Agreed, I've put a line in the Agenda for tomorrow call.

hannesm: so `~source` is added by mirage-entropy ([unix](https://github.com/mirage/mirage-entropy/blob/master/unix/entropy_unix.ml#L66) and [xen](https://github.com/mirage/mirage-entropy/blob/master/xen/entropy_xen.ml#L165) version).

pqwy: I'm trying to figure out what to even do here.

If requesting entropy hooks it up into `nocrypto`,
* there is no point in requesting an `ENTROPY` provider in the first place -- what you get with this is the entire crypto library; and
* a transient, internal API that used to live between `nocrypto` and `tls` suddenly goes public.

IOW if `ENTROPY` pulls in `nocrypto` it's completely misnamed; what it pulls in is, in fact, a particular crypto library instantiated with an entropy source, which gives nothing useful by itself. If `nocrypto` depends on `ENTROPY`, OTOH, the entire library needs to be functorized over it, and subsequently all of its clients over that, which is something I want to avoid.

Can I please request a small freeze on this to think it through? As it currently is, a small, internal link between two libraries is just becoming a public API in a shape in which I would rather not have it, and we end up with a roundabout way of expressing a hard dep on `nocrypto`.

That, and at some point [this](https://github.com/mirage/mirage-entropy/commit/a4a318902d5112625bf3e94c2d7b6e3d665a4cd0) got released, which I cannot get to work reliably and did not intend to release, but once out, short of rolling back, it precluded me from integrating the [alternative solution](https://github.com/pqwy/mirage-entropy/commits/pinata) which is what, in fact, I want the CSPRNG to be fed with.

I am beginning to think that the very existence of `ENTROPY` and `mirage-entropy` is creating more problems than it solves. A much simpler way would be to internalize that code as submodules of `nocrypto`, provide an idempotent trigger to start entropy harvesting, and switch it on in the init preamble of everything `mirage` tool knows about and that might care about cryptography (`conduit`, `tls`).

talex5: Whatever the solution, it would be nice if we could get rid of the global mutable variable `Rng.Accumulator.acc`. It's very useful to see all sources of non-determinism in a program. As it is, there is no visible link between `entropy` and `tls` in the generated `main.ml`.

PR #393, opened 2015-04-16
Adapt to latest changes in conduit and mirage-http

samoht:
I've aslo deleted some dead code.


samoht:
The test are failing because it needs a release of conduit, mirage-http and tcpip. I have checked manually that I can compile mirage-skeleton, so merging.

PR #394, opened 2015-04-26
Enthalpy

pqwy:
* Removes mentions of the `ENTROPY` interface.
* Removes `TLS_over_conduit` that depended on it.
* If `nocrypto` is found in the set of transitive dependencies, they are extended with its appropriate seeding module and the seeding process is triggered just before the application-side of the unikernel runs.

I apologise for simply deleting tls+conduit, wasn't sure what to do with it. In general, anything that depends on `tls` should just work now, so it should be a breeze to re-add.

The detection code is horribly wonky. Since there is no clearly defined point when the variables `ps` and `ls` are consumed, and to avoid running `ocamlfind` several times, it uses a side-effect and a fragile assumption about the ordering of calls to `packages` and `libraries`. I could not figure out a non-intrusive way to extend the tool and make this clean at the same time.

Another problem is in the detection logic: it scans for `nocrypto` only in the set of (opam-level) *packages* depended on, not in the set of (ocamlfind-level) *libraries*. The detection will fail if someone has, say, `tls` installed and wants to use it, but forgets to `add_to_opam_packages` it. I wasn't sure what exactly to do here -- feedback appreciated.

(Needed for [this](https://github.com/mirleft/ocaml-nocrypto/issues/55).)

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -26,6 +26,8 @@ module StringSet = struct
     List.iter (fun e -> s := add e !s) l;
     !s
 
+  let to_list s = fold (fun x xs -> x :: xs) s []
+
 end
 
 let main_ml = ref None
@@ -1939,6 +1941,43 @@ type tracing = Tracing.t
 let mprof_trace ~size () =
   { Tracing.size }
 
+module Nocrypto_entropy = struct
+
+  module SS = StringSet
+
+  let needed = ref None
+
+  let rec detect_if_needed packages =
+    let scan () =
+      let trans = OCamlfind.query ~recursive:true (SS.to_list packages) in
+      List.exists ((=) "nocrypto") trans in
+    if !needed = None then needed := Some (scan ())
+
+  let is_needed () =
+    match !needed with
+    | None   -> failwith "Nocrypto_entropy: query before detection"
+    | Some n -> n
+
+  let packages packages =
+    detect_if_needed packages ;
+    match (!mode, is_needed ()) with
+    | (`Xen, true) -> SS.singleton "mirage-entropy-xen"
+    | _            -> SS.empty
+
+  let libraries () =
+    match (!mode, is_needed ()) with
+    | (`Xen             , true) -> SS.singleton "nocrypto.xen"
+    | ((`Unix | `MacOSX), true) -> SS.singleton "nocrypto.lwt"
+    | _                         -> SS.empty
+
+  let preamble () =
+    match (!mode, is_needed ()) with
+    | (`Xen             , true) -> "Nocrypto_entropy_xen.initialize ()"
+    | ((`Unix | `MacOSX), true) -> "Nocrypto_entropy_lwt.initialize ()"
+    | _                         -> "return_unit"
+
+end
+
 type t = {
   name: string;
   root: string;
@@ -1993,6 +2032,7 @@ let packages t =
       let ps = StringSet.of_list (Impl.packages j) in
       StringSet.union ps set
     ) ps t.jobs in
+  let ps = StringSet.union ps (Nocrypto_entropy.packages ps) in
   StringSet.elements ps
 
 let ls = ref StringSet.empty
@@ -2012,6 +2052,7 @@ let libraries t =
       let ls = StringSet.of_list (Impl.libraries j) in
       StringSet.union ls set
     ) ls t.jobs in
+  let ls = StringSet.union ls (Nocrypto_entropy.libraries ()) in
   StringSet.elements ls
 
 let configure_myocamlbuild_ml t =
@@ -2361,7 +2402,9 @@ let configure_main t =
   List.iter configure_job t.jobs;
   let names = List.map (fun j -> Printf.sprintf "%s ()" (Impl.name j)) t.jobs in
   append_main "let () =";
-  append_main "  OS.Main.run (join [%s])" (String.concat "; " names)
+  append_main "  OS.Main.run (%s >>= fun () -> join [%s])"
+              (Nocrypto_entropy.preamble ())
+              (String.concat "; " names)
 
 let clean_main t =
   List.iter Impl.clean t.jobs;
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1943,35 +1943,50 @@ let mprof_trace ~size () =
 
 module Nocrypto_entropy = struct
 
+  (* XXX
+   * Nocrypto needs `mirage-entropy-xen` if compiled for xen.
+   * This is currently handled by always installing this library, regardless of
+   * usage of Nocrypto.
+   * As `libraries` is run after opam setup, it will correctly detect Nocrypto
+   * usage and inject appropriate deps for linkage, if needed.
+   * The `packages` function, unconditional installation of
+   * `mirage-entropy-xen`, and this comment, should be cleared once we have
+   * conditional dependencies support in opam.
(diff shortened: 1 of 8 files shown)


pqwy:
Detection maybe makes a little more sense now:

The seeding package is pulled in iff any package needs (or is) `nocrypto`.

The seeding library *and* the starting trigger are pulled in iff any library needs (or is) `nocrypto`.

samoht:
> I apologise for simply deleting tls+conduit, wasn't sure what to do with it. In general, anything that depends on tls should just work now, so it should be a breeze to re-add.

this makes sense, these will be replaced by a proper HTTPs combinator later.

Regarding the patch: the only point which is not totally clear by reading the patch is when the call to `ocamlfind` is done. To be valid it needs to be done **after** the `opam` packages are installed, as `ocamlfind query -r <xxx>` will return stuff about installed packages. [An other option is to ask opam instead but that will be slower (as you need to call the solver) and less precise so the current approach is better]

I'll try to play with the patches tomorrow.

pqwy:
First the presence of a package is detected through [`configure_opam`](https://github.com/pqwy/mirage/blob/enthalpy/lib/mirage.ml#L2416), which calls `packages` and this is remembered. Then [`configure_makefile`](https://github.com/pqwy/mirage/blob/enthalpy/lib/mirage.ml#L2418) sees the same answer for packages, and calls `libraries`.

So I think you are right:

On a clean installation, if a user declares `conduit` via, say, a combinator, it in turn declares `tls`. As this is not yet installed, `nocrypto` is not seen by `ocamlfind`, and `mirage-entropy-xen`, its dep for entropy on xen, is not added to the package set. But `nocrypto` is still pulled in by opam to satisfy `tls`, and since `mirage-xen` is pulled in too, opam configures `nocrypto` to also require `mirage-entropy-xen`. Depending on whether opam can express that the presence of package `A` causes package `B` to require package `C`, this either installs everything nicely, or aborts because `nocrypto` now requires `mirage-entropy-xen`.

If the opam succeeded, the call to `libraries` correctly detects `nocrypto` (the lib) as being transitively present, adds the correct entropy lib to the linkage set, and emits the init code.

If there is a way to make opam install `mirage-entropy-xen` before installing `nocrypto` if `mirage-xen` is installed, i.e. if there is a way to conditionally *require* (as opposed to just conditionally using), we can delete the code around `package` and rely on that. Is this possible?

Otherwise, yes, this is incorrect and the package detection might have to go via opam. Can we just ask it for immediate deps?

samoht:
Continuing the discussion of https://github.com/mirleft/ocaml-nocrypto/pull/56#issuecomment-96599311 here to centralise stuff.

>  while this is not nice and clean, I thought the mirage tool would dependency-inject mirage-entropy-xen if nocrypto is in use (nocrypto needs an optional dependency to mirage-entropy-xen, then it'll be recompiled after installation of mirage-entropy-xen)

I don't really like the double compilation scheme, and this won't work very well with the current mirage tool implementation (could be changed though, but I'll try to minimise build time as much as possible).

I think the simpler and not too hacki-ish solution is to always install `mirage-entropy-xen` when using the xen mode for now on. And then use the detection scheme introduce in that PR to know if we need to link with the ocamlfind library or not.

samoht:
The only problem with that scheme is that if the user:
* starts from scratch
* configure with the `--no-opam` option
* installs the opam package manually

but I think it's fine if it is properly documented and if the mirage tool warns the user that using `--no-opam` might be a terrible idea and if we hard-code some well-known packages that should bring the linking options (ie. "tls").

hannesm:
I'm fine with that. mirage-entropy-xen is also rather lightweight and thus shouldn't harm too much to have compiled whenever someone chooses mirage on xen...

avsm:
> I think the simpler and not too hacki-ish solution is to always install `mirage-entropy-xen` when using the xen mode for now on. 

Agreed.  Some comments in the tool referencing why it's included by default (and possibly referencing this PR) would help us remember this whole discussion.

samoht:
An other approach is to add a virtual `mirage-noxen` package which will be in conflict with `mirage-xen`. And then use it to encode implication in the dependency constraints:

```
depends: [
...
  ("mirage-noxen" | ("mirage-xen" & "mirage-xen-entropy"))
]
```

Would be better if it is tracked directly by opam though, pinging @AltGr to comment about the possibility of having https://github.com/ocaml/opam/issues/2074#issuecomment-96608153 in opam 1.2.2 ...

pqwy:
I was also toying with the idea of just unconditionally installing `mirage-entropy-xen` as a quick fix for now.

This is what the patch now does.

Glad to squash the commits if your prefer that.

samoht:
Looks good to me. If we want to be extra careful we can also verify that the ocamlfind libraries that we are requiring are correctly installed on the system and if not print a warning about potentially missing initialisation function. Not sure it will happen in practice, but as we are providing the `--no-opam` option people can start making false assumptions ...

pqwy:
How would you add that? I'm a little reluctant to add early exits to functions that looks like they are computing extra deps..

lib/mirage.ml
```diff
@@ -26,6 +26,8 @@ module StringSet = struct
     List.iter (fun e -> s := add e !s) l;
     !s
 
+  let to_list s = fold (fun x xs -> x :: xs) s []
```
dbuenzli: This could be `let to_list = elements`.

samoht:
I'd do it at the end of the configuration phase, before emitting the preamble code.

lib/mirage.ml
dbuenzli: Actually it doesn't seem to be used by the PR (also my suggestion would be in fact the reverse of the current `to_list`).

pqwy: Forgot about it, thanks!

pqwy:
Updated. Can still squash the commits.

lib/mirage.ml
```diff
@@ -2027,6 +1939,62 @@ type tracing = Tracing.t
...
+  let linkage_ = ref None
+
+  let linkage () =
+    match !linkage_ with
+    | None   -> failwith "Nocrypto_entropy: `linkage` queried before `libraries`"
+    | Some x -> x
+
+  let packages () =
+    match !mode with
+    | `Xen -> SS.singleton "mirage-entropy-xen"
```
hannesm: should we be explicit about the version? (such as `>=0.3.0`)!? previous `mirage-entropy-xen` will not work...

pqwy: The idea was to handle that via opam in nocrypto?

samoht: putting the constraint in nocrypto should be fine.

lib/mirage_version.ml
```diff
@@ -0,0 +1 @@
+let current = "2.4.0"
```
samoht: this file should not be here

samoht:
Looks good to me. Ready to merge and release when the rest of libraries are ready.

hannesm:
since we will have `mirage-no-xen` anyways (for gmp-xen dependency in zarith), I'd put a conditional dependency in `nocrypto`: `("mirage-no-xen" | ("mirage-xen" & "mirage-entropy-xen"))`.
Than, we can remove the hardcoded dependency on `mirage-entropy-xen` within the mirage tool (in this PR).

pqwy:
@samoht The rest has been merged and tagged, with a PR waiting in `opam-dev`. This is the last blocker now.

PR #400, opened 2015-05-06
ARP module type; build ipv4 on top of arpv4 explicitly

yomimono:
mirage-types: Instead of hiding the ARP implementation within IPv4, split out a separate module type for ARP for modules (e.g., the ARP module in `mirage-tcpip`) to implement.

mirage: When building stacks, correctly build ipv4 on top of an ethernet impl and arpv4 impl.


yomimono:
Shouldn't be merged without https://github.com/mirage/mirage-tcpip/pull/134 or something like it.

lib/mirage.ml
```diff
@@ -917,6 +917,69 @@ let ethernet = Type ETHERNET
...
   impl ethernet network (module Ethif)
 
+module Arpv4 = struct
+  type t = {
+    ethernet: ethernet impl;
+    clock: clock impl;
+    time: time impl;
+  } (* is the type checker not inferring that this is an arp impl? I think this
+    is the root problem.  not sure how to tell it that?  how does it work for
+       other modules? *)
```
samoht: to answer the question in the comment: what you are defining here is how to configure an ARP implementation, ie. what parameters needs to be given to (i) the functor and (ii) the constructor function (ie, `create` or `connect` or ...)

yomimono: writing notes to yourself in potentially public places: cool, until it isn't.

lib/mirage.ml
```diff
@@ -917,6 +917,67 @@ let ethernet = Type ETHERNET
...
+      (Impl.module_name t.ethernet) 
+      (Impl.module_name t.clock)  
+      (Impl.module_name t.time);
+    newline_main ();
+    append_main "let %s () =" name;
+    append_main "   %s () >>= function" (Impl.name t.ethernet);
+    append_main "   | `Error _ -> %s" (driver_initialisation_error name);
+    append_main "   | `Ok eth  ->";
+    append_main "   %s.connect eth >>= function" mname;
+    append_main "   | `Error _ -> %s" (driver_initialisation_error "ARP");
```
samoht: there's a good opportunity here to improve the current situation by using `string_error` (or whatever here the name of the error printing function in `V1.DEVICE`) to pretty-print the error in case the configuration failed.

lib/mirage.ml
samoht: no worries, that's a legitimate question :-)

lib/mirage.ml
yomimono: Good point; I'll put in a commit for that shortly.

yomimono: ...actually it doesn't look like we require any such thing in `V1.DEVICE` at the moment.  `V1.FLOW` has such a `val error_message: error -> string`, but I don't see a similar requirement in any of the other module types.

samoht: ha yes, you are right. sorry for the noise then!

hannesm:
I'm still curious why/if we need to support multiple ip addresses within arp. shouldn't there be a single ip -- mac address mapping? also, I'd prefer a more pure interface without `add_ip`, `set_ips`, `remove_ip` etc., but this is likely not scope of this PR.

yomimono:
@hannesm - I'd like to keep the scope of this PR small, since it's apparently been enough of a challenge for me just to not break the build!

To address your point, I think a nicer way to support multiple IP addresses on a given interface would be to call Arpv4.connect multiple times on the same Ethif.t, then call Ip.connect on the respective Arpv4.t's.  It seems sensible to me for there to be separate caches per IP, even if they're on the same Ethif.t -- relatedly, I think a sensible ARP cache would disregard updates from unexpected networks.  Ipv4 has the same issue with its IP-binding interface, and I think it would be nice to remove the claimed multiple network support there as well, in favour of the expectation that users will construct separate Ipv4.t's, possibly with the same Ethif.t.

(As far as I know, there are currently no users of Ipv4/Arpv4's facility for multiple IP addresses in a single Ipv4.t or Arpv4.t .)

avsm:
> As far as I know, there are currently no users of Ipv4/Arpv4's facility for multiple IP addresses in a single Ipv4.t or Arpv4.t

That's my understanding as well.

samoht:
why why why? :-)

yomimono:
ugh, sorry

yomimono:
I'd like to clean up both this and the related PR in mirage-tcpip and resubmit soon, although I'll confess that the mess I just made in here was completely the result of carelessness :(

Issue #401, opened 2015-05-06
add a --no-depext option


avsm:
It would be useful to either add a no-depext option, or to let the generated Makefile skip it if an env variable is set.

samoht:
Related to https://github.com/ocaml/opam-depext/issues/15

samoht:
Fixed in 2.5.0

PR #402, opened 2015-05-13
Add --no-depext flag

talex5:
This is useful if you don't have root access, since otherwise mirage
tries to run sudo and fails.

Closes: #401

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,5 +1,3 @@
-export FORK_USER=dsheets
-export FORK_BRANCH=yorick
 export EXTRA_REMOTES=https://github.com/mirage/mirage-dev.git
 export PINS="mirage:. mirage-types:. conduit mirage-conduit"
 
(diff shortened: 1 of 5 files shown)


samoht:
Could you turn on the new option to build a new GCC compiler in Travis: `UPDATE_GCC_BINUTILS=1`

See https://github.com/ocaml/ocaml-travisci-skeleton/pull/39 (which should probably be documented somewhere ... @hannesm?)

talex5:
I've added the flag, but it seems to make no difference.

samoht:
```
export FORK_USER=dsheets
FORK_USER=dsheets
```
it still uses @dsheets fork, I think https://github.com/mirage/mirage/blob/master/.travis.yml#L3 should point to ocaml/master

talex5:
Ah, didn't spot that one! Fixed now.

samoht:
Thanks!

PR #408, opened 2015-05-29
Misleading "Compiling for target" message from `mirage build`

lnmx:
This is minor, but `mirage build` says "Compiling for target: Unix" after configuring for Xen:

```
$ mirage configure --xen
Mirage      Using scanned config file: config.ml
Mirage      Compiling for target: Xen
...
                                                                                                                               
$ mirage build
Mirage      Using scanned config file: config.ml
Mirage      Compiling for target: Unix
...
```

I thought the build had switched targets, but it's just `Mirage.load` printing the unset/default value of `Mirage.mode`.  It should only be printed during `configure`.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -2325,6 +2325,7 @@ let clean_main t =
 
 let configure t =
   info "%s %s" (blue_s "Using configuration:") (get_config_file ());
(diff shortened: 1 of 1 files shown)


samoht:
Thanks!

PR #415, opened 2015-06-10
Improve CHANGES to mention API changes in the combinators.

samoht:


samoht:
@avsm please review to see if that makes sense. I'll check on mirage-www if the contents needs to be updated.

CHANGES
```diff
@@ -1,8 +1,20 @@
...
+* Remove `Mirage.conduit_client` and `Mirage.conduit_server` types.
+* Change the type of the `Mirage.conduit_direct` combinator.
+  Previously, it took an optional `vchan` implementation, an optional
+  `tls` immplementation and an optional `stackv4` implemenation. Now,
+  it simply takes a `stackv4` implementation and a boolean to enable
+  or disable the `tls` stack. Users who want to continue to use
+  `vchan` with `conduit` should now use the `Vchan` functors inside
+  `unikernel.ml` instead of the combinators in `config.ml`.
+* Change the type of the `Mirage.http_server` combinator. The first argument
+  (the conduit server configuration) is removed and should now be provided
```
avsm: Could you move the http server one to the top; it affects almost everyone. An example of before after would be useful to make it really easy

CHANGES.md
```diff
@@ -1,18 +1,58 @@
...
+(* [config.ml] *)
+(* in 2.4 *) let http = http_server (`TCP (`Port 80)) conduit
+(* in 2.5 *) let http = http_server conduit
+
+(* [unikernel.ml] *)
+let start http =
+(* in 2.4 *) http (S.make ~conn_closed ~callback ())
+(* in 2.5 *) http (`TCP 80) (S.make ~conn_closed ~callback ())
+    ```
+
```
avsm: Thanks! This is very useful

PR #417, opened 2015-06-17
docs: clarify some expected behaviour for FS module type

yomimono:
Add some words to docstrings about the expected behavior of some functions in the FS module type.


samoht:
Thanks!

PR #419, opened 2015-06-19
Provide an ARP module type separate from IPV4.

yomimono:
Provide a module type for ARP.  Remove IPV4.input_arpv4 from IPV4 module type.  In the frontend tool, build stacks correctly reflecting this change.  (This is a rebase of https://github.com/mirage/mirage/pull/400 .)

In this model, users can create stacks with any ARP implementation which fulfills the module type, including (for example) an implementation which is completely static, rather than being bound to the implementation included in `mirage-tcpip`.

An update to `mirage-tcpip` which provides an arpv4 module satisfying this type is here: https://github.com/mirage/mirage-tcpip/pull/155 , and this PR shouldn't be merged without that one.


types/V1.mli
```diff
@@ -457,13 +457,49 @@ module type IP = sig
...
+
+  type ipaddr
+  type buffer
+
+  (** Type of the result of an ARP query.  One of `Ok Macaddr.t (for successful
+      queries) or `Timeout (for attempted queries that received no response). *)
+  type result = [ `Ok of Macaddr.t | `Timeout ]
+
+  (** Prettyprint cache contents *)
+  val prettyprint : t -> string
```
mor1: Minor point: Isn't there a growing idiom of calling these `pp` (and friends) rather than `prettyprint`?

dbuenzli: @mor1 if you call that `pp` then the signature should be `Format.formatter -> t -> unit`.

samoht: indeed, it would be better to have `val pp: Format.formatter -> t -> unit` here. I'll slowly deprecate all the other `t -> string` signatures.

yomimono:
pp returns unit io because a non-standard arp module I'd like to have fulfill the type needs this rather than unit.  If there are objections, I'm happy to change it.

samoht:
That's not ideal. The idea of having such a signature is that makes easier to compose them using `%a`:

```ocaml
Format.printf "The cache content is %a and it is looking good" Arp.pp arp
```

Also here, the function simply fill a buffer, which should not be a (potentially) blocking operation. So `unit` is better than `unit io` here.

samoht:
But maybe I misunderstood what you mean. Do you have an example on non-standard ARP module where this could be useful?

yomimono:
I'm thinking of irmin-arp, where reading the cache potentially blocks depending on the backend in use.

samoht:
Hmm, I see your point but I'm not totally happy with loosing the use of `%a` ... maybe @yallop would have a better idea?

Something a bit more complex would be:
```ocaml
val pp: t -> (Format.formatter -> unit) io

let ... =
  pp arp >>= Format.printf "haha %t hoho"
```

dbuenzli:
Another way could be to introduce an intermediate type. 

```
val to_repr : t -> repr io
val pp : Format.formatter -> repr -> unit
```

hannesm:
what is the status of this pr (and the related in tcpip)? I'm in favour of merging these rather sooner than later

samoht:
I'm happy to merge that as soon as it is ready.

yomimono:
I'll do a rebase tonight and fuss with Travis if necessary. If there are other requests for changes I'm happy to implement them, but I'm not planning on making further changes to this PR and would call it ready.

hannesm:
:100: great to hear, I think there aren't any requests for changes

yomimono:
Sorry for the delay; tests are passing here and [in mirage-tcpip](https://github.com/mirage/mirage-tcpip/pull/155) now.

PR #425, opened 2015-07-06
Mirage env hack

mor1:
@samoht as discussed earlier today, with reference to #360, here's the hack to enable some commonality of `mirage configure` environment interactions. Depends on http://github.com/mor1/mirage-env.


avsm:
Not for merge, will be superseded by #441

samoht:
Hopefully should be fixed by #441

PR #432, opened 2015-07-19
Support attaching files via a read-only tar-formatted BLOCK

djs55:
The unikernel `config.ml` can use something like `archive_of_files ~dir:"./htdocs" ()` and then:

- `mirage configure` will generate a tar archive
- the .xl config file will have a completed `disk` section
- the .xe script will have a disk upload section

All the device numbering stuff is hidden from the user.

Example changes: [mirage/mirage-skeleton#99]

Requires tar-format.0.4.0: [ocaml/opam-repository#4503] (now merged)

Minimal wiki changes + support for "FS=archive": [mirage/mirage-www#372]

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -628,6 +628,104 @@ let block = Type BLOCK
 let block_of_file filename =
   impl block filename (module Block)
 
+module Archive = struct
+
+  type t = {
+    block  : block impl;
+  }
+
+  let name t =
+    let key = "archive" ^ Impl.name t.block in
+    Name.of_key key ~base:"archive"
+
+  let module_name t =
+    String.capitalize (name t)
+
+  let packages t =
+    "tar-format"
+    :: Impl.packages t.block
+
+  let libraries t =
+    "tar.mirage"
+    :: Impl.libraries t.block
+
+  let configure t =
+    Impl.configure t.block;
+    append_main "module %s = Tar_mirage.Make_KV_RO(%s)"
+      (module_name t)
+      (Impl.module_name t.block);
+    newline_main ();
+    let name = name t in
+    append_main "let %s () =" name;
+    append_main "  %s () >>= function" (Impl.name t.block);
+    append_main "  | `Error _ -> %s" (driver_initialisation_error name);
+    append_main "  | `Ok dev  -> %s.connect dev" (module_name t);
+    newline_main ()
+
+  let clean t =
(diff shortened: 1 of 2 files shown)


samoht:
I've just push a fix for the failing Travis scripts, and I've restarted the jobs for this PR.

lib/mirage.ml
```diff
@@ -2100,6 +2234,17 @@ let configure_main_xe t =
...
+    append oc "echo Uploading data VDI %s" b.Block.filename;
+    append oc "echo VDI=$VDI";
+    append oc "SIZE=$(stat --format '%%s' %s/%s)" t.root b.Block.filename;
+    append oc "POOL=$(xe pool-list params=uuid --minimal)";
+    append oc "SR=$(xe pool-list uuid=$POOL params=default-SR --minimal)";
+    append oc "VDI=$(xe vdi-create type=user name-label='%s' virtual-size=$SIZE sr-uuid=$SR)" b.Block.filename;
+    append oc "xe vdi-import uuid=$VDI filename=%s/%s" t.root b.Block.filename;
+    append oc "VBD=$(xe vbd-create vm-uuid=$VM vdi-uuid=$VDI device=%d)" b.Block.number;
+    append oc "xe vbd-param-set uuid=$VBD other-config:owner=true";
+  ) (Hashtbl.fold (fun _ v acc -> v :: acc) all_blocks []);
```
mor1: Excuse my Xen toolchain naivety-- this when using `xe` only AIUI. Are there equivalent commands required for `xl` etc?

djs55: Have a look up at line 2191 -- that's where the xl config file is built. The main semantic difference is that `xe` uses the remote interface and uploads a copy of the disk, whereas the `xl` config file uses the disk in-place.

mor1: Yup, saw that you now built the `disk = ` line.
So if I understand correctly, `xl` using the disk image in-place means there's no equivalent required to the `xe` upload commands above. Ta!

djs55:
Hold off merging until I've made this work with a bigger example -- mirage-skeleton is a bit small.

djs55:
mirage-www is working nicely for me on both Unix and Xen with a bug fix in tar-format.0.4.1: [ocaml/opam-repository#4520]. I think this is safe to merge now.

avsm:
very nice -- worth advertising this on mirageos-devel

PR #435, opened 2015-07-27
Provide noop configuration for default_time

yomimono:
Without this patch, unikernels that attempt to take a module of type `V1_LWT.TIME` and a `time impl` will fail at time of `make`, because mirage generates a `main.ml` with invalid syntax.


samoht:
LGTM. I'll merge and make a release when Travis is happy (I've restarted the failing job).

PR #439, opened 2015-07-30
Fix block device naming on Xen when kv_ro_of_fs

samoht:
This should make mirage.io live again. I'm currently testing this, if this works, I'll make a bug-fix 2.6.1 release.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -645,8 +645,7 @@ let all_blocks = Hashtbl.create 7
 let dummy_filename = "xx"
 
 let block_of_file =
-  (* NB: reserve number 0 for the boot disk *)
-  let next_number = ref 1 in
+  let next_number = ref 0 in
   fun filename ->
     let b =
       if Hashtbl.mem all_blocks filename
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1009,7 +1009,7 @@ module Arpv4 = struct
     time: time impl;
   }
 
-  let name t = 
+  let name t =
     Name.of_key ("arpv4" ^ Impl.name t.ethernet) ~base:"arpv4"
 
   let module_name t =
@@ -1019,13 +1019,13 @@ module Arpv4 = struct
     "tcpip" :: Impl.packages t.time @ Impl.packages t.clock @ Impl.packages t.ethernet
 
   let libraries t =
-    let arp_library = 
+    let arp_library =
       match !mode with
       | `Unix | `MacOSX -> "tcpip.arpv4-unix"
       | `Xen  -> "tcpip.arpv4"
     in
-    arp_library :: 
-    Impl.libraries t.ethernet @ 
+    arp_library ::
+    Impl.libraries t.ethernet @
     Impl.libraries t.clock @
     Impl.libraries t.ethernet
 
@@ -1033,8 +1033,8 @@ module Arpv4 = struct
     let name = name t in
     let mname = module_name t in
     append_main "module %s = Arpv4.Make(%s)(%s)(%s)" mname
-      (Impl.module_name t.ethernet) 
-      (Impl.module_name t.clock)  
+      (Impl.module_name t.ethernet)
+      (Impl.module_name t.clock)
       (Impl.module_name t.time);
     newline_main ();
     append_main "let %s () =" name;
@@ -1056,8 +1056,8 @@ end
 type arpv4 = Arpv4
 let arpv4 = Type Arpv4
 
-let arp ?(clock = default_clock) ?(time = default_time) (eth : ethernet impl) = impl arpv4 
-    { Arpv4.ethernet = eth; 
+let arp ?(clock = default_clock) ?(time = default_time) (eth : ethernet impl) = impl arpv4
+    { Arpv4.ethernet = eth;
       clock;
       time
     } (module Arpv4)
@@ -1096,7 +1096,7 @@ module IPV4 = struct
     String.capitalize (name t)
 
   let packages t =
-    "tcpip" :: Impl.packages t.arpv4 
+    "tcpip" :: Impl.packages t.arpv4
 
   let libraries t  =
     (match !mode with
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -642,6 +642,8 @@ let block = Type BLOCK
 
 let all_blocks = Hashtbl.create 7
 
+let dummy_filename = "xx"
+
 let block_of_file =
   (* NB: reserve number 0 for the boot disk *)
   let next_number = ref 1 in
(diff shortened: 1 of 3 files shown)


samoht:
With the last commit https://mirage.io works again. @djs55 I'm not sure to understand the comment that I removed in that commit: why do you need to start numbering the block device from 1 instead of 0?

djs55:
It's useful to reserve disk 0 for the kernel, and start data from 1 since
many systems insist that the kernel is inside a disk (with grub etc). An
alternative is to use partitions but I thought whole extra disks is
probably simpler.

samoht:
But I don't understand how this could work. `mirage.io` seems to map `/dev/loopX` to the 0-th device, not the 1st one.

djs55:
I'm on a phone ATM, so can't browse the code easily, but does Mirage.io use
the generated .xl config file (with new block support in 2.6.0) or does it
have something hardcoded? The generated .xl config file should guarantee
the block device numbers match the mirage generated code

samoht:
Ha I see! it's using a hard-coded value: `"disk = [ '/dev/loopX,,xvda' ]" So should I be using `xvdb` instead? Does the autogen `xl` file should support the fat files? How does it guess which loopback device to use?

djs55:
In the autogen xl file it says something like

Target=path, format=raw, vdev=xvdb

And libxl magically creates a /dev/loop or fires up a tapdisk. A few xen
releases ago this probably didn't work very reliably, but I think it's good
now.

Let me take a look tomorrow (just got back)

samoht:
> In the autogen xl file it says something like

> Target=path, format=raw, vdev=xvdb

> And libxl magically creates a /dev/loop or fires up a tapdisk. A few xen
releases ago this probably didn't work very reliably, but I think it's good
now.

That sounds great!

> Let me take a look tomorrow (just got back)

cool thanks! The related code is https://github.com/mirage/mirage-www-deployment/blob/master/scripts/deploy-tls.sh#L53

djs55:
I think we could always fix this by updating the (global) checked-in .xl
file in the mirage-www-deployment repo but it seems inconsistent to check
in new kernels but keep only one VM config, when there are links between
the two. I think it would be cleaner to always check in the pair of
(kernel, VM config) and customise the VM config at runtime by adjusting
paths etc. I suppose the world would be cleaner if the VM config were
somehow part of the kernel itself, and you could just directly execute the
kernel (a bit like ocamlrun and bytecode)

How about the following plan:

* update the DEPLOY action of travis_mirage.ml to `git commit` the
generated .xl file as well as the kernel. The .xl file has the auto
generated disk configuration where the virtual device (xvda, xvdb) matches
the one compiled-in to the kernel.
* update the prepare-config.sh of mirage-www-deployments to customise only
the specific bits it case about: kernel path, disk path, memory, on_crash,
network instead of rewriting the file from scratch. This way in future we
can change how the VM boots (e.g. switch to booting HVM) without hopefully
having to change the code here too (maybe)

For reference, here's how the .xl files look now:
```
# Generated by Mirage (Sat, 1 Aug 2015 10:43:00 GMT).

name = 'block_test'
kernel = '<path to kernel>'
builder = 'linux'
memory = 256
on_crash = 'preserve'

disk = [ 'format=raw, vdev=xvdb, access=rw, target=<path to disk>' ]

# The network configuration is defined here:
# http://xenbits.xen.org/docs/4.3-testing/misc/xl-network-configuration.html
# An example would look like:
# vif = [ 'mac=c0:ff:ee:c0:ff:ee,bridge=br0' ]
```

What do you think?


On Fri, Jul 31, 2015 at 10:47 PM, Thomas Gazagnaire <
notifications@github.com> wrote:

> In the autogen xl file it says something like
>
> Target=path, format=raw, vdev=xvdb
>
> And libxl magically creates a /dev/loop or fires up a tapdisk. A few xen
> releases ago this probably didn't work very reliably, but I think it's good
> now.
>
> That sounds great!
>
> Let me take a look tomorrow (just got back)
>
> cool thanks! The related code is
> https://github.com/mirage/mirage-www-deployment/blob/master/scripts/deploy-tls.sh#L53
>
> —
> <https://github.com/mirage/mirage/pull/439#issuecomment-126820914>.
>



-- 
Dave Scott

talex5:
> I suppose the world would be cleaner if the VM config were somehow part of the kernel itself, and you could just directly execute the kernel (a bit like ocamlrun and bytecode)

I think these need to be separate: the kernel comes from the VM owner and can be changed by them freely, while the config comes from the sys-admin and defines the resources the kernel is permitted to use. You wouldn't want to let the VM owner control the config.

For tar images, perhaps we should add initrd support instead?

djs55:
Unfortunately the .xl file contains both
- the resources (disks, networks) the kernel is permitted to use
(controlled by the sys-admin I agree)
- configuration tied to the kernel code such as: PV versus HVM; and if HVM
stuff about the virtual hardware need to boot.

It's a shame the file isn't more cleanly separated into an 'internal, only
I should care' section and an 'external stuff I need to access, beware'
part.

I think we can choose between:
1. sysadmin writes their own config file from scratch; this affords maximum
control but it means they need to edit the config files occasionally when
mirage changes (e.g. if we drop PV mode for HVM or if we change the block
device numbering scheme :-)
2. sysadmin bases their config file on the mirage config by some kind of
#include/sed mechanism; this makes it easier to miss a config option (e.g.
if we add a new default that they don't actually want) but avoids the need
to track all our changes.

Unfortunately I think initrds are loaded into RAM, and the goal of using
the tar stuff is to read large datafiles from BLOCK devices directly, to
reduce memory overhead. I'd quite like to make mirage-www-deployments in
future use something other than crunch for all the data, so we can serve
more pictures and pdfs without worry :-)

The BLOCK device numbering change (starting from 1 rather than 0) was
motivated by wanting to support booting from disk images using pygrub and
perhaps real boot loaders in the longer term. It's not tar-specific,
although perhaps I should have made it a separate pull request!

On Sat, Aug 1, 2015 at 12:21 PM, Thomas Leonard <notifications@github.com>
wrote:

> I suppose the world would be cleaner if the VM config were somehow part of
> the kernel itself, and you could just directly execute the kernel (a bit
> like ocamlrun and bytecode)
>
> I think these need to be separate: the kernel comes from the VM owner and
> can be changed by them freely, while the config comes from the sys-admin
> and defines the resources the kernel is permitted to use. You wouldn't want
> to let the VM owner control the config.
>
> For tar images, perhaps we should add initrd support instead?
>
> —
> <https://github.com/mirage/mirage/pull/439#issuecomment-126903610>.
>



-- 
Dave Scott

samoht:
I'm happy to close this in favour of #440.

djs55:
OK, when #440 is in I'll try to update the deployment scripts to take advantage of it.

samoht:
Closing as this has been resolved more cleanly by @djs55

PR #440, opened 2015-08-02
Generate an .xl.in as well as an .xl file

djs55:
The .xl file is unchanged and is mostly useful as an example of a working .xl file (except the network configuration which is still missing). For example you can build `mirage-skeleton/block` and immediately run it with the generated `block_test.xl`.

The .xl.in file contains everything needed for the VM to boot (again except the network configuration) but with `@VARIABLES@` which must be substituted by the sysadmin according to their preferences. For example every required `BLOCK` device is present as a `disk`, with the correct `vdev` (matching the `Block.connect` inside the kernel) but with a variable `@BLOCK:filename@` which should be replaced with the full path to the expected block device file on the host.

Ideally deployments will base themselves on the .xl.in file, substituting the variables to suit local conditions. Then, if we change how mirage boots (e.g. PV to HVM), nothing on the deployment side needs to be changed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,12 @@
+* Xen: improve the .xl file generation. We now have
+  - `name.xl`: this has sensible defaults for everything including the
+    network bridges and should "just work" if used on the build box
+  - `name.xl.in`: this has all the settings needed to boot (e.g. presence of
+    block and network devices) but all the environmental dependencies are
+    represented by easily-substitutable variables. This file is intended for
+    production use: simply replace the variables for the paths, bridges, memory
+    sizes etc. and run `xl create` as before.
+
 ### 2.6.0 (2015-07-28)
 
 * Better ARP support. This needs `mirage-tcpip.2.6.0` (#419, by @yomimono)
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -2227,15 +2227,53 @@ let configure_main_libvirt_xml t =
 let clean_main_libvirt_xml t =
   remove (t.root / t.name ^ "_libvirt.xml")
 
-let configure_main_xl t =
+(* We generate an example .xl with common defaults, and a generic
+   .xl.in which has @VARIABLES@ which must be substituted by sed according
+   to the preferences of the system administrator. *)
+
+module Substitutions = struct
+  type v =
+    | Name
+    | Kernel
+    | Memory
+    | Block of Block.t
+
+  let string_of_v = function
+    | Name -> "@NAME@"
+    | Kernel -> "@KERNEL@"
+    | Memory -> "@MEMORY@"
+    | Block b -> Printf.sprintf "@BLOCK:%s@" b.Block.filename
+
+  type t = (v * string) list
+
+  let lookup ts v =
+    if List.mem_assoc v ts
+    then List.assoc v ts
+    else string_of_v v
+
+  let defaults t =
+    let blocks = List.map (fun b ->
+      Block b, Filename.concat t.root b.Block.filename
+    ) (Hashtbl.fold (fun _ v acc -> v :: acc) all_blocks []) in [
+      Name, t.name;
+      Kernel, Printf.sprintf "%s/mir-%s.xen" t.root t.name;
+      Memory, "256";
+    ] @ blocks
+end
+
+let configure_main_xl ?substitutions t =
+  let open Substitutions in
+  let substitutions = match substitutions with
+    | Some x -> x
+    | None -> defaults t in
   let file = t.root / t.name ^ ".xl" in
   let oc = open_out file in
   append oc "# %s" generated_by_mirage;
   newline oc;
-  append oc "name = '%s'" t.name;
-  append oc "kernel = '%s/mir-%s.xen'" t.root t.name;
+  append oc "name = '%s'" (lookup substitutions Name);
+  append oc "kernel = '%s'" (lookup substitutions Kernel);
   append oc "builder = 'linux'";
-  append oc "memory = 256";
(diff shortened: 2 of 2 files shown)


samoht:
Nice! Should we do something about the vifs as well?

djs55:
I think that's a good idea--  even if we don't know exactly which bridge to
connect them to, we do know that the VM expects the vif devices to exist.

samoht:
do you think automatically detecting `br0` vs. `xenbr0` running `ifconfig` is reliable? If think that's the only edit I do usually to my `.xl` files.

djs55:
I had a think about that and I think we can do something sensible. We now have the separation between:
- the .xl.in file: this is intended for production use where the deployment host is not the same as the build host and therefore you need to change the kernel path, bridge name etc by substituting the `@VARIABLES@`.
- the .xl file: this is intended for running the VM immediately on the build host. It has all the external dependencies filled-in already, based on values from the build host. Note we already hardcoded the kernel path *on the build host* so I think it's perfectly ok to take a peek at the network configuration of the build host and fill in a plausible bridge name too.

With this change I was able to build [mirage-skeleton/network] and run `xl create -c network.xl` and have it "just work" on my XenServer dom0 using "xenbr0" (apart from a misconfiguration of /etc/xen/xl.conf which is XenServer's problem, and I've added a comment about it to the .xl file in case anyone else needs to know)

samoht:
That sounds really great!!

samoht:
Can you update the CHANGE file and check that mirage-www is still up-to-date with the explanations? I'll cut a new release with the changes when that's ready.

djs55:
One of the travis jobs failed with
```
console     => opam install --yes mirage-console mirage-entropy-xen mirage-xen xen-evtchn xen-gnt xenstore
[NOTE] Package mirage-console is already installed (current version is 2.1.3).
The following dependencies couldn't be met:
  - mirage-xen -> mirage-xen-ocaml
  - mirage-xen -> ocaml-src
Your request can't be satisfied:
  - mirage-xen-ocaml is not available because your system doesn't comply with ocaml-version >= "4.01.0" & os = "linux" & ocaml-version < "4.02.2".
  - ocaml-src is not available because it requires OCaml >= 4.03.0.
```

samoht:
Seems that the ocaml version constraint for `mirage-xen-ocaml` is wrong. Should be `< "4.02.3"`

samoht:
Actually this should be fixed by https://github.com/ocaml/opam-repository/pull/4648

samoht:
job restarted

djs55:
Thanks!

djs55:
Building mirage-xen-2.2.1:
```
# +idir=/home/travis/.opam/system/include/mirage-xen/include
# +mkdir -p /home/travis/.opam/system/include/mirage-xen/include/caml
# +cd ocaml-src/byterun
# +PUBLIC_INCLUDES=alloc.h callback.h config.h custom.h fail.h hash.h intext.h   memory.h misc.h mlvalues.h printexc.h signals.h compatibility.h
# +sed -f ../tools/cleanup-header alloc.h
# sed: can't read alloc.h: No such file or directory
# make[3]: *** [install] Error 2
# make[2]: *** [build] Error 2
# make[1]: *** [install-runtime] Error 2
# make: *** [xen-install] Error 2
```

samoht:
@talex5 any idea what's going on here?

lnmx:
Likely related to https://github.com/ocaml/ocaml/commit/7ca29ef3f7ada4eefae4f6c58e77c92e6587ce69

samoht:
err, this means that mirage is broken on ocaml 4.02.3. Can someone try to fix mirage-platform?

talex5:
There is some initial support for OCaml 4.02.3 in `mirage-platform`'s `master` branch, but it's not in 2.3.3 (and anyway Travis is using mirage-xen 2.2.1).

samoht:
ha! so that's maybe just a missing version constraint...

djs55:
OK, so to fix travis on this repo (independent of this PR) I think we need:

- to release mirage-platform (2.3.4?) -- this will cope with the header move
- add an upper-bound on the OCaml version for all other mirage-platform versions (ocaml-version <= 4.01)

Is that right?

samoht:
The release is on its way: https://github.com/mirage/mirage-platform/pull/143 and yes I guess we need an upper-bound constraint somewhere :-)

samoht:
I've submitted PR for mirage-xen-ocaml 2.3.4 to opam, and the upper bounds seem to already be there. Not sure exactly why we see errors in mirage-dev, I'll investigate.

samoht:
travis is trying to install an old version of mirage (2.2.*) so I've added the missing constraints in https://github.com/ocaml/opam-repository/pull/4685

btw, the new travis console is completely broken that's *very* annoying...

djs55:
Yes, the console causes me to rerun lots of builds, since incremental
updates seem OK but downloading the whole lot is useless. I thought it was
just my browser...

samoht:
All green now!

djs55:
OK, this still works for me && travis is green so I think we can merge.

Drup:
Irk, rebasing this with https://github.com/mirage/mirage/pull/441 is going to be messy. :x

Drup:
Would it be possible to move all the xen file generation logic in a standalone library ? It would also allow to create a standalone tool that does this substitution.

djs55:
Oops -- sorry for the rebasing pain :(

djs55:
A standalone library is an interesting idea -- I'll look into that. It might be useful for other tools too, and as you say, it would be good to have a tool do the substitution rather than ad-hoc sed scripts.

PR #451, opened 2015-10-31
Fix OSX linking

samoht:
Seem that it went unoticed in previous versions of Mirage. The new version seems to be more strict so it exposes that bug. That's good.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -378,7 +378,7 @@ let ethernet_conf = object
   method name = "ethif"
   method module_name = "Ethif.Make"
   method packages = Key.pure ["tcpip"]
-  method libraries = Key.(if_ is_xen) [ "tcpip.ethif" ] [ "tcpip.ethif-unix" ]
+  method libraries = Key.pure ["tcpip.ethif"]
   method connect _ modname = function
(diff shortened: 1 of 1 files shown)


Drup:
So, what are the `-unix` packages for now ?

samoht:
They instantiate the functor for quick access by Unix (ie. non-mirage) consumers.

avsm:
What was the error message in the bug here?

samoht:
2 messages:

- A warning message saying that `ntetif.cmi` was appearing in 2 directories: `$lib/mirage-net-unix/` and `$lib/mirage-net-osx`:
    ```
findlib: [WARNING] Interface netif.cmi occurs in several directories: /Users/thomas/.opam/foo/lib/mirage-net-macosx, /Users/thomas/.opam/foo/lib/mirage-net-unix
    ```

- The usual link error saying that 2 different `Netif` implementations were being linked together:
    ```
File "_none_", line 1:
Error: Files main.cmx
       and /Users/thomas/.opam/foo/lib/mirage-net-macosx/mirage-net-macosx.cmxa
       make inconsistent assumptions over interface Netif
    ```

PR #453, opened 2015-10-31
Misc improvements

Drup:
Something I considered including: making keys (such as dhcp and net) generate an env variable by default. I feel slightly uneasy about this, but it seems to be rather common. What would you prefer ?

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -326,8 +326,8 @@ let kv_ro_of_fs x = kv_ro_of_fs_conf $ x
 
 (** generic kv_ro. *)
 
-let generic_kv_ro ?group dir =
-  match_impl Key.(value @@ kv_ro ?group ()) [
+let generic_kv_ro ?(key = Key.value @@ Key.kv_ro ()) dir =
+  match_impl key [
     `Fat    , kv_ro_of_fs @@ fat_of_files ~dir () ;
     `Archive, archive_of_files ~dir () ;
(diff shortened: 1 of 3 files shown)


Drup:
@samoht Opinion about having env variables by default ?

samoht:
I think that's good. Not sure if people use it, but that's useful for demos :p

Drup:
For everything ? including IP, all that ?

samoht:
yes I'd say for everything.

Drup:
Note that for demos, it doesn't matter, you can use the command line. It's mostly for scripts ... which should probably move to command line arguments.

samoht:
For demos, I want to be able to tweak some default values without having to add them on the command-line where I can forget to write the right thing. That's not super important though: @avsm what would you prefer/use in general?

hannesm:
I dislike environment variables for this -- although the command line gets shorter and looks more impressive, there's implicit information used, and I rather like it to be explicit.  If you think you forget stuff during a demo, either you get a warning from the unikernel (or you can use shell scripts or aliases to pass information implicitly).  Certainly, this is just my opinion ;)

mor1:
I agree with @hannesm on this -- I'd imagine I'm more likely to forget to set the right env var, or to tweak the right thing than I am to edit a command line where (as @hannesm says) at least everything is right there in front of your face :)

samoht:
ok let's forget about environment variables then :-)

Issue #456, opened 2015-11-02
Nocrypto autodetection is wrong


samoht:
See https://github.com/mirage/mirage-www/pull/402#issuecomment-152870284

PR #458, opened 2015-11-02
Add the correct dependencies for conduit/nocrypto.

Drup:
So, As far as I understand:
- conduit doesn't actually need entropy unconditionally (it only needs it with tls).
- the entropy device's dependencies were wrong.

@hannesm, @samoht, please review. It seems to work in practice, but the dependency graph for those packages is far from being clear.

Fix https://github.com/mirage/mirage/issues/456

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -766,8 +766,8 @@ let nocrypto = impl @@ object
 
     method packages =
       Key.(if_ is_xen)
-        [ "mirage-entropy-xen" ]
-        []
+        [ "nocrypto" ; "mirage-entropy-xen" ; "zarith-xen" ]
+        [ "nocrypto" ]
 
     method libraries =
       Key.(if_ is_xen)


hannesm:
@pqwy should review.

on xen, we need the opam package `mirage-entropy-xen` (and library `nocrypto-entropy-xen`, which is provided by nocrypto when built with mirage-entropy-xen and zarith-xen around),
on unix we need no additional package, but the library `nocrypto-entropy-lwt` (provided when nocrypto is built with lwt around)

Drup:
> on unix we need no additional package, but the library nocrypto-entropy-lwt (provided when nocrypto is built with lwt around)

I have no such package, only `nocrypto.lwt`.

hannesm:
right, sorry. library names are `nocrypto.xen` and `nocrypto.lwt`

Drup:
Ok, I'm confused now. Does conduit needs the entropy to be initialized when *not* using tls ? It seems like `conduit.mirage` has an (ocamlfind) dependency on `tls`.

Drup:
I added back the unconditional initialization of entropy when conduit is involved. I think it can occasionally be superfluous, but I can't magically detect how conduit as been compiled, so there is no choice.

mor1:
@drup when you say "can't magically detect..." do you mean you need to find whether `nocrypto.xen` or `nocrypto.lwt` (@hannesm @pqwy surely that should be `nocrypto.unix`?) is installed?

in which case, fwiw, the practice I've seen for this is to examine shell return code after executing the appropriate `ocamlfind query`...

Drup:
@mor1 It's not about nocrypto, it's about whether conduit has been compiled with or without tls support. The issue is that it's the wrong way around: In functoria, it's not possible to choose devices based on which libraries/packages are available, since the devices you choose determines what to install.

In this case, it means entropy will sometimes be initialized while it's not used. It's not *too* bad.

avsm:
A future conduit could still use the entropy in order to do random source port allocation. It's not a bad dependency to keep.

pqwy:
Here is the thing: someone might want to use cryptography regardless of `tls`.

`nocrypto` needs an entropy provider. Note that the coupling is loose; it does not *depend* on it, it will just bail out dynamically if there is no entropy. Entropy is (currently) provided by one of three sub-libraries: `nocrypto.xen`, which does CPU- and time-based magic, `nocrypto.unix`, which reads `/dev/urandom` *once*, and [the suspiciously named] `nocrypto.lwt`, which reads `/dev/urandom` *periodically*.

In the unikernel context, for `nocrypto` to properly work, you need to link to `nocrypto.xen` for Xen, or to `nocrypto.lwt` for Unix, and you need to call the initialization function. Unikernel authors could just link to the appropriate entropy provider and arrange initialization manually, but it seemed nicer if the "platform" could do that in a smart way.

To that end, the [last iteration](https://github.com/mirage/mirage/commit/cbfbd707de7988057316db0d9b18bbfeb91b7273) of this logic would detect if `nocrypto` was pulled in as a dependency, and add the additional entropy provider to the linkage list and a call to the init function to `main.ml`. There was nothing special about `conduit` or indeed `tls`, it was all conditioned on the presence of `nocrypto`, which is the actual piece of software that deals with entropy providers.

This meant that I could add a dep on `nocrypto` and start doing my PK crypto (which mostly requires RNG) without thinking about entropy, or that someone could add a dep on `tls` and just run it without considering entropy, or that `conduit` could be pulled in again without special entropy consideration, or any mix of those.

I believe it would be **awesome** if we could retain that logic, where simple usage of the crypto backend pulls in its corresponding entropy provider. `conduit` is not the place to deal with this.

mor1:
Retaining that ability seems reasonable -- I presume the decision is based
on explicit specification by the user, so there's no chance of ambiguity
whether to use Xen or Unix?

However, are there future situations where the platform should *not* do
this? Eg., because in a particular deployment some much better source of
entropy is available?

Also, at risk of narging about naming: what's the rationale behind
`nocrypto.unix` though?  It seems that a more intuitive naming scheme
should be possible (`.unix` and `.singleton` rather than `.lwt` and `.unix`
or something??)

On 3 November 2015 at 10:21, David Kaloper <notifications@github.com> wrote:

> Here is the thing: someone might want to use cryptography regardless of
> tls.
>
> nocrypto needs an entropy provider. Note that the coupling is loose; it
> does not *depend* on it, it will just bail out dynamically if there is no
> entropy. Entropy is (currently) provided by one of three sub-libraries:
> nocrypto.xen, which does CPU- and time-based magic, nocrypto.unix, which
> reads /dev/urandom *once*, and [the suspiciously named] nocrypto.lwt,
> which reads /dev/urandom *periodically*.
>
> In the unikernel context, for nocrypto to properly work, you need to link
> to nocrypto.xen for Xen, or to nocrypto.lwt for Unix, and you need to
> call the initialization function. Unikernel authors could just link to the
> appropriate entropy provider and arrange initialization manually, but it
> seemed nicer if the "platform" could do that in a smart way.
>
> To that end, the last iteration
> <https://github.com/mirage/mirage/commit/cbfbd707de7988057316db0d9b18bbfeb91b7273>
> of this logic would detect if nocrypto was pulled in as a dependency, and
> add the additional entropy provider to the linkage list and a call to the
> init function to main.ml. There was nothing special about conduit or
> indeed tls, it was all conditioned on the presence of nocrypto, which is
> the actual piece of software that deals with entropy providers.
>
> This meant that I could add a dep on nocrypto and start doing my PK
> crypto (which mostly requires RNG) without thinking about entropy, or that
> someone could add a dep on tls and just run it without considering
> entropy, or that conduit could be pulled in again without special entropy
> consideration, or any mix of those.
>
> I believe it would be *awesome* if we could retain that logic, where
> simple usage of the crypto backend pulls in its corresponding entropy
> provider. conduit is not the place to deal with this.
>
> —
> <https://github.com/mirage/mirage/pull/458#issuecomment-153307661>.
>



-- 
Richard Mortier
mort@cantab.net

talex5:
I think it's really confusing that entropy works differently to every other Mirage device. Now that we have functoria, we can support sharing properly (unlike the old version) so we should be able to make Nocrypto be a functor that accepts entropy and have mirage set everything up automatically I think.

Drup:
> [...] detect if nocrypto was pulled in as a dependency, and add the additional entropy provider to the linkage list and a call to the init function to main.ml [...]

As I said earlier, my issue is that it's the wrong way around: we can't choose devices depending on which libraries are used, since we use devices to figure out the libraries. Also, I would like to argue that ocamlfind libraries are really the wrong things to query. We should encode the dependencies at the level of devices. This way, we could choose (using flags) between different sources of entropy when it's relevant ....

In this context, the `tls` and `conduit` devices do need to care about the entropy device, since it's one of their dependencies.

Drup:
Note that this is irrelevant to the bug at hand: are the new ocamlfind dependencies for nocrypto correct ?

pqwy:
`libraries` looks correct.

`packages` seems overspecified. Does `Key.(if_ is_xen) [ "nocrypto" ; "mirage-entropy-xen" ] [ "nocrypto" ]` not suffice?

Drup:
@pqwy I followed the opam file, where the dependency is : `(mirage-no-xen | (mirage-xen & mirage-entropy-xen & zarith-xen))`.

I could remove lwt, since it's assumed in the `mirage` device anyway, but better safe than sorry. It's a depopts in nocrypto.

pqwy:
* `lwt` should be redundant.
* `mirage-xen` should be present if you're on xen anyway.
* `mirage-entropy-xen` is a proper dep, if on xen
*  `zarith-xen` is package-dep trick to force a different build of `zarith`. It would maybe be better if we didn't have that baked into the tool directly.

But yeah, none of it is incorrect. Some of it is just implied in any case.

pqwy:
`zarith` linkage errors. I guess I was wrong and it needs `zarith-xen` after all.

Drup:
I don't really understand the source of the linking failure. It was not there before ...

pqwy:
I have a nagging suspicion that the mysterious errors are related to [this](https://github.com/ocaml/opam-repository/commit/6c83f81e6a4c962b15ea521448d7c4ed56b61fc2), but not completely sure yet...

hannesm:
yes, that's very suspicious -- zarith-1.4 contains more symbols, and zarith-xen copies over `META` and `libzarith.a` into `lib/zarith`... we should at least depend on the very same zarith version in zarith-xen..

Drup:
That would explain things. Tell me when all of this is fixed in the opam repo.

hannesm:
fixed zarith in https://github.com/ocaml/opam-repository/pull/5081, awaiting merge...

samoht:
I've restarted the build

PR #465, opened 2015-11-09
Update Changes

Drup:
I was not really sure about what to put, so I mostly copied the breaking changes.
The blogpost doesn't have an URL just yet.


avsm:
Thanks!

PR #466, opened 2015-11-12
Restore previous tracing API

### LINKED ISSUE #460: Tracing doesn't work since functoria merge
Test-case (mirage-console + tracing):

    open Mirage

    let main =
      foreign "Unikernel.Main" (console @-> job)

    let () =
      let tracing = mprof_trace ~size:1000000 () in
      register "console" ~tracing [main $ default_console]

Gives:

    File "main.ml", line 43, characters 2-12:
    Error: This expression has type unit but an expression was expected of type
             'a Lwt.t

Also, the code seems to be emitted in the wrong order (tracing should be initialised before everything else, so you can trace the boot sequence and it sees the labels for the initial threads).

talex5:
Fixes #460.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -929,13 +929,15 @@ let argv_dynamic = if_impl Key.is_xen argv_xen argv_unix
 
 (** Tracing *)
 
-let tracing i =
+type tracing = job impl
+
+let mprof_trace ~size () =
   let unix_trace_file = "trace.ctf" in
-  let key = Key.tracing i in
+  let key = Key.tracing_size size in
   impl @@ object
     inherit base_configurable
     method ty = job
-    method name = "tracing"
+    method name = "mprof_trace"
     method module_name = "MProf"
     method keys = [ Key.abstract key ]
     method packages = Key.pure ["mirage-profile"]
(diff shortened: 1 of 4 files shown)


Drup:
I really don't like this code emission thing you made. I tried very hard to emit AST properly, `connect` is supposed to return an *expression*. This is not enforced by the type system (because we don't use metaocaml *yet*) but it is respected in the whole code base.

talex5:
If functoria provides a nice way to do this, feel free to change my code. However, the version in master currently generates code that is *broken* (does not compile) and would initialise in the wrong order even if it did.

talex5:
Will merge this today unless someone suggests a better solution, on the grounds that working code is better than elegant code.

Drup:
Yes, there is a way, by modifying the prelude, but I don't really have time to try it.

djs55:
Perhaps we should make an issue with a few notes on the problem so we don't
forget? We can then ask Amir to bring it up occasionally at the weekly
meeting for extra pressure ;-)

On Sun, Nov 29, 2015 at 4:46 PM, Gabriel Radanne <notifications@github.com>
wrote:

> Yes, there is a way, by modifying the prelude, but I don't really have
> time to try it.
>
> —
> <https://github.com/mirage/mirage/pull/466#issuecomment-160431400>.
>



-- 
Dave Scott

mor1:
Yes-- @drup if you could give some indication of how you'd go about it,
perhaps someone else will take it on.
Ultimately I don't want to hold up the functoria merge because it breaks
profiling-- and although, as @talex5 says, working code beats elegant code,
it would be nice to try and keep this more elegant than not!

On 29 November 2015 at 17:51, Dave Scott <notifications@github.com> wrote:

> Perhaps we should make an issue with a few notes on the problem so we don't
> forget? We can then ask Amir to bring it up occasionally at the weekly
> meeting for extra pressure ;-)
>
> On Sun, Nov 29, 2015 at 4:46 PM, Gabriel Radanne <notifications@github.com
> >
> wrote:
>
> > Yes, there is a way, by modifying the prelude, but I don't really have
> > time to try it.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/mirage/mirage/pull/466#issuecomment-160431400>.
> >
>
>
>
> --
> Dave Scott
>
> —
> <https://github.com/mirage/mirage/pull/466#issuecomment-160431654>.
>



-- 
Richard Mortier
mort@cantab.net

Drup:
That's the thing, I didn't tried, so I don't really know. The base idea is to make the [prelude](https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L1443) parametric, so it will be different if needs be. 

The issue here is not that it's inelegant: it breaks functoria's assumption about code emission. We can get away with it now, but I can't guarantee that it will stay that way.

Drup:
(But sure, let's not delay release for that)

mor1:
Merged. I'll create an issue with the notes from this thread to mark this as violating functoria's assumptions.

Issue #467, opened 2015-11-17
Xen default_console not found after functoria merge


talex5:
When I deploy a Xen unikernel (e.g. `skeleton/console`) built with functoria, I get:

`Console.connect console_xen_0: doesn't currently exist, waiting for hotplug`

Looks like

    Console_xen.connect "console_xen_0"

should be just

    Console_xen.connect "0"

PR #471, opened 2015-11-27
Pass the original name, not the prefixed name, to 'connect'.

yallop:
Fixes https://github.com/mirage/mirage/issues/467.  @drup, could you confirm that this is the right fix?


Drup:
It looks right, but you probably know the `console` API better than I do. Why is it not the same than Unix ?

yallop:
That's a good point.  I think both the Unix and Xen implementations need to be changed. The problem showed up with the Xen backend because there's a [special case for ""](https://github.com/mirage/mirage-console/blob/f334147120f971e9a407a624ee871aa7225b11f0/xen/console_xen.ml#L155-L158).

PR #475, opened 2015-12-03
Add LOGS signature

### LINKED ISSUE #156: mirage-types: Add printf-style logging functions
Most uses of `log` end up calling sprintf, so we might as well expose proper printf loggers.

talex5:
Many Mirage programs define their own signatures for logging. This adds a basic logging signature, based on the (currently unreleased) Logs library. This can probably close #156.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -916,3 +916,33 @@ module type KV_RO = sig
   (** Get the value size. *)
 
 end
+
+(** {1 Logging} *)
+
+type 'a log = ('a, Format.formatter, unit) format -> ('a -> unit) -> unit
+(** Type for a function that accepts log messages. *)
+
+module type LOG = sig
+  (** These functions each take a format string and a function that provides
+      the values. This function is called only if the log message will be
+      recorded (e.g. [Log.debug] does not call the function by default).
+      Example: [Log.info "Server listening on port %d" (fun f -> f port)] *)
+
+  val error : 'a log
+  (** Log an error. An error indicates a serious problem; a human should be
+      alerted that this has happened. *)
+
+  val warn : 'a log
+  (** Log a warning. This indicates a problem that should be fixed, but does
+      not require immediate attention. Warnings should be reviewed
+      periodically, and checked first when diagnosing a fault. *)
+
+  val info : 'a log
+  (** Log an informational message.
+      Such messages are recorded by default but do not generate alerts
+      and do not indicate that anything is wrong. *)
+
+  val debug : 'a log
+  (** Log a debug message. Such messages are discarded by default, but
+      debug-level logging can be enabled by the user when required. *)
+end


talex5:
OK, I've updated it to remove the reference to the (now released - thanks!) Logs library.

I believe we still need this signature however. By default, the Logs library silently discards all log messages, including warnings and errors, so a library using it directly would fail to report important messages for many users. Taking a logger as a functor argument reminds the user to do something. (If Logs provided a default logger that wrote to stderr then I think libraries could just use the Logs API directly.)

The signature cannot exactly match the Logs one because the tags type is abstract, but wrapping the Logs library should be trivial.

dbuenzli:
> I believe we still need this signature however. By default, the Logs library silently discards all log messages, including warnings and errors, so a library using it directly would fail to report important messages for many users. Taking a logger as a functor argument reminds the user to do something. (If Logs provided a default logger that wrote to stderr then I think libraries could just use the Logs API directly.)

So you add a signature to mirage types because you are not happy with the design of Logs... That's a curious way of designing the mirage system. Logs doesn't write to stderr by default because it is a *library* and it's up to the application to decide what it wants to do with its standard streams.

dbuenzli:
Btw. it's very easy to define your own `Setup_logs` module that installs a `Logs_stdo` reporter and that you simply link just after `Logs` in your unikernels to have the behaviour you are asking for.

dbuenzli:
In fact this could be provided under a `Logs_force` package by logs itself. This would let applications link against the `logs.force` package rather than `logs` to have the behaviour you want.

talex5:
There are two options here:

1. Libraries use Logs directly.
2. Libraries take a Mirage LOGS functor argument, which the application provides (probably via Logs).

If I modify e.g. `tcpip` to use Logs directly then users' unikernels will compile as before, but silently lose errors and warnings.

If I modify tcpip to use the LOGS signature, then users will be prompted to supply the module, which they can do using Logs (after configuring it). This way, errors and warnings will not be lost.

dbuenzli:
> If I modify e.g. tcpip to use Logs directly then users' unikernels will compile as before, but silently lose errors and warnings.

What about telling them to setup a reporter or to link against a module that does so ? Adding new signatures and functor arguments moves you away from the mirage.

samoht:
I really would like to avoid functorizing over LOGS if possible. The goal should be to use `Logs` as the standard log library for mirage: if not good enough for our uses, we send patches to Daniel to make that happens.

samoht:
The rational is that using logs is so pervasive that you will have to add a LOG parameter to everyone of your file. This is not very nice...

talex5:
Actually, we could have the functor take a `Logs.LOG`, rather than introducing our own signature, assuming it will be reasonably stable.

@samoht, @dbuenzli has already said he doesn't want to change the library in this way.

Anyway, if application authors need to take action then having their code fail to compile seems preferable to hoping they read the release notes for all their libraries on `opam upgrade`... especially since, if they don't, they won't immediately notice any problem.

dbuenzli:
> Anyway, if application authors need to take action then having their code fail to compile seems preferable to hoping they read the release notes for all their libraries on opam upgrade... especially since, if they don't, they won't immediately notice any problem.

I think you are a little bit blowing things out of proportions here. If each time application authors need to take action you need to have their code fail on compilation by introducing new signatures in mirage types and adding functor parameters you will quickly end up with a very ugly system. 

Besides their compilation will fail since they will need to link against `Logs` at which point you can simply direct them to link against a hypothetical `logs.force` package if you wish so. 

I also had a few other comments on this PR that misteriously disappeared but the main one being that functorizing logging in general will not encourage per source logging since instancers will likely simply thread the same functor argument in each subcomponents and hence fail at making meaningful log source distinctions which are better left to library designers.

hannesm:
my impression was, now that `Logs` is released, that mirage (the tool) might want to provide some reporter (and command-line arguments to silence / write to file / write to xen console)!?

(but this is just my opinion, would be great to have a log reporter in mirage (the command line tool) which (for now) enables logging to console as a start)..

Drup:
But the mirage tool only configures, why would it fiddle with runtime logging ?

Maybe what you want can be done simply with a device installing the custom logger with a key that tells where to redirect.

hannesm:
@Drup good idea! yes, it should actually be a distinct device, and mirage should drag it in (by default?!) as a dependency

talex5:
Another issue that needs sorting out is that Logs currently requires OCaml >= 4.02.0, while Mirage currently also supports 4.01.

> Besides their compilation will fail since they will need to link against Logs at which point you can simply direct them to link against a hypothetical logs.force package if you wish so. 

This will not be the case for applications that use ocamlfind and/or opam, which will have the extra dependency handled automatically.

@hannesm : I agree it would be good for the mirage tool to configure logging by default (but Mirage libraries are intended to be used outside of Mirage too, so this doesn't solve the problem of warnings being lost by default).

dbuenzli:
> (but Mirage libraries are intended to be used outside of Mirage too, so this doesn't solve the problem of warnings being lost by default).

Again... this is a non issue. As I have already said more than once in that thread, you can solve this by linking against another package that will link logs and your reporter before anything else. 

You are not making any point here: the actual "problem" happens at link time which is fully in control of the application developer or whatever framework if mirage wants to become a framework (uuuugh). So if you want to create a framework where nothing is lost by default you can perfectly do it. Besides if libraries play by the rules and never log before they are are explicitly initialized or called, nothing will be lost anyway if the application starts by setting up its reporter (either itself or through linking against a library/framework).

So instead of propagating FUD please do actually respond to the issues I raise about your completely dumb logging interface that will prevent meaningful log source specification by developers and thus prevent meaningful filtering for the end user. And this is not even getting in the discussion about the actual conceptual semantics of mirage-types... 

I do understand that you may be guided by down to earth technical problems but I don't think that mirage-types is the right place to solve these. That is unless you see mirage as being a framework and then I'm loosing my time here --- and the project wouldn't live by its name in my opinion.

PR #477, opened 2015-12-04
meta_ipv4 parenthesizes its output

yomimono:
If meta_ipv4 is called in a context where it may be used in a
constructor (e.g. Some meta_ipv4), failing to parenthesize the call to Ipaddr.V4.of_string_exn
results in a compilation error on a currified constructor.
Parenthesizing hurts nothing, so just do it by default.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -431,7 +431,7 @@ let meta_triple ppx ppy ppz =
   Fmt.parens @@ pp_triple ~sep:(Fmt.unit ",@ ") ppx ppy ppz
 
 let meta_ipv4 ppf s =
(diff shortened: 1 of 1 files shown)


Drup:
Indeed, did you spot other combinators that would need a similar fix ?

yomimono:
I discovered this by trying to build something that used a resolver and didn't look around much when going in to patch it.  I can take a gander, though.

Drup:
Yes, please do.

yomimono:
Scanned around and didn't see anything.  I tried compiling everything in `mirage-skeleton` and didn't run into any similar problems.

Drup:
Great, merging!

PR #482, opened 2016-01-12
Remove `nat-script.sh` from the scripts directory

avsm:
To be available as an external script since it is GPLv2 licensed.


djs55:
Remove looks good!

PR #483, opened 2016-01-18
typo/grammar edits in docstrings

yomimono:


types/V1.mli
```diff
@@ -482,7 +482,8 @@ module type ARP = sig
   val set_ips : t -> ipaddr list -> unit io
 
   (** [remove_ip arp ip] removes [ip] to the bound IP address list in
-      the [arp] value, which will xmit a GARP packet also. *)
+      the [arp] value, which will xmit a GARP packet for any remaining IPs in
```
Drup: "xmit" ?

yomimono: fair enough, changed

Drup: Oh, It's a real term, not just a typo ?

yomimono: It's a shorthand (see http://painintheenglish.com/case/1053 ), and I don't think saving ~20 bytes is worth even half a second of human confusion.

PR #484, opened 2016-01-28
xe: Start VM using UUID, not name-label

xen-git:
This makes the .xe script robust against a VM already existing with the same
name.

Signed-off-by: Si Beaumont <simon.beaumont@citrix.com>


xen-git:
Oops... logged in as wrong user, I'll close this and re-open.

PR #485, opened 2016-01-28
xe: Start VM using UUID, not name-label

simonjbeaumont:
This makes the .xe script robust against a VM already existing with the same
name.

Signed-off-by: Si Beaumont <simon.beaumont@citrix.com>


djs55:
Thanks for the patch!

djs55:
It looks like one of the travis jobs failed because of something unrelated in `mirage-profile`. The other 2 jobs passed and this patch looks safe to me. I'll merge.

simonjbeaumont:
It passed in #484 which I accidentally opened as @xen-git.

PR #495, opened 2016-02-22
Minor readme fix

mootpt:


djs55:
Thanks!
Travis has failed for a totally different reason, this patch is clearly safe to merge :)

Issue #493, opened 2016-02-18
Functoria: need a way to disable Bootvar


talex5:
Is there a way to disable the `Bootvar` parsing? Some platforms don't give you control of what kernel arguments you get, and if functoria can't parse them then the unikernel is unbootable. Qubes, for example, passes lots of arguments intended for Linux. As a work-around, I manually edit my `main.ml` after running `mirage` to:

    let argv_xen1 = lazy (return (`Ok [| "" |]))

Drup:
Hum, no, there isn't. I'm curious though, wouldn't argv be empty in this case ? If it's not the case, shouldn't bootvar-xen be fixed ? It doesn't seem very good to already disable bootvars during configuration.

Drup:
If you insist that being able to disable bootvar at configure time, it should be very easy to add one more argv device, that always returns `""` and hide the whole thing behind an `if_impl`.

talex5:
It doesn't like it being empty for some reason:

    MirageOS booting...
    Initialising timer interface
    Initialising console ... done.
    Netif: add resume hook
    gnttab_stubs.c: initialised mini-os gntmap
    qubes-firewall: internal error, uncaught exception:
		    Invalid_argument("argv array must have at least one element")
		    Raised at file "pervasives.ml", line 31, characters 25-45
    Mirage exiting with status 1
    Do_exit called!

Drup:
Oh, that is indeed a bug...

- Cmdliner removes the first element of argv (and returns this error when argv is empty).
- mirage-bootvar-xen doesn't always add the executable name at the beginning.

Which also means it's probably going to ignore the first options ....

@MagnusS Your opinion ? Should I add the executable name myself in functoria, or should it be done in bootvar ?

MagnusS:
Since we now have `argv` and `to_argv` in bootvar I think we should fix them to work as expected and always set first entry to kernel name or a placeholder.

Drup:
Actually, I just looked in more details, and we do add one empty string at the beginning of the array
@talex5 How did you triggered that ?

talex5:
If you configure with `--unix` it generates:

    OS.Env.argv () >>= (fun x -> Lwt.return (`Ok x))

`OS.Env.argv ()` is just `Lwt.return (Sys.argv)`, so it looks like it does expect the program name here.

You should be able to reproduce it with any unikernel by editing the `main.ml` to give an empty array and then running it.

Drup:
[...]

PR #497, opened 2016-02-22
Add the possibility to ignore the command line.

Drup:
Very rough fix for https://github.com/mirage/mirage/issues/493 for @talex5 and @sgrove 

It doesn't prevent the dependency on cmdliner, which is slightly annoying. It does prevent the dependency on bootvar. Is it worth trying to avoid the cmdliner dependency ?

However, if it turns out that most usage of mirage unikernels (EC2, Qubes ..) don't have precise control over their command line, we should find a more general solution than just disabling it.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -925,7 +925,18 @@ let argv_xen = impl @@ object
     method connect _ _ _ = "Bootvar.argv ()"
   end
 
-let argv_dynamic = if_impl Key.is_xen argv_xen argv_unix
+let argv_empty = impl @@ object
+    inherit base_configurable
+    method ty = Functoria_app.argv
+    method name = "argv_empty"
+    method module_name = "Mirage_runtime"
+    method connect _ _ _ = "Lwt.return (`Ok [|\"\"|])"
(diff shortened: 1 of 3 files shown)


talex5:
Good point. Qubes at least allows you to *add* additional arguments (but not get rid of the default ones). Perhaps we could pass some marker - e.g. get Bootvar to ignore everything before `--`?

Drup:
This should actually go in functoria directly, so don't merge this one.

Drup:
(It was decided this could be merged on the last two mirage calls, waiting transfer it into functoria)

hannesm:
the consent was to merge this here now, and later (once someone with the right skills (@drup?) has time) move the functionality to functoria :)

samoht:
To be fair, porting that to Functoria can be done by someone else than Drup... I can try to have a look tonight if no one else have time.

samoht:
I've done the first split of that patch, need to do the `dynamic_argv` bits. Probably not tonight for me.

samoht:
Closing in favour of #534

PR #501, opened 2016-02-28
Remove pin to functoria branch in .travi.travis.yml

samoht:


Drup:
We still need to pin it, since it's not in opam repository. I think it's best to pin it to the `mirage-dev` branch.

Drup:
@samoht Please do the same for functoria. :)

samoht:
See https://github.com/mirage/functoria/pull/51

PR #508, opened 2016-03-07
Update Readme

amirmc:
Edited initial paragraph. Commented out the other material and
instead pointed at the main website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,12 +28,24 @@ There are several diverse backends in MirageOS that require rather specialised
 build steps (from JavaScript to Xen unikernels), and this complexity is wrapped
 up in the tool.
 
-To work with Mirage, you'll need the following prerequisites installed:
+To work with `mirage`, you'll need to either install prerequisites
+locally or use the Docker image.
 
-* a working [OCaml](http://ocaml.org) compiler (4.02.1 or higher).
+**Local install**
+
+You will need the following:
+
+* a working [OCaml](http://ocaml.org) compiler (4.02.3 or higher).
 * the [OPAM](https://opam.ocaml.org) source package manager (1.2.0 or higher).
 * an x86\_64 or armel Linux host to compile Xen kernels, or FreeBSD, OpenBSD or
   MacOS X for the userlevel version.
+
+**Docker image**
+
+There is a maintained Docker image at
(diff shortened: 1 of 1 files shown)


amirmc:
I left a bunch of stuff commented out. Happy to remove it but wanted to check first.

README.md
```diff
@@ -21,16 +21,16 @@ available online.
...
 
-There are several diverse backends in Mirage that require rather specialised
+There are several diverse backends in MirageOS that require rather specialised
 build steps (from JavaScript to Xen unikernels), and this complexity is wrapped
 up in the tool.
 
 To work with Mirage, you'll need the following prerequisites installed:
 
-* a working [OCaml](http://ocaml.org) compiler (4.01.0 or higher).
+* a working [OCaml](http://ocaml.org) compiler (4.02.1 or higher).
```
avsm: 4.02.3

avsm:
a short thing about the `docker run -it unikernel/mirage` container would also be useful. I'll be maintaining that moving forward.

README.md
hannesm: 4.02.2 should be fine!? (that's the lower bound of ppx sexp)...

hannesm:
shouldn't docker instructions be kept in some (to-be-written) wiki/install-docker?  there shouldn't be the impression that MirageOS depends on docker...

amirmc:
I think it's fine to mention the Docker images since the aim is to maintain them.  If it makes it quicker to get started (for those who already have Docker) then it's worth highlighting them.

avsm:
> shouldn't docker instructions be kept in some (to-be-written) wiki/install-docker? 

just as a "quickstart" in the README is fine, along with manual install instructions.  The OPAM PPAs are getting increasingly difficult to maintain and these are the ones used in bulk builds, so most likely to work...

avsm:
re: 4.02.2 vs 4.02.3, again only the latest point release is tested, so best to recommend that as the lower bound

avsm:
also need to do a quick sweep for other compiler version deps -- there is a version check on opam in `mirage` but not sure about compiler version. @samoht may remember

samoht:
I've just added more checks to the tool (mirage-types, mirage-types-lwt, and OCaml version) so should be better. Merging now.

PR #510, opened 2016-03-12
Clean-up ocamlbuild tags

samoht:
Remove camlp4 tags, add debug and warnings.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1329,20 +1329,17 @@ let configure_makefile ~target ~root ~name info =
   append fmt "PKGS   = %s" packages;
   begin match target with
     | `Xen  ->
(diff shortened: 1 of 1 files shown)


lib/mirage.ml
```diff
@@ -1327,24 +1327,23 @@ let configure_makefile ~target ~root ~name info =
   newline fmt;
   append fmt "LIBS   = %s" libraries;
   append fmt "PKGS   = %s" packages;
+  let default_tags =
+    "warn_error(+1..49),warn(A-4-41-44),debug,bin_annot,\
+     strict_sequence,principal"
```
Drup: principal ? Really ? It doesn't play very well with disambiguation ...
Also, no safe string ?

PR #512, opened 2016-03-13
Make all projects depend on mirage-platform

### LINKED ISSUE #511: Unbound module OS
Seems that mirage's generated dependencies don't include mirage-platform:

`config.ml`:

    open Mirage
    let handler = foreign "Unikernel.Main" (clock @-> job)
    let () = register "test" [handler $ default_clock]

`unikernel.ml`:

    module Main (C:V1.CLOCK) = struct let start () = Lwt.return () end

Compiling gives:

    $ mirage configure --unix
    $ make
    File "main.ml", line 5, characters 0-11:
    Error: Unbound module OS

Looks like this dependency usually comes via `mirage-console`, but you might not have it if you use logging instead.

talex5:
Fixes #511.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1464,10 +1464,12 @@ module Project = struct
         let l = [ "lwt"; "sexplib" ] in
         Key.(if_ is_xen) ("mirage-xen" :: l) ("mirage-unix" :: l)
 
-      method libraries = Key.pure [
+      method libraries =
+        let l = [
           "lwt.syntax" ; "mirage.runtime" ;
           "mirage-types.lwt" ; "sexplib"
-        ]
+        ] in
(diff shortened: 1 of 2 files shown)


lib/mirage.ml
```diff
@@ -1464,10 +1464,12 @@ module Project = struct
         let l = [ "lwt"; "sexplib" ] in
         Key.(if_ is_xen) ("mirage-xen" :: l) ("mirage-unix" :: l)
 
-      method libraries = Key.pure [
+      method libraries =
+        let l = [
           "lwt.syntax" ; "mirage.runtime" ;
```
Drup: @samoht Did you kept this `lwt.syntax` on purpose ?

samoht: nope!

PR #514, opened 2016-03-14
Support early configuration of logging

talex5:
To get logging, wrap your job with `mirage_logs`, like this:

    open Mirage

    let handler = foreign
      "Unikernel.Main" (stackv4 @-> job)

    let stack = generic_stackv4 tap0

    let () =
      register "stackv4" [mirage_logs (handler $ stack)]

This requires a patch to functoria: https://github.com/talex5/functoria/tree/logging

It's currently a bit hacky, but I'm sure someone who understands functoria can make it nicer. The key idea is that a `configurable` now gets given its lazy inputs directly, rather than having functoria force them automatically. This allows us to set up logging before forcing the dependencies.

(However, if a dependency is also a dependency of logging then it will be forced first. For example, if you configure a network time source as the log time-stamper then networking will be initialised before logging. I think this is the desired behaviour.)

Note: This patch is on top of the `better-errors` branch and is not as large as it looks. Just look at the last commit.

/cc @Drup

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -949,6 +949,60 @@ let argv_xen = impl @@ object
 
 let argv_dynamic = if_impl Key.is_xen argv_xen argv_unix
 
+(** Logging *)
+
+type logger = LOGGER
+let logger = Type LOGGER
+
+let pp_level f = function
+  | `Error    -> Format.pp_print_string f "Logs.Error"
+  | `Warning  -> Format.pp_print_string f "Logs.Warning"
+  | `Info     -> Format.pp_print_string f "Logs.Info"
+  | `Debug    -> Format.pp_print_string f "Logs.Debug"
+
+let mirage_logs_conf =
+  impl @@ object
+    inherit base_configurable as super
+    method ty = clock @-> logger
+    method name = "mirage_logs"
+    method module_name = "Mirage_logs.Make"
+    method packages = Key.pure ["mirage-logs"]
+    method libraries = Key.pure ["mirage-logs"]
+
+    method connect _ modname _args =
+      Printf.sprintf "return (`Ok (%s.create ()))" modname
+  end
+
+let mirage_logs ?(clock=default_clock) () =
+  mirage_logs_conf $ clock
+
+let with_logger_conf level underlying =
+  impl @@ object
+    inherit base_configurable as super
+    method ty = logger @-> job
+    method name = "with_logger"
+    method module_name = "Logger"
+    method packages = Key.pure ["logs"]
+    method libraries = Key.pure ["logs"]
+    method deps = [abstract underlying]
+
+    method! connect_raw ~error ~names ~info ~modname fmt =
+      match names with
+      | [logger; underlying] ->
+          let logger = meta_init fmt logger |> meta_connect error fmt in
+          Fmt.pf fmt
+            "Logs.set_level (Some %a);@ \
+             %s.run %s @@@@ fun () ->@ "
+             pp_level level
+             modname logger;
+          let underlying = meta_init fmt underlying |> meta_connect error fmt in
+          Fmt.string fmt (super#connect info modname [underlying])
+      | _ -> failwith "Incorrect number of inputs!"
+  end
+
+let with_logger ?(level=`Info) logger underlying =
+  with_logger_conf level underlying $ logger
+
 (** Tracing *)
 
 type tracing = job impl
@@ -1468,6 +1522,7 @@ module Project = struct
   let version = Mirage_version.current
   let prelude =
     "open Lwt\n\
+     module Logger(X:sig type t val run: t -> (unit -> 'a Lwt.t) -> 'a Lwt.t end) = X\n\
      let run = OS.Main.run"
   let driver_error = driver_error
   let argv = argv_dynamic
@@ -1503,6 +1558,18 @@ end
 
 include Functoria_app.Make (Project)
 
+let in_parallel = function
+  | [singleton] -> singleton
+  | jobs ->
+    impl @@ object
+      inherit base_configurable
+      method ty = job
+      method name = "group"
+      method module_name = "Functoria_runtime"
+      method connect _ _mod _names = "Lwt.return (`Ok ())"
+      method deps = List.map abstract jobs
+    end
+
 (** {Deprecated functions} *)
 
 let get_mode () = Key.(get (get_base_context ()) target)
@@ -1517,12 +1584,14 @@ let add_to_opam_packages l =
 
 (** {Custom registration} *)
 
-let register ?tracing ?keys ?(libraries=[]) ?(packages=[]) name jobs =
+let default_log = with_logger (mirage_logs ()) ~level:`Info
+
+let register ?tracing ?(log=default_log) ?keys ?(libraries=[]) ?(packages=[]) name jobs =
   let libraries = !libraries_ref @ libraries in
   let packages = !packages_ref @ packages in
+  let jobs = log (in_parallel jobs) in
   let jobs = match tracing with
-    | None -> jobs
-    | Some tracing ->
-      tracing :: jobs
+    | None -> [jobs]
+    | Some tracing -> [tracing; jobs]
   in
   register ?keys ~libraries ~packages name jobs
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ install: wget https://raw.githubusercontent.com/ocaml/ocaml-travisci-skeleton/ma
 script: bash -ex .travis-ci.sh
 env:
   global:
-          - PINS="mirage-skeleton.dev:https://github.com/yomimono/mirage-skeleton.git#charrua-0.3 tcpip:https://github.com/mirage/mirage-tcpip.git mirage-types:https://github.com/mirage/mirage.git mirage:https://github.com/mirage/mirage.git"
+          - PINS="functoria mirage-skeleton.dev:https://github.com/yomimono/mirage-skeleton.git#charrua-0.3 tcpip:https://github.com/mirage/mirage-tcpip.git mirage-types:https://github.com/mirage/mirage.git mirage:https://github.com/mirage/mirage.git"
   matrix:
   - UPDATE_GCC_BINUTILS=1 OCAML_VERSION=4.02 PACKAGE=mirage
   - UPDATE_GCC_BINUTILS=1 OCAML_VERSION=4.02 PACKAGE=mirage-types
--- a/lib/mirage.mli
+++ b/lib/mirage.mli
@@ -42,6 +42,7 @@ val mprof_trace : size:int -> unit -> tracing
 
 val register :
   ?tracing:tracing ->
+  ?log:(job impl -> job impl) ->
   ?keys:Key.t list ->
   ?libraries:string list ->
   ?packages:string list -> string -> job impl list -> unit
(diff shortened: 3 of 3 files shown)


Drup:
Any reason you didn't tried to implement https://github.com/mirage/functoria/issues/45 instead ?

talex5:
@drup I don't see any concrete proposal there. Passing the lazy values directly seems more general than having a list of startup jobs, but if you have something working I'm happy to try it.

Drup:
> I don't see any concrete proposal there.

Add an extra declaration in [the argument of the functor](http://mirage.github.io/functoria/Functoria_app.S.html) that takes a list of jobs to run before the rest. [@yallop's proposal](https://github.com/mirage/mirage/issues/460#issuecomment-168101663) is a bit different. You add a `insert_let` function in the api that will create a new symbol at an appropriate placeholder. In our case, it will always be a toplevel declaration just before the current connect.

I'm quite strongly against introducing this `connect_raw`. For multiple reasons:

- It means you can override one fields with the other (connect/connect_raw). This is irregular (and a quite bad API, imho).
- You basically have to understand how everything is emitted to be able to write a device, and if anything, I want to move *away* from that. 
- @yallop or my proposals have cleaner APIs

A good first step, that you started, is to add a logging device. I still think that the fact that your API needs wrapping is a mistake and is not really consistent with the logs library. You should separate logging from wrapping and wrapping should be done in relation to the whole error reporting question.

talex5:
The logging is slightly different to tracing. For example, the default logger requires a `CLOCK` as input. So we don't necessarily want to start logging before *everything* else, just (typically) before everything that isn't a dependency of logging (we might also want network logging, although it would probably be better to start logging immediately to a ring buffer and then dump the ring to the network once that's available).

Could we have a functoria combinator `in_parallel` that runs a list of jobs in parallel? Then `register` could simply take a single job, and we could wrap multiple jobs with a single logger configuration.

Drup:
@talex5 Isn't it "the earlier the better" ? The question of organizing the order of devices in a tricky one. For now, given a set of devices, we can only add things that will load *after* them. The job initialization idea seems like a good first step towards a more expressive set of combinators that would allow a richer description of the initialization graph.

I have been considering the addition of a grouping feature, that takes a set of devices and consider them as one big device. In the meantime, `in_parallel` is easy to implement as a device with dependencies (take a list, put them all as dependency, the connect returns the tuple of the values).

talex5:
Earlier probably is better, but logging in static sequences will usually obviously fail to work, and there's no existing code to support. With tracing it's harder, because code isn't written explicitly to support tracing.

However, anything that works is fine by me :-) @samoht notes that you probably don't want to have two log reporters at once, so it might be better to make logging an argument to `register`. Also, that would allow us to have log reporting built in by default (though I'm not sure I like that).

talex5:
Note: on the [better-errors](https://github.com/talex5/functoria/tree/better-errors) branch I've improved the API slightly by removing `connect` from `configurable` and instead making it `private` on `base_configurable`. Also, that branch simplifies things further by removing the error handling code from functoria (since devices need to do their own error handling in order to report errors sensibly).

My previous `connect_raw` implementation just initialised logging and then called `super#connect_raw`, so it didn't need to know about the details. However, it only got away with that because its `CLOCK` dependency doesn't have instances.

I've now exposed `meta_init` and `meta_connect` so it can initialise things in the correct sequence:

    method! private connect_raw ~names ~info ~modname fmt =
      match names with
      | [clock; underlying] ->
          let _clock = meta_init fmt clock |> meta_connect fmt in
          Fmt.pf fmt
            "Logs.set_level (Some %a);@ \
             %s.(create () |> run) @@@@ fun () ->@ "
             pp_level level
             modname;
          let underlying = meta_init fmt underlying |> meta_connect fmt in
          Fmt.string fmt (super#connect info modname [underlying])
      | _ -> failwith "Incorrect number of inputs!"

which produces:

    let mirage_logs1 = lazy (
      let __clock1 = Lazy.force clock1 in
      __clock1 >>= fun _clock1 ->
      Logs.set_level (Some Logs.Info);
      Mirage_logs1.(create () |> run) @@ fun () ->
      let __f11 = Lazy.force f11 in
      __f11 >>= fun _f11 ->
      return (_f11)
      )

talex5:
I've updated this as suggested:

- In https://github.com/mirage/mirage-dev/pull/107#issuecomment-198536268 @samoht asked for logging to be enabled by default, with no changes needed to `config.ml`.

- In https://github.com/mirage/mirage/pull/514#issuecomment-196298813 @Drup suggested separating the error device from the wrapping. This allows the unikernel to get the logger object if it wants to interact with it further.

The current API is:

```
val register :
  ?tracing:tracing ->
  ?log:(job impl -> job impl) ->
  ?keys:Key.t list ->
  ?libraries:string list ->
  ?packages:string list -> string -> job impl list -> unit
(** [register name jobs] registers the application named by [name]
    which will executes the given [jobs].
    @param libraries The ocamlfind libraries needed by this module.
    @param packages The opam packages needed by this module.
    @param keys The keys related to this module.

    @param tracing Enable tracing.
    @param log Replace the default logger ({!mirage_logs}). To disable logging, pass the identity function.
*)

(** {2 Logging} *)

type logger
(** Abstract type for loggers. *)

val logger: logger typ

val mirage_logs: ?clock:clock impl -> unit -> logger impl
(** [mirage_logs ?clock ()] is a log reporter that prints log messages to the console,
    timestampted with [clock] (default: [default_clock]). *)

val with_logger:
  ?level:[`Error | `Warning | `Info | `Debug] ->
  logger impl ->
  job impl ->
  job impl
(** [with_logger ?level logger job] is a job that behaves like [job], but wrapped by
    [logger].
    It initialises the log level to [level] (default: [`Info]), connects
    [logger] and then connects [job] in a context provided by [logger].
    It avoids forcing [job]'s dependencies until logging is set up. *)
```

Currently there's no `LOG_REPORTER` type in `mirage-types`, so if you want to do something with the log device you have to define the signature yourself. We might want to standardise this in future.

However, this main benefit of this patch is that, once merged, all mirage libraries can start using logging. https://github.com/mirage/mirage-dev/pull/107 tracks PRs on other libraries adding this support.

hannesm:
What does `logging to be enabled by default, with no changes to config.ml` mean?  Is there be a way to turn off logging (I'd assume by not providing a logger via `?log`)?

On another note, since log reporter via syslog is on the horizon (see https://github.com/verbosemode/logs-syslog), will that need to modify the mirage tool again, or can we move the library-specific code into `mirage-logs` / `mirage-syslog`?  I believe providing a `LOG_REPORTER` signature would be good (and users can decide between no reporter, the console reporter, and the syslog one)!?

talex5:
To disable logging, you can use `~log:(fun x -> x)`, which wraps the jobs with the identity function only (the doc comment mentions this, but we could also provide a convenience value if desired, e.g. `~log:no_logger`).

There are various ways syslog could fit in here. You could use `~log:(with_logger syslog ...)` (assuming syslog provides a suitable implementation of the `syslog` device). Syslog can control the intialisation sequence with the new API. e.g.

1. Initialise clock.
2. Start collecting timestamped log entries to the console + ring buffer.
3. Initialise networking.
4. Send ring buffer contents to network service.
5. Initialise the unikernel, with logging going to the network.

Possibly syslog could use mirage-logs internally, registering itself as an additional output to the network. That would avoid duplicating the ring buffer, console and mirage-profile support.

talex5:
Any objections to merging this now? @Drup, are you happy with this at least as a temporary solution?

lib/mirage.ml
```diff
@@ -1468,6 +1522,7 @@ module Project = struct
   let version = Mirage_version.current
   let prelude =
     "open Lwt\n\
+     module Logger(X:sig type t val run: t -> (unit -> 'a Lwt.t) -> 'a Lwt.t end) = X\n\
```
Drup: Put that somewhere else, it has nothing to do here.

lib/mirage.ml
```diff
@@ -1503,6 +1558,18 @@ end
...
 include Functoria_app.Make (Project)
 
+let in_parallel = function
+  | [singleton] -> singleton
+  | jobs ->
+    impl @@ object
+      inherit base_configurable
+      method ty = job
+      method name = "group"
+      method module_name = "In_parallel"
```
Drup: It's preferable to give a module that at least exists, just in case. `Functoria_runtime` of `Pervasive`, for example.

hannesm:
since you mention "use mirage-logs internally", shouldn't it be like: mirage contains a device which buffers log messages to a ring buffer, and automatically initialises this very early on (hey, it does this with the RNG seeding as well).

at any point in time, a reporter (either console or syslog) can connect to Mirage, then the ring buffer is immediately drained and from then on logs go directly to the reporter.

this would also bypass initialisation problems (that first a clock and console is needed before there can be logs), wouldn't it?

it is likely only movement of code from mirage-logs to mirage.  I'd really like to be able to use a mirage unikernel without any console device.

lib/mirage.ml
```diff
@@ -1503,6 +1558,18 @@ end
...
+      method connect _ _mod _names = "Lwt.return (`Ok ())"
```
Drup: Don't you want to return the underlying values ?

hannesm:
(my comment is only nitpicking, and I'm fine with getting this merged, and do changes in that direction later, once there is actually a syslog reporter)

Drup:
I'm still of the opinion that the fact your logging API needs wrapping is wrong, and you never really answered that. `Log` was designed precisely so that you can have some initialization code somewhere that you can swap to something, and the API used in your program would be the exact same. 

tracing and/or argv is done that way, and it's much saner.

talex5:
@hannesm Having syslog extend mirage-logs is one possibilty, but I'm not sure we want to force this design. It also means that you can't configure early logging (e.g. maybe you want to log with micro-second precision or something). And, it forces you to depend on logging since a unikernel that doesn't want a log reporter will need to use a log reporter and then remember to add code to turn it off and discard the ring after boot. You still need to initialise the clock even if only logging to a ring buffer (although `default_clock` is rather trivial, indeed).

@Drup The wrapping is because if the program fails then we may want to dump out the full log ring buffer (e.g. show recent debug-level messages, that aren't shown by default).

Drup:
@talex5 then why do you expose individual wrapping ? Again, it conflates error and logging ...

lib/mirage.ml
talex5: I don't think so. They should all be unit. This is essentially just `Lwt.join`.

lib/mirage.ml
talex5: I'm not sure where to put it.

Maybe there's a better alternative. I don't really want to define a new functor. I just want to know the name of the logging module, so that I can call `run` on it. This identity functor seems a bit of a hack, but I didn't see a better way

talex5:
@Drup Whatever we do, something has to install the logger. `Logger.run main` seems more general than `Logger.install (); main ()`. If the logger doesn't need to wrap, it's just a tail-call and the effect is the same.

talex5:
I propose to merge this in its current state. While it would be tidier to avoid the extra line in the prologue of the generated main.ml, it's only cosmetic and doesn't seem easy to change:

* We could make further changes to functoria to expose the module types of the arguments passed to connect. However, is a more invasive change and would require changes to all the other drivers. Perhaps we could do it later in a new PR if desired?

* We could put the `Logger` helper in its own package. However, creating a whole opam package for a one-liner seems silly, especially if future changes to functoria make it unnecessary.

hannesm:
@talex5 shouldn't `Logger` be in mirage-types?

talex5:
@hannesm we might want a `LOGGER` type at some point too, but this `Logger` is something different (it's a helper functor used by `with_logger` to get the module type of its logger argument).

samoht:
Closing in favour of #534

PR #519, opened 2016-03-20
Fix regression in `opam depext` and in truncated opam output

samoht:
```
opam install mirage-types>=2.6.0
```
Seems to sometimes be interpreted as a shell redirection. Fix that by quoting the arguments of `opam install`. Ideally this should be fixed in functoria.


samoht:
Also it seems that `opam depext` is confused by the `>=` constraints. Bug reported upstream: https://github.com/ocaml/opam-depext/issues/47

Drup:
@samoht I though this kind of constraints where enforced with the conflict field in mirage ?

samoht:
True. I just wanted to be sure :-) as I sometime saw `mirage configure` try to remove the current version of mirage to install an older version of `mirage-types` ..

Drup:
@samoht did that happen since functoria and/or with a proper solver ? I know it used to happen.

samoht:
No I haven't seen it since a long time, so that should be fine.

Issue #520, opened 2016-03-20
Changing default tags to contain strict 'warn_error(+1..49)' breaks builds


ansiwen:
Since 2.7.1 the default tags contain "warn_error(+1..49)" which breaks builds that worked before (in my case a CI job for example) I appreciate strict warnings, but if they are enabled _and_ enforced in one iteration, there is no chance to fix warnings. At least there should be a warning and a hint how to change/overwrite the default tags. The only option I see at the moment is to overwrite the SYNTAX variable in a Makefile.user file.

PR #521, opened 2016-03-20
Remove warn_error from default tags

ansiwen:
Adding warn_error to the default tags for warnings that have been
enabled in the same step is a regression, since there has been no
inermediate period where this warnings could have been fixed.  This
change removes the warn_error tag from the default tags.

Fixes: #520


samoht:
Ok, I'll merge this and add a new flag `--warn-error` to turn on/off that behaviour. The default behaviour will be the same as it was for 2.7.0.

PR #523, opened 2016-03-23
Introduce a module type for ICMP and ICMPV4.

yomimono:
Currently, ICMP functionality is implicitly provided by the modules implementing IPV4 and IPV6 module types.  Define a module type for ICMP to facilitate pulling this functionality out of IPV4/6 in order to make it more explicit and extensible.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -522,6 +522,27 @@ module type IPV6 = sig
   include IP
 end
 
+(** {1 ICMP module} *)
+module type ICMP = sig
+  include DEVICE
+
+  type ipaddr
+  type buffer
+
+  val pp_of_error : Format.formatter -> error -> unit
+  (** pretty-print an error. *)
+
+  val input : t -> src:ipaddr -> dst:ipaddr -> buffer -> unit io
+(** [input t src dst buffer] reacts to the ICMP message in [buffer]. *)
+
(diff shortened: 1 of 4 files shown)


hannesm:
great!  the `PINS` in travis.yml can be cleaned up now (also in tcpip); also your charrua changes to skeleton can be merged since 0.3 was released (https://github.com/ocaml/opam-repository/pull/5990)

yomimono:
Oh nice, thanks! Missed the charrua-core release somehow even though I was looking out for it.

hannesm:
and there should be a changelog for mirage-types, so we track where we introduced new stuff and modified things in `V1.mli`/`V1_LWT.mli` (which should be renamed to `M.mli` / `M_LWT.mli`) :)

yomimono:
@hannesm - just submitted some PRs for 2.8.0 releases which include changelog entries.

types/V1.mli
```diff
@@ -522,6 +522,27 @@ module type IPV6 = sig
...
 end
 
+(** {1 ICMP module} *)
+module type ICMP = sig
+  include DEVICE
+
+  type ipaddr
+  type buffer
+
+  val pp_of_error : Format.formatter -> error -> unit
```
talex5: I think the normal convention is to call this `pp_error`. It prints as a side-effect, rather than converting from one thing to another.

PR #526, opened 2016-03-30
Followon #523: Icmp module type and multidistro CI

avsm:
/cc @yomimono -- testing multidistro CI with your ICMP PR


avsm:
Not sure what the CI failure is with mirage-minios-xen...

talex5:
Is there a docker container I can use to reproduce this?

avsm:
just `ocaml/opam` will give you a container that can repro it if you pin this branch and install it; see log at https://travis-ci.org/mirage/mirage/jobs/119752048

talex5:
It works for me in `ocaml/opam` with the same pins (though it shouldn't matter, since mini-os doesn't depend on any of them anyway).

But I'm not convinced I'm using the same versions of things as travis. e.g. the travis log shows 

    $ gcc --version
    gcc (Ubuntu 4.8.4-2ubuntu1~14.04) 4.8.4

while `ocaml/opam` shows:

    opam@d9d8c26b9702:~$ gcc --version
    gcc (Debian 4.9.2-10) 4.9.2

But then travis runs some "local-build" container. What's that?

hannesm:
this is now more alive in #590

PR #537, opened 2016-05-01
Build with ocamlbuild -r

### LINKED ISSUE #536: Hint: Recursive traversal of subdirectories was not enabled for this build
Every time a mirage build fails, we print this hint to the user:

```
Command exited with code 2.
Hint: Recursive traversal of subdirectories was not enabled for this build,
  as the working directory does not look like an ocamlbuild project (no
  '_tags' or 'myocamlbuild.ml' file). If you have modules in subdirectories,
  you should add the option "-r" or create an empty '_tags' file.
  
  To enable recursive traversal for some subdirectories only, you can use the
  following '_tags' file:
  
      true: -traverse
      <dir1> or <dir2>: traverse
```

I think we've got used to ignoring this, but for new users it must be confusing. We should find a way to suppress this message.

Perhaps mirage could generate an empty `_tags` file if there isn't one already?

talex5:
This is the default behaviour if you have a _tags or myocamlbuild.ml,
but not otherwise in case you're running from your home directory. In
our case, we know the current directory is a valid project and can
always have this on.

Avoids a long and misleading error that is otherwise printed after every
compilation failure telling the user to enable this.

Fixes #536.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ env:
   global:
   matrix:
   - PACKAGE=mirage UPDATE_GCC_BINUTILS=1 OCAML_VERSION=4.02 PINS="functoria mirage-logs"
-  -  PACKAGE=mirage-types UPDATE_GCC_BINUTILS=1 OCAML_VERSION=4.02
+  - PACKAGE=mirage-types UPDATE_GCC_BINUTILS=1 OCAML_VERSION=4.02 PINS="functoria"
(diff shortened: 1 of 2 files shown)


hannesm:
the `Unbound module Key_gen` happened to me when I used `functoria-1.1.0` and `mirage<2.9.0`... (which is constrained since https://github.com/ocaml/opam-repository/pull/6396)

PR #541, opened 2016-05-17
Require "pseudoheaders" function for IP module types.

yomimono:
Some transport layer protocols want to calculate a checksum over fields from the underlying network layer (the "pseudoheader").  Require modules to provide a `pseudoheader` function which assembles the correct pseudoheader given the parameters this should express.

Implementations of `pseudoheader` for IPv4 and IPv6 are included in a PR to `mirage-tcpip`.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -2,7 +2,7 @@ bash -ex .travis-opam.sh
 
 eval `opam config env`
 opam install mirage -y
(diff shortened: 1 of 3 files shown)


talex5:
LGTM (but needs releasing in sync with the new tcpip, of course)

yomimono:
Tests are failing because the current mirage-tcpip master breaks charrua-core.

PR #544, opened 2016-06-12
cleanups for generated Makefile: EXTRA_LD_FLAGS, link libgcc.a only on ARM

hannesm:


hannesm:
I've tested this to work with both unix and xen backends, and hope this is an uncontroversial change.  The libgcc.a linking was discussed in the commit https://github.com/mirage/mirage/commit/c17f2f60a6309322ba45cecb00a808f62f05cf82#commitcomment-17573123 ; the EXTRA_LD_FLAGS (and using a set instead of multiple occurences of `-L` flags) is mainly cosmetic atm (but opens the ability to easily integrate a byte usage visualisation into the toolchain - see https://hannes.nqsb.io/Posts/BottomUp).

It would be great if someone would review and merge this (travis is also happy, and now uses both 4.02 and 4.03!)

hannesm:
any opinion (@samoht @talex5 maybe?) :)

avsm:
Does this preclude using clang on ARM? Not sure if it worked before, but it definitely doesn't now :-)

avsm:
I have an Alpine 3.4 setup just installed on my rPi2, can built test this as soon as it's stable.

hannesm:
@avsm thx, clang was never supported, see https://github.com/mirage/mini-os/issues/1 as well (vanilla mini-os works with clang, ours does not)

hannesm:
rebased onto @yomimono pseudoheader merge; and separated the changes to .travis.yml into #545

yomimono:
This patch looks reasonable to me.

hannesm:
anyone against merging? @avsm

avsm:
hang on, just want to test it in CI first -- I've got the `docker run -it ocaml/opam:raspbian-8_ocaml-4.02.3` working now

hannesm:
@avsm it is still unclear to me what you want to test in the CI.  does your raspbian compile (and run) xen on arm?  since the x86 unix and xen parts work, I don't see (since there's no conditional code) how arm-unix could fail.  is there an estimation when your test runs will finish?

hannesm:
the `libgcc.a` linkage is done in solo5 by using `pkg-config` and the `LDLIBS` var... which maybe is more elegant than doing it deep inside the mirage tool... will certainly need a coordinated release of mirage and mirage-xen if we decide to do so (mirage-xen-posix provides some pkg-config files, and its `build.sh` script already contains conditionals for the platform it is compiled for, thus there's no added complexity [famous last words])

but in the end it does not matter too much, I don't see that we'll be able to generate `Makefile` rules which are independent of the platform and backend...

PR #546, opened 2016-06-16
Change source and destination ip and port names to be consistent.

yomimono:
It drives me a little mad that the labelling of source and destination addresses, and source and destination port numbers, is not consistent across the various function definitions in V1.mli .  This patch and the corresponding mirage-tcpip patch standardize on `src` and `dst` for source and destination addresses, and `src_port` and `dst_port` for source and destination ports.


hannesm:
I'm not a big fan of abbreviations.  Why not use `source` and `destination`?  IMHO the `get_` can be dropped.

yomimono:
I don't feel strongly about what the names are, just that I not have to remember that they're one thing in one place and another in another.  I chose `src` and `dst` to minimize the amount of API change, since there are more functions that refer to those than to `source` and `destination` or some other name.

mor1:
I would probably tend towards removing the `get_` as well.
AFAIK `src`/`dst` are pretty common (widely understood) abbreviations in this context so I've no problem with these are abbreviations. (And `destination` in particular takes up too many columns :)
I absolutely endorse rationalising the somewhat ad hoc naming conventions though!

Issue #203, opened 2014-01-28
V2 improvements:


avsm:
Keeping a list of improvements to module types in V2:

- [ ] Remove use of float in time-related operations, and switch to int64 instead.  This will make the kFreeBSD port easier and generally remove the dependency on floating point.

- [ ] Turn the stats gathering (e.g. in NETWORK) into an module type that can be included for every device.  Possible link to the RRD stats gathering daemon  (/cc @djs55)

- [ ] Checked error types (TODO: add mailing thread ref about this).

- [x] CONSOLE should satisfy FLOW (as well as log_*)

- [ ] Change non-polymorphic variant error in KV_RO into a polymorphic variant

- [ ] move error types outside of the the signatures and add 'with sexp' to every error type - add an error monad (@samoht)

- [ ] improve FS / add an Irmin module type

yomimono:
A number of these items are being carried forward to Mirage 3.0. Closing this in favor of that issue, which is tracked at #571.

PR #547, opened 2016-06-16
TIME: use an int64 for sleep

hannesm:
I got derailed while wanting to adapt the RANDOM module type with the TIME module type.

Its documentation:
```OCaml
  val sleep: float -> unit io
  (** [sleep nsec] Block the current thread for. {b FIXME:} remove float. *)
```

I propose to have: `val sleep : Int64 -> unit io`, the amount of nanoseconds (this is what we get out of the monotonic clock on [xen](https://github.com/mirage/mirage-platform/blob/ed19a1a318e944a7020239fa01c0b674b02cf471/bindings/clock_stubs.c#L39), and likely as well from Lwt_unix.sleep.  The range will be up to 9223372036854775807L nanoseconds (106751 days).

I've PRs for (compiles + test run with 4.03 on Unix (FreeBSD)):
- [mirage-platform 188](https://github.com/mirage/mirage-platform/pull/168) (mirage-xen and mirage-unix)
- [mirage-solo5 8](https://github.com/mirage/mirage-solo5/pull/8)
- [dns 95](https://github.com/mirage/ocaml-dns/pull/95)
- [tcpip 216](https://github.com/mirage/mirage-tcpip/pull/216)
- [mirage-skeleton 171](https://github.com/mirage/mirage-skeleton/pull/171)

anyone aware of other users of (OS.)Time.sleep (our website tutorial for sure)?  Any comments on this proposed API change?

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -73,8 +73,9 @@ module type TIME = sig
   type +'a io
   (** The type for potentially blocking I/O operation *)
 
-  val sleep: float -> unit io
-  (** [sleep nsec] Block the current thread for. {b FIXME:} remove float. *)
+  val sleep_ns: int64 -> unit io
+  (** [sleep_ns n] Block the current thread for [n] nanoseconds, treating
+      the [n] unsigned.  *)
 end
 
 (** {1 Random}
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -2,7 +2,7 @@ bash -ex .travis-opam.sh
 
 eval `opam config env`
 opam install mirage -y
-git clone -b mirage-dev https://github.com/mirage/mirage-skeleton.git
+git clone -b sleep-nsec https://github.com/hannesm/mirage-skeleton.git
 cd mirage-skeleton
 MODE=unix make
 MODE=xen  make
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ script: bash -ex .travis-ci.sh
 env:
   global:
   - UPDATE_GCC_BINUTILS=1
-  - EXTRA_REMOTES="https://github.com/mirage/mirage-dev.git"
+  - EXTRA_REMOTES="git://github.com/hannesm/mirage-dev.git#sleep-nsec"
   matrix:
   - PACKAGE=mirage OCAML_VERSION=4.03
   - PACKAGE=mirage-types OCAML_VERSION=4.03


types/V1.mli
```diff
@@ -51,8 +51,8 @@ module type TIME = sig
   type +'a io
   (** The type for potentially blocking I/O operation *)
 
-  val sleep: float -> unit io
-  (** [sleep nsec] Block the current thread for. {b FIXME:} remove float. *)
+  val sleep: int64 -> unit io
+  (** [sleep n] Block the current thread for n nanoseconds. *)
```
dbuenzli: Signed ? Unsigned ?

hannesm: signed (no feelings about it, but that's what I can rely on using Caml's Int64).

Behaviour for negative values is implementation-defined - I couldn't find the source for lwt, but [our xen implementation](https://github.com/mirage/mirage-platform/blob/ed19a1a318e944a7020239fa01c0b674b02cf471/xen/lib/time.ml#L72) skips a negative sleep.

dbuenzli: > Behaviour for negative values is implementation-defined

Please no, raise `Invalid_argument`.

hannesm: now raises in 70fc815 (implementation updated in https://github.com/hannesm/mirage-platform/tree/sleep-nsec)

yomimono:
Could we also provide convenience functions like `sleep_microseconds`, `sleep_milliseconds`, `sleep_seconds`?  I often want to sleep for durations like that.

hannesm:
I don't see any benefit.  If you need micro/milli/seconds often enough locally, you can write your own local functions.  If it is popular enough, develop a small (pure) library providing these functions.  If we embed it into `TIME`, every implementation (xen/unix/solo5) has to have the same boilerplate.  Also, you'll need to define what happens at the boundaries (would a `sleep_milliseconds` get an `int64`? should it then loop if it exceeds the native nanoseconds interface?  using an `int` would prevent it from having the same domain (you can express more sleep in the nanosecond case than in an `int` milliseconds one); and those cases should be handled uniformly by all backends (thus, as said, I'd prefer a simple and pure library).

I initially switched to milliseconds, but when I disovered that internally nanoseconds are used anyways, and everything is converted into a `int64`, We should not hide this from the user, but instead expose the raw interface to avoid unnecessary conversions.

talex5:
Hmm, this will break just about every Mirage program :-( And hiding the unit in the comment makes mistakes very likely (e.g. ms vs us vs ns).

What's wrong with float seconds here? Seconds are the SI units for time, which is what we want, and floats give more precision for small values, which is also what we want.

dbuenzli:
> And hiding the unit in the comment makes mistakes very likely (e.g. ms vs us vs ns).

+1

> What's wrong with float seconds here? Seconds are the SI units for time, which is what we want, and floats give more precision for small values, which is also what we want.

No, you will only be able to represent *seconds* up to 2^53 precisely. Most sub-second durations you may want to express will not be represented precisely.

talex5:
Well, `sleep` isn't precise in any case. It's only a lower-bound on the actual sleep time. And since time is continuous, there's no reason to think rounding to an integer will be an improvement (unlike with e.g. money where you want to represent 0.10 exactly but don't care about 0.101).

hannesm:
I'm happy to rename the `sleep` to `nanosleep` to avoid confusion about orders of magnitude.

I don't consider `Hmm, this will break just about every Mirage program :-(` to be a problem since we're talking about a new major release of something which user base is pretty small.

I don't think float should be exposed from any reasonable systems interface: they are imprecise, the underlying hardware deals with integers anyway, they are very tricky to reason about (since basic algebraic laws do not hold).  Also #203 has as goal to remove floating point numbers (make MirageOS more portable, there are various platforms without floating point operations out there), which is also happening in the revised CLOCK interface afaict.

I think that converting an amount (in the developer's head as milliseconds) to a floating point number (in seconds), just to convert it into an int64 (and scale it into nanoseconds) to perform it on a CPU sounds horrible to me.  Also, note that Unix provides a `sleep` in seconds and a `nanosleep` with a struct consisting of seconds and nanoseconds, both as integral numbers.

yomimono:
On 06/17/2016 12:00 PM, Hannes Mehnert wrote:

> I'm happy to rename the |sleep| to |nanosleep| to avoid confusion 
> about orders of magnitude.
>

I'd appreciate that.  The combination of that + a thin library for 
conversions sounds much less prone to developer confusion to me (your 
earlier points about extra boilerplate and sizes vs types being well taken).

> I don't consider |Hmm, this will break just about every Mirage program 
> :-(| to be a problem since we're talking about a new major release of 
> something which user base is pretty small.
>

I agree with this.  A major version release is a great time to break 
just about every Mirage program. ;)

hannesm:
this (and all its dependencies) now has `Time.nanosleep`

dbuenzli:
> this (and all its dependencies) now has Time.nanosleep

This is personal but my own conventions about this (which you have e.g. in `Ptime` and `Mtime`) is to suffix the function with the units using the SI abbreviations so I would rather call that `Time.sleep_ns`.

hannesm:
@dbuenzli point taken and renamed again. let's be consistent with the rest of the ecosystem if we make such a change.

yomimono:
Your timing is fortuitous; I was just thinking about merging this.  I'm going to submit PRs for the branches you made, unless you object, and merge this soon.

yomimono:
@hannesm, ping me when the merge conflict here is resolved and I'll start clicking green buttons where I have the power to do so.

hannesm:
@yomimono this is now rebased [as are the other PRs] on top of master, and using the EXTRA_REMOTE mirage-dev (in a branch hopefully pointing to the required things)

hannesm:
an alternative (or adjustment) would be to introduce a `type uint64 = int64` and get sleep to work with full 64 bit (needs a uint64 to float conversion for the unix backend, and further investigation for the xen backend).  Discussion about getting uint32/uint64 into Caml is ongoing in mantis http://caml.inria.fr/mantis/view.php?id=7280

since it would be a type alias, the clients would not need to change luckily :)
a helper library doing the conversions might be [mtime](https://github.com/dbuenzli/mtime) if it chooses to use uint64 instead of float.

thus, good to merge, but may be not final yet.

hannesm:
I provide a [conversion library](https://github.com/hannesm/duration), and rebased the PRs using this (tcpip/dns/mirage-skeleton).  I also decided to treat the int64 passed in to `sleep_ns` as unsigned.

PR #549, opened 2016-06-22
Minor API docs updates: use the present tense and active voice

djs55:
This standardises on the present tense and active voice (rather than e.g. the future tense). Most of the API was already like this, but it was inconsistent.

There's no implied semantic change in this PR (at least I hope not!)

This PR also starts a new section at the top called "General conventions" which calls out

1. errors which the application is supposed to handle (e.g. connection refused) are represented as error values
2. errors which the application is not supposed to handle (e.g. assertion failures) are represented as exceptions which are expected to abort the program with diagnostic data
3. operations which perform I/O return a `type +'a io` which allows the application to either wait for completion or let them run asynchronously.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -224,16 +224,18 @@ module type BLOCK = sig
   (** Query the characteristics of a specific block device *)
 
   val read: t -> int64 -> page_aligned_buffer list -> [ `Error of error | `Ok of unit ] io
-  (** [read device sector_start buffers] returns a blocking IO
-      operation which attempts to fill [buffers] with data starting at
-      [sector_start].  Each of [buffers] must be a whole number of
-      sectors in length. The list of buffers can be of any length. *)
+  (** [read device sector_start buffers] reads data starting at [sector_start]
+      from the block device into [buffers]. [Ok ()] means the buffers have been filled.
+      [Error _] indicates an I/O error has happened and some of the buffers may not be filled.
+      Each of elements in the list [buffers] must be a whole number of sectors in length.
+      The list of buffers can be of any length. *)
 
   val write: t -> int64 -> page_aligned_buffer list -> [ `Error of error | `Ok of unit ] io
-  (** [write device sector_start buffers] returns a blocking IO
-      operation which attempts to write the data contained within
-      [buffers] to [t] starting at [sector_start]. When the IO
-      operation completes then all writes have been persisted.
+  (** [write device sector_start buffers] writes data from [buffers]
+      onto the block device starting at [sector_start].
+      [Ok ()] means the contents of the buffers have been written.
+      [Error _] indicates a partial failure in which some of the writes may not have
+      happened.
 
       Once submitted, it is not possible to cancel a request and there
       is no timeout.
@@ -407,7 +409,7 @@ module type IP = sig
       function, or the [default] function for unknown IP protocols. *)
 
   val allocate_frame: t -> dst:ipaddr -> proto:[`ICMP | `TCP | `UDP] -> buffer * int
-  (** [allocate_frame t ~dst ~proto] retrurns a pair [(pkt, len)] such that
+  (** [allocate_frame t ~dst ~proto] returns a pair [(pkt, len)] such that
       [Cstruct.sub pkt 0 len] is the IP header (including the link layer part) of a
       packet going to [dst] for protocol [proto].  The space in [pkt] after the
       first [len] bytes can be used by the client. *)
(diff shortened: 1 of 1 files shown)


types/V1.mli
```diff
@@ -749,21 +751,21 @@ module type STACKV4 = sig
...
-  (** [listen_tcpv4 t ~port cb] will register the [cb] callback on the
+  (** [listen_tcpv4 t ~port cb] registers the [cb] callback on the
       TCPv4 [port] and immediatey return.  If [port] is invalid (not
       between 0 and 65535 inclusive), it raises [Invalid_argument].
       Multiple bindings to the same port will overwrite previous
       bindings, so callbacks will not chain if ports clash. *)
 
   val listen: t -> unit io
-  (** [listen t] will cause the stack to listen for traffic on the
+  (** [listen t] requests the stac listen for traffic on the
```
yallop: "requests the stac listen" ~> "requests **that** the stac**k** listen"

djs55: oops! Looks like I made that one worse

types/V1.mli
```diff
@@ -125,40 +125,38 @@ module type FLOW = sig
...
+      the remote return [`Eof].
 
       Note it is still possible for the remote endpoint to [write] to
       the flow and for the local endpoint to call [read]. This state where
       the local endpoint has called [close] but the remote endpoint
       has not called [close] is similar to that of a half-closed TCP
       connection or a Unix socket after [shutdown(SHUTDOWN_WRITE)].
 
-      The result [unit io] will become determined when the remote endpoint
+      The result [unit io] determines when the remote endpoint
```
talex5: I find this difficult to read. Maybe `completes` or `terminates`?

We should also decide what we mean by a function finishing. Do we mean when the OCaml function returns the thread, or when that thread finishes? It would probably be fine to say `[close flow] waits until the remote end has also called close before returning`.

djs55: Yeah I agree -- I think when I originally wrote the word "determined" I was thinking about Async terminology, but we don't use the word anywhere else. I'll try your suggestion.

types/V1.mli
```diff
@@ -125,40 +125,38 @@ module type FLOW = sig
       logging. *)
 
   val read: flow -> [`Ok of buffer | `Eof | `Error of error ] io
-  (** [read flow] will block until it either successfully reads a
-      segment of data from the current flow, receives an [Eof]
-      signifying that the connection is now closed, or an [Error]. *)
+  (** [read flow] reads some data from the flow and returns a
+      fresh buffer (of arbitrary size), an [`Eof] if the other side has
+      called [close] or an [`Error]. *)
```
hannesm: in the previous comment I read `block` and assume blocking behaviour if no data is there to be read; whereas the new does not specify the blocking semantics.  since basically all `io` functions are blocking, it might not be worth to mention it at every function, but in some general introduction at the top!?

djs55: Hm good point. Re-reading it again I think when I read the previous "successfully reads a segment of data" I imagine it returning more than a 0 byte segment. Perhaps my comment "(of arbitrary size)" makes it sound like 0 length is acceptable ... quite like non-blocking I/O on Unix? Perhaps I should clarify that the "arbitrary size" must be > 0 bytes which implies that the function must block until at least 1 byte is available?

djs55:
I added a "General conventions" section at the top which I hope isn't too controversial but will happily remove if so. I couldn't resist adding something about error handling there -- hope this is ok. It basically says that

- errors which should be handled (e.g. connection refused) are represented as error values
- errors which should not be handled (e.g. assertion failures) which should cause the program to abort with diagnostics are represented as exceptions
- blocking I/O operations are represented with `type +'a io` which allows the program to either wait for completion or leave them to run in the background.

WDYT?

hannesm:
I like it!

types/V1.mli
```diff
@@ -19,7 +19,27 @@
...
+    {1 General conventions}
+
+    {ul
+    {- errors which application programmers are expected to handle
+       (e.g. connection refused or end of file) are encoded as result types
+       where [`Ok x] means the operation was succesful and returns [x] and
+       where [`Error e] means the operation has failed with error [e]. }
+    {- errors which represent programming errors such as assertion failures
+       or illegal arguments are encoded as exceptions and the application
+       is expected to output diagnostics and abort. }
```
dbuenzli: This is a little bit unclear. I'd say that either the application tries to catch the exceptions and tries to recover (which may not be a good idea) or you let the whole thing crash and it should be the runtime's system duty to output diagnostics and abort.

types/V1.mli
```diff
@@ -19,7 +19,27 @@
...
+    {- operations which perform I/O return values of [type +'a io] which
+       allow the application to either wait for the I/O to be completed or
+       leave it running asyncnronously. If the I/O completes with an error
```
dbuenzli: asynchronously

djs55:
Thanks! The little green tick is telling me to merge...

djs55:
Oh I missed @dbuenzli's comment but at least I haven't merged yet :)

PR #550, opened 2016-06-28
[RFC] Add `disconnect` to `FLOW`

djs55:
`FLOW` already includes `close` which performs a graceful shutdown
of one half of a bidirectional connection (also known as a half-close
and written in Unix as `Unix.(shutdown fd Unix.SHUTDOWN_SEND)`).
For example in the Mirage tcpip stack, [close](https://github.com/mirage/mirage-tcpip/blob/29a5e068f6175207bb42965edd228d334157f56f/lib/tcp/flow.ml#L74) will send a FIN
and wait for the FIN ACK and then continue accepting data in the
other direction. This is important for protocols like HTTP where a
common pattern is to `connect`, `write` a request and then `close`
(a.k.a. `Unix.(shutdown fd SHUTDOWN_SEND)`), and then `read` the
response.

This patch requires a `disconnect` function which performs an
ungraceful shutdown of both halves of a bidirectional connection.
`disconnect` waits until all local resources have been cleaned up
(within limits imposed by the platform). Users who are happy to
let resources be freed in the background are welcome to call
`disconnect` asychronously.

Notes regarding implementations:
- in the Mirage tcpip stack, connection state is freed in the
  background as connections close and are timed-out. This extra
  function allows us to (i) request the state be cleaned up sooner;
  and (ii) optionally wait for the cleanup to be complete.
- in a Unix shim, `disconnect` should call `Unix.close` (or its
  equivalent) which will drop any buffered data rather than try to
  resend it, perform a FIN/ACK handshake and return. Unfortunately
  the Unix socket API is limited and it's not possible for us to
  wait while the kernel manages the connection in the "TIME-WAIT"
  state. This means it's possible even for careful applications
  to run out of kernel resources.

Currently it's not possible to write a TCP proxy with FLOWs
because of the confusion between half-close and full-close.

Although the terminology is different to Unix, I quite like the use of
`disconnect` here since it mirrors what we already do with `DEVICE`s
and is the opposite of `connect`.

An alternative would be:
- redefine `FLOW.close` to mean the same as `Unix.close` i.e.
  full close, drop incoming date (so one would `disconnect` a `BLOCK`
  but `close` a `FLOW`?)
- add a `FLOW.shutdown` to mean the same as `Unix.shutdown`
  to half-close the connection
- change `mirage-tcpip` to use both of these.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -898,6 +898,10 @@ module type CHANNEL = sig
   (** [close t] calls {!flush} and then close the underlying
       flow. *)
 
+  val disconnect: t -> unit io
+  (** [disconnect t] calls {!disconnect} on the underlying flow without
+      flushing first. *)
+
 end
 
 (** {1 Filesystem} *)
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -190,6 +190,22 @@ module type FLOW = sig
       before returning. At this point no data can flow in either direction
       and resources associated with the flow can be freed.
       *)
+
+  val disconnect: flow -> unit io
+  (** [disconnect flow] disconnects from the flow without flushing and
+      frees local resources.
+
+      Warning: [disconnect flow] may drop in-flight data: to ensure data is
+      transmitted use [close flow] first.
+
+      [disconnect flow] waits for the local resources to be freed. The time
+      taken depends on the underlying implementation; for example an implementation
+      over a Unix pipe can be freed quickly, while a connection to a
+      remote machine over TCP may require the connection to remain in a
+      "TIME-WAIT" state for several minutes, see
+      {{:http://www.isi.edu/touch/pubs/infocomm99/infocomm99-web/}The TIME-WAIT
+      state in TCP and Its Effect on Busy Servers}.
+  *)
 end
 
 (** {1 Console input/output} *)


avsm:
I'm in favour of remaining consistent with our use of `connect`/`disconnect` wrt state management

yomimono:
I'm in favor of this.

yomimono:
@djs55, thanks for the work you've been doing adding `disconnect` to our `FLOW` implementations.  :) I'm willing to merge this once we have a version of `tcpip.tcp` that respects it.  If nobody's done that by mid-August, I'll take it on.

djs55:
No problem, I'm quite motivated! I'm looking at `cohttp`/`conduit`/`mirage-http` at the moment to make sure we don't end up with a connection leak on Unix. Help definitely appreciated with `tcpip`! It seems to compile at the moment with no changes which is very suspicious...

I'm also building a `mirage-dev` branch -- I'll share that when I have `mirage-skeleton` building properly.

avsm:
@djs55, do we still need this for mirage3?

djs55:
I'm still keen on considering this: this would enable writing Unix socket-based TCP proxies over `FLOW`s without having to choose between either leaking file descriptors or dropping data. (note it's the sockets that are the problem -- pure Mirage unikernels are fine)

At the moment I have a workaround which is to use an extended flow interface. I can keep doing that, but it's slightly confusing having multiple `FLOW` variants lying around. Not the end of the world.

IIRC the main outstanding item was to consider `tcpip`. Perhaps the most minimal not-quite-compliant-but-good-enough implementation would be (very approximately)
```
let disconnect t =
  (* set some state so further read/writes get `Eof *)
  t.disconnected <- true;
  (* be nice by giving the remote a signal *)
  Lwt.async (close t);
  (* let the GC/ timers clean things up *)
  Lwt.return_unit
```
WDYT?

In summary: nice to have, but the world will not end if we don't get it :)

types/V1.mli
```diff
@@ -190,6 +190,22 @@ module type FLOW = sig
       before returning. At this point no data can flow in either direction
       and resources associated with the flow can be freed.
       *)
+
+  val disconnect: flow -> unit io
```
avsm: Why does this type need to be `unit io` and not just `unit`?  The substantive difference with `close` is that this does a dirty disconnect, and also the resources being freed asynchronously doesn't seem that bad.  The main objection to doing it asynchronously is that there won't be any global way to sync on (e.g.) not running out of file descriptors.  However, this global pool can be handled by the implementation of `type t` (that can track the total set of allocated fds), instead of individually at the `disconnect` function call level.

So I'm inclined to argue for `disconnect` to be a non-blocking `flow -> unit` function with asynchronous freeing of resources under all conditions.

types/V1.mli
```diff
@@ -882,6 +898,10 @@ module type CHANNEL = sig
   (** [close t] calls {!flush} and then close the underlying
       flow. *)
 
+  val disconnect: t -> unit io
+  (** [disconnect t] calls {!disconnect} on the underlying flow without
```
avsm: Minor: `[disconnect] calls disconnect` is not a very helpful docstring ;-)

types/V1.mli
djs55: My main objective is to clean up the resources rather than leak ... so I don't mind whether it's `unit io` or `unit`. I suggested `unit io` to offer callers the choice of waiting for the resource cleanup or not but `unit` would also work.

If we go for `unit`: would we require implementations of `connect` to catch any out-of-resources errors and wait for pending cleanups or would we pass the errors to the application and require the application to sleep and retry?

dinosaure:
We probably should close this PR. `mirage-tcpip` has the `disconnect` function as far I can say. @avsm can you confirm that we don't miss anything?

hannesm:
Thanks for this PR and the valuable discussion. mirage-types is no longer, but the feature in mind:
"be able to shutdown _and_ close a flow" has been reconsidered, and is currently proposed as https://github.com/mirage/mirage-flow/pull/48

NB: That proposal applies your other option, so the old close is now the Unix.close, and a new function shutdown is introduced. This avoids potential confusion due to using other names. Also, the connect/disconnect functions are those implemented by many interfaces to connect a stack (block device, ...), and disconnect it -- but there can only be a single disconnect in a module.

I hope that suits you all well, and please comment on the above linked mirage-flow PR if there's something unclear. I just noticed that we should likely update the documentation of `close` to reflect that it closes and frees all resources.

PR #551, opened 2016-06-29
[RFC] simplify random

hannesm:
this is only the interface change -- after discussion I'll adapt the mirage tool to do the magic.

- in contrast to common believe, there is no need for random to have blocking behaviour
- helper functions which generate *your favourite number representation here* between bounds will be done in a helper library (no need to convolute the interface with it)
- my current plan is to have two implementations: OCaml's `Random` and nocrypto's fortuna.  if anywhere in your dependency chain you have `nocrypto` already, that one is used as `default_random` (the [logic for this is already in the mirage tool](https://github.com/mirage/mirage/blob/074cb5fe080b23779bddeb24ddb59f24766fd5c0/lib/mirage.ml#L763-L781)).  you can always explicitly pass the [lagged-Fiobnacci PRNG](https://github.com/ocaml/ocaml/blob/4515477da8034cf21001ed3a6ae49b1a00a61121/stdlib/random.ml) ([seeded with](https://github.com/ocaml/ocaml/blob/4515477da8034cf21001ed3a6ae49b1a00a61121/byterun/sys.c#L412-L452) `/dev/urandom` or `gettimeofday` etc. -- fine for MirageOS on Unix, brittle on Xen (esp. on [ARM](https://github.com/talex5/mini-os/blob/444542b05e0f8a6220129b90f4697563d4bd0e1b/arch/arm/time.c#L75-L85) where there is no [wall clock](https://github.com/talex5/mini-os/blob/444542b05e0f8a6220129b90f4697563d4bd0e1b/arch/arm/time.c#L54-L56))) to your modules instead of the `default_random`.  Initialisation of any PRNG is done via its `connect` function (as atm for nocrypto, but not for `Random`).

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -113,6 +113,17 @@ end
 
 let stdlib_random = impl stdlib_random_conf
 
+let nocrypto_random_conf = object
+  inherit base_configurable
+  method ty = random
+  method name = "random"
+  method module_name = "Nocrypto.Rng"
+  method packages = Key.pure ["nocrypto"]
+  method libraries = Key.pure ["nocrypto"]
+end
+
+let nocrypto_random = impl nocrypto_random_conf
+
 type console = CONSOLE
 let console = Type CONSOLE
 
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -100,14 +100,18 @@ let default_monotonic_clock = impl monotonic_clock_conf
 type random = RANDOM
 let random = Type RANDOM
 
-let random_conf = object
+let stdlib_random_conf = object
   inherit base_configurable
   method ty = random
   method name = "random"
-  method module_name = "Random"
+  method module_name = "Stdlibrandom"
+  method packages = Key.pure ["mirage-stdlib-random"]
+  method libraries = Key.pure ["mirage-stdlib-random"]
+  method connect _ modname _args =
+    Printf.sprintf "Lwt.return (`Ok (%s.initialize ()))" modname
 end
 
-let default_random = impl random_conf
+let stdlib_random = impl stdlib_random_conf
 
 type console = CONSOLE
 let console = Type CONSOLE
@@ -655,7 +659,7 @@ end
 let tcp_direct_func () = impl (tcp_direct_conf ())
 
 let direct_tcp
-    ?(clock=default_monotonic_clock) ?(random=default_random) ?(time=default_time) ip =
+    ?(clock=default_monotonic_clock) ?(random=stdlib_random) ?(time=default_time) ip =
   tcp_direct_func () $ ip $ time $ clock $ random
 
 let tcpv4_socket_conf ipv4_key = object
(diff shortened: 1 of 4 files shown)


yomimono:
This seems like a reasonable change to me, including the implementation plan in your comment.

hannesm:
The first parts are done in here:
- `default_random` is gone (but there's a `stdlib_random` now)
- `Random.self_init ()` is called via `Functoria` in `connect` of the `stdlib_random` thingy
- There is a supporting library https://github.com/hannesm/randomconv
- There is a stdlib-random library https://github.com/hannesm/mirage-stdlib-random
- TCP/IP code is updated https://github.com/hannesm/mirage-tcpip/tree/random

I don't know of other users of the V1.RANDOM interface:
- in mirage-skeleton `ethifv4` uses the system provided `Random` directly
- in mirage-skeleton `lwt/src/unikernels.ml` use `Random` directly, not the one from mirage-types
- both are adapted in https://github.com/hannesm/mirage-skeleton/tree/random

There are still some libraries (tcpip tests, ocaml-dns, ipaddr) which manually use `Random.self_init ()` and thus the stdlib Random by default (some of which - such as ocaml-dns protocol.ml - are not inside of a mirage application).  Comments on how to deal with that is welcome.

Also, maybe @Drup has some advice how to get "either nocrypto or stdlib random" bound, depending on whether the unikernel already depends on nocrypto (sth to fill in here: `let direct_tcp ?(clock=default_monotonic_clock) ?(random=stdlib_random) ?(time=default_time) ip = tcp_direct_func () $ ip $ time $ clock $ random`) -- there's still some (trivial) boilerplate missing for a `nocrypto_random_conf`, similar to `stdlib_random_conf`.

avsm:
Should randomconv use ocaml-integers for the int16/etc types?

hannesm:
it predates ocaml-integers, and as mentioned in https://github.com/ocamllabs/ocaml-integers/issues/2 randomconv shouldn't need to depend on ocaml-integers.

avsm:
Makes sense!

hannesm:
and there's 548ce90 the nocrypto rng for your pleasure....

yallop:
> randomconv shouldn't need to depend on ocaml-integers.

That's true if you're using standard types (`int`, `int32`, `int64`), or very small types (`uint8`, `uint16`), but it's not true in other cases (`uint32`, `uint64`, and platform-specific types).

hannesm:
(and before we start to functorize all the code, maybe we should realise that a sensible OS will always have a single RNG, a single TIME, a single PCLOCK, and a single MCLOCK -- and there's no need to functorize all the libraries over these resources -- see https://github.com/pqwy/mirage-os-shim/ on how we'd get there)

hannesm:
@yallop this is the case, see https://hannesm.github.io/randomconv/doc/Randomconv.html

hannesm:
Travis is happy (in mirage-skeleton over [here](https://github.com/mirage/mirage-skeleton/pull/185))
changes:
  - this PR
  - https://github.com/mirage/mirage-tcpip/pull/253
  - https://github.com/mirage/mirage-skeleton/pull/184 (yes, 184! 185 (where travis is happy) is same as 184 but adds a mirage-dev remote which is not needed if the above two PRs are merged)

PR #553, opened 2016-07-08
add Qubes target

yomimono:
This PR adds an additional mode for Qubes.  If invoked with `-t qubes`, `mirage` will include in the generated `main.ml` some invocations of the libraries in [mirage-qubes](https://github.com/talex5/mirage-qubes) necessary to boot in that environment.

Much like the `MacOSX` target has a lot in common with `Unix`, `Qubes` is mostly the same as `Xen` with the exception of these extra invocations.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1246,6 +1246,7 @@ let mprof_trace ~size () =
       end
 
     method connect i _ _ = match Key.(get (Info.context i) target) with
+      | `Virtio | `Ukvm -> failwith  "tracing is not currently implemented for solo5 targets"
       | `Unix | `MacOSX ->
         Fmt.strf
           "Lwt.return ())@.\
@@ -1255,7 +1256,7 @@ let mprof_trace ~size () =
              MProf.Trace.Control.start trace_config@]"
           Key.serialize_call (Key.abstract key)
           unix_trace_file;
-      | `Xen | `Virtio | `Ukvm | `Qubes ->
+      | `Xen | `Qubes ->
         Fmt.strf
           "Lwt.return ())@.\
            let () = (@ \
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -32,11 +32,79 @@ let get_target i = Key.(get (Info.context i) target)
 
 (* Mirage implementation backing the target. *)
 let backend_predicate = function
-  | `Xen            -> "mirage_xen"
+  | `Xen | `Qubes   -> "mirage_xen"
   | `Virtio | `Ukvm -> "mirage_solo5"
   | `Unix | `MacOSX -> "mirage_unix"
 
 (** {2 Devices} *)
+let qrexec = job
+
+let qrexec_qubes = impl @@ object
+  inherit base_configurable
+  method ty = qrexec
+  val name = Name.ocamlify @@ "qrexec_"
+  method name = name
+  method module_name = "Qubes.RExec"
+  method packages = Key.pure ["mirage-qubes"]
+  method libraries = Key.pure ["mirage-qubes"]
+  method configure i =
+    match Key.(get (Info.context i) target) with
+    | `Qubes -> Result.Ok ()
+    | _ ->
+      Log.error "Qubes remote-exec invoked for non-Qubes target, stopping."
+  method connect _ modname _args =
+     Fmt.strf
+"@[<v 2>\
+         %s.connect ~domid:0 () >>= fun qrexec ->@ \
+         Lwt.async (fun () ->@ \
+         OS.Lifecycle.await_shutdown_request () >>= fun (`Poweroff | `Reboot ) ->@ \
+         %s.disconnect qrexec);@ \
+         Lwt.return (`Ok qrexec)@]"
+     modname modname
+end
+
+let gui = job
+
+let gui_qubes = impl @@ object
+  inherit base_configurable
+  method ty = gui
+  val name = Name.ocamlify @@ "gui"
+  method name = name
+  method module_name = "Qubes.GUI"
+  method packages = Key.pure ["mirage-qubes"]
+  method libraries = Key.pure ["mirage-qubes"]
+  method configure i =
+    match Key.(get (Info.context i) target) with
+    | `Qubes -> Result.Ok ()
+    | _ ->
+      Log.error "Qubes GUI invoked for non-Qubes target, stopping."
+  method connect _ modname _args =
+     Fmt.strf
+"@[<v 2>\
+         %s.connect ~domid:0 () >>= fun gui ->@ \
+         Lwt.async (fun () -> %s.listen gui);@ \
+         Lwt.return (`Ok gui)@]"
+     modname modname
+end
+
+type qubesdb = QUBES_DB
+let qubesdb = Type QUBES_DB
+
+let qubesdb_conf = object
+  inherit base_configurable
+  method ty = qubesdb
+  method name = "qubesdb"
+  method module_name = "Qubes.DB"
+  method libraries = Key.pure ["mirage-qubes"]
+  method packages = Key.pure [ "mirage-qubes" ]
+  method connect _ modname _args =
+     Fmt.strf
+"@[<v 2>\
+         %s.connect ~domid:0 ()@]"
+     modname
+end
+
+let default_qubesdb = impl qubesdb_conf
 
 type io_page = IO_PAGE
 let io_page = Type IO_PAGE
@@ -143,7 +211,7 @@ let nocrypto = impl @@ object
 
     method packages =
       Key.match_ Key.(value target) @@ function
-      | `Xen            -> ["nocrypto"; "zarith-xen"]
+      | `Xen | `Qubes   -> ["nocrypto"; "zarith-xen"]
       | `Virtio | `Ukvm -> ["nocrypto"; "zarith-freestanding"]
       | `Unix | `MacOSX -> ["nocrypto"]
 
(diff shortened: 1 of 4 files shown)


avsm:
fantastic -- so for build testing this a normal x86_64 linux is sufficient?

yomimono:
Yes, it should build fine on such a system.

On July 8, 2016 11:50:02 AM GMT+01:00, Anil Madhavapeddy <notifications@github.com> wrote:
>fantastic -- so for build testing this a normal x86_64 linux is
>sufficient?
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/mirage/mirage/pull/553#issuecomment-231331341

-- 
Composed on a touchscreen keyboard; please forgive typos and brevity.

yomimono:
A notable TODO here is network configuration.  Qubes doesn't, by default, provide a DHCP server on the bridge to which it attaches VMs, but rather expects the VMs to retrieve network configuration information from QubesDB (see @talex5's https://github.com/talex5/qubes-mirage-skeleton for an example).  I'm not sure how to implement this nicely in the current paradigm; I'd really like not to have to change anything in `config.ml` when testing on a vanilla Xen machine versus on my Qubes laptop.

hannesm:
I don't know much about Qubes startup, but your comments sound that adding a target qubes to the mirage tool will lead to any unikernels of the mirage-skeleton project to work flawlessly and out-of-the-box as qubes guests?  If this is the case, that is awesome!  I suspect since you mention network configuration, that some minor code (hidden in a match (which we have for unix and xen nowadays anyways to support direct/socket stacks)) changes are necessary, which could be upstreamed since there is another backend (qubes, next to xen)?

Or do I misunderstand the aims of this PR and what the idea of the qubes target is (which initially felt slightly wrong since it is just a xen)?

lib/mirage.ml
```diff
@@ -33,6 +33,83 @@ let driver_error name =
...
+  method connect i modname _args =
+  match Key.(get (Info.context i) target) with
+  | `Unix | `MacOSX | `Xen ->
+  Fmt.strf
+  "return (`Ok ())"
+  | `Qubes ->
+     Fmt.strf
+"@[<v 2>\
+         %s.connect ~domid:0 () >>= fun qrexec ->@ \
+         Lwt.async (fun () ->@ \
```
talex5: Not sure about this (disconnecting rexec whenever we receive a shutdown request).
Seems more like something the main unikernel application should decide about
(qubes-mirage-skeleton just happens to respond to rexec finishing by exiting).

lib/mirage.ml
```diff
@@ -33,6 +33,83 @@ let driver_error name =
...
+  val name = Name.ocamlify @@ "qrexec_"
+  method name = name
+  method module_name = "Qubes.RExec"
+  method packages = Key.pure ["mirage-qubes"]
+  method libraries = Key.pure ["mirage-qubes"]
+  method connect i modname _args =
+  match Key.(get (Info.context i) target) with
+  | `Unix | `MacOSX | `Xen ->
+  Fmt.strf
+  "return (`Ok ())"
```
talex5: We should probably report an error if you ask for rexec on a non-Qubes platform.

talex5:
I think we need to expose at least QubesDB and QRExec in the Mirage API (`mirage.mli`). Most Qubes unikernels will want access to these.
Networking can depend on `default_qubesdb` on Qubes by default, I think.

It's not clear what should call the `listen` functions though. e.g., if the unikernel wants to provide any qrexec command handlers then it needs to call listen itself. But a generic unikernel should get some default handlers.

There's a similar problem with shutdown events. If a unikernel doesn't provide any shutdown handler then it should probably just exit when requested. But if it does, we want to use that.

yomimono:
@hannesm - yes, you've got the right sense of it (including the network configuration targets).

@talex5 - it does seem like a solution which gives default handlers, but also allows the `impl` for QRExec and QubesDB to be passed to the unikernel, is what's really required.  In that case we do indeed need to expose these in the `mli`.  The frontend tool could then pass the initialized `impl`s along to the unikernel's `start`.  The `listen` functions are still a problem there -- they could be passed along in an optional argument from `config.ml` if the user wants them, I suppose.

lib/mirage.ml
```diff
@@ -33,6 +33,83 @@ let driver_error name =
...
+  | `Qubes ->
+     Fmt.strf
+"@[<v 2>\
+         %s.connect ~domid:0 () >>= fun gui ->@ \
+         Lwt.async (fun () -> %s.listen gui);@ \
+         Lwt.return (`Ok gui)"
```
Drup: You don't close your boxes. :)

Drup:
Did you forgot to commit the changes to `Mirage_key` ?

What should be the behavior of `Key.is_xen` ? For the Solo5 branch, all the occurrences of `Key.is_xen` have been removed. Given the proliferation of targets, maybe it's the best solution.

yomimono:
> Did you forgot to commit the changes to Mirage_key ?

I forgot to commit a *bunch* of changes, apparently.  I definitely had this building at one point but have either trashed the rest of the state or hidden it from myself extremely effectively, so I'm reconstructing it now. :/

yomimono:
Just like `MacOSX` is true for `is_unix`, I set `Qubes` to be true for `is_xen`.  In the case of Qubes it's pretty much true; Qubes literally is Xen with extra bits added.

lib/mirage.ml
Drup: I agree, the best is to use the `configure` method for that purpose (until we have a better solution, see https://github.com/mirage/functoria/issues/12). 

In any cases, having the two connect implementation that returns different types is a bad idea.

Drup:
Ok, the new part looks good to me. From what I can see, the old part mainly need proper failure in case of incompatibility.

yomimono:
Thanks very much for looking at it so closely! I'm writing a patch that moves the target matching logic to `configure` as you suggested.

yomimono:
This rebase includes commits that enable autoconfiguration of network stacks.  This was the last major TODO keeping `mirage configure -t qubes` from "just working" in most examples with no changes to `config.ml` or the application code.

However, most of @talex5's very reasonable comments upthread aren't quite addressed.  I'd like to merge this first and then submit another patch implementing the optional designation of custom listeners and handlers; I'd appreciate another look at this from anyone who's still interested.

lib/mirage.ml
```diff
@@ -870,19 +960,27 @@ let stackv4_socket_conf ?(group="") interfaces = impl @@ object
...
-    (socket_stackv4 ?group [Ipaddr.V4.any])
-    (if_impl Key.(pure ((=) true) $ dhcp_key)
-      (dhcp_stack ?group default_time tap)
-      (static_ipv4_stack ?config ?group tap))
+  if_impl Key.(pure ((=) `Qubes) $ value target)
+     (qubes_ipv4_stack ?group tap)
+    (if_impl Key.(pure ((=) `Socket) $ net_key)
+      (socket_stackv4 ?group [Ipaddr.V4.any])
+      (direct_autochooser ?group ?config ?dhcp_key tap)
+    )
```
hannesm: at least I'm slightly confused by the nested conditionals, which now call out to another function (only referenced once)... could we instead use a match_impl over several keys to make the code a bit clearer, please?

lib/mirage.ml
```diff
@@ -724,7 +810,7 @@ let udpv4_socket_conf ipv4_key = object
   method libraries =
     Key.match_ Key.(value target) @@ function
     | `Unix | `MacOSX -> [ "tcpip.udpv4-socket" ]
-    | `Xen | `Virtio | `Ukvm -> failwith "No socket implementation available for unikernel"
+    | `Xen | `Virtio | `Ukvm | `Qubes  -> failwith "No socket implementation available for unikernel"
```
hannesm: maybe if_unix?

lib/mirage.ml
```diff
@@ -771,7 +857,7 @@ let tcpv4_socket_conf ipv4_key = object
   method libraries =
     Key.match_ Key.(value target) @@ function
     | `Unix | `MacOSX -> [ "tcpip.tcpv4-socket" ]
-    | `Xen | `Virtio | `Ukvm  -> failwith "No socket implementation available for unikernel"
+    | `Xen | `Virtio | `Ukvm | `Qubes  -> failwith "No socket implementation available for unikernel"
```
hannesm: maybe if_unix

hannesm:
I added some notes (skipped over the qubes-specific code) -- LGTM!

lib/mirage.ml
yomimono: It is pretty confusing :( the other function is there to attempt to make it a bit more readable.  Using `match_impl` isn't very nice in this case because we want to consider several different values, and what goes in there isn't actually a pattern match.  I'll think it over a bit more.

yomimono:
Thanks @hannesm! :)

lib/mirage.ml
hannesm: it doesn't hurt to evaluate the keys, or does it?  if not, then do some bindings and use a standard match on them, maybe?

lib/mirage.ml
yomimono: It seems the behavior of `Key.match_` and a pattern is subtly different in the evaluation results from `Key.if_ is_unix` -- replacing this pattern match as suggested results in an immediate failure for any `mirage configure` with the `"No socket implementation"` message.  Possibly both branches of the `if` are evaluated in that case (and only the one which matches the actual value used), but this isn't done with `match_`?

lib/mirage.ml
```diff
@@ -959,27 +959,29 @@ let stackv4_socket_conf ?(group="") interfaces = impl @@ object
...
+  in
+  let p = Functoria_key.((pure choose)
+          $ eq `Qubes Key.(value target)
+          $ eq `Socket net_key
+          $ eq true dhcp_key) in
+  match_impl p [
+    `Dhcp, dhcp_ipv4_stack ?group tap;
+    `Socket, socket_stackv4 ?group [Ipaddr.V4.any];
+    `Qubes, qubes_ipv4_stack ?group tap;
+  ] ~default:(static_ipv4_stack ?config ?group tap)
```
hannesm: this looks _much_ better! thanks!

yomimono:
I'd like to remove it, yes; feel free to PR if you get to it first.

PR #559, opened 2016-07-14
[Tracking] Better errors

talex5:
This is the improved error handling described in http://canopy.mirage.io/Posts/Errors

It allows each device to handle errors in a way suited to that device (which could be using a `pp_error` function, but other options are available too - in particular, some devices never fail and so no error handling is needed).

I've updated it to remove the old logging support (since a different system got merged in the end). Haven't tested the update much. Might need a few tweaks to get https://github.com/mirage/mirage-dev/pull/107 passing again.

Depends on https://github.com/mirage/functoria/pull/55, which I've also updated.

(not very happy with the `die` function - should probably fail, and catch in the main project instead)

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -217,8 +217,11 @@ class block_conf file =
          else (1 lsl 28)  lor (b.number lsl 8)) |> string_of_int
 
     method connect i s _ =
-      Printf.sprintf "%s.connect %S" s
-        (self#connect_name (get_target i) @@ Info.root i)
+      Format.sprintf "@[<v 2>%s.connect %S >|= function@ \
+                      | `Error e -> die \"@@[<v 2>Error connecting block device %%S:@@ %%a@@]\" %S Block.pp_error e@ \
+                      | `Ok _ as ok -> ok@]" s
+      (self#connect_name (get_target i) @@ Info.root i)
+      file
 
   end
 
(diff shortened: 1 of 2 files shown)


samoht:
Can we switch to `Result.result` as well instead of a custom error type?

talex5:
Yes. The main goal here is to make sure that we can change one device at a time, rather than requiring everything to change at once.

lib/mirage.ml
```diff
@@ -1525,6 +1570,9 @@ module Project = struct
   let version = Mirage_version.current
   let prelude =
     "open Lwt\n\
+     let die fmt =\n\
+       Format.kfprintf (fun _ -> exit 1) Format.err_formatter\n\
+         (\"*** Setup failure ***@\\n@\\n\" ^^ fmt ^^ \"@.\")\n\
```
Drup: Shouldn't `die` use the logger ?

talex5: Maybe. It would be unfortunate if the error occurred before logging was configured though. Then the error wouldn't appear at all.

avsm:
This one one doesn't build against latest functoria due to missing `get_base_context`

hannesm:
superseeded by #602, can we get the mirage-types changes separately (including `pp_error` and `type error = private [>`)?

Issue #461, opened 2015-11-04
Errors in config.ml result in unhelpful error message


talex5:
If `config.ml` doesn't compile then you get an error about unknown options instead. e.g.

    $ cat config.ml 
    foo

    $ mirage configure --unix
    mirage: unknown option `--unix'.
    Usage: mirage configure [OPTION]... 
    Try `mirage configure --help' or `mirage --help' for more information.

Running without arguments shows the real error:

    $ mirage configure
    [ERROR]
      
    ocamlfind ocamldep -package mirage -modules config.ml > config.ml.depends
    ocamlfind ocamlc -c -annot -bin-annot -package mirage -o config.cmo config.ml
    + ocamlfind ocamlc -c -annot -bin-annot -package mirage -o config.cmo config.ml
    File "config.ml", line 1, characters 0-3:
    Error: Unbound value foo
    Command exited with code 2.

Drup:
Duplicated of https://github.com/mirage/functoria/issues/37

mmaker:
yo, this is not yet fixed. With mirage 2.7.0 and functoria  1.0.0: 
```
$ cat config.ml 
foo
$ mirage configure --unix
mirage: unknown option `--unix'.
Usage: mirage configure [OPTION]... 
Try `mirage configure --help' or `mirage --help' for more information.
$ mirage configure 
[ERROR]
  
ocamlfind ocamldep -package mirage -modules config.ml > config.ml.depends
ocamlfind ocamlc -c -annot -bin-annot -package mirage -o config.cmo config.ml
+ ocamlfind ocamlc -c -annot -bin-annot -package mirage -o config.cmo config.ml
File "config.ml", line 1, characters 0-3:
Error: Unbound value foo
Command exited with code 2.
```


The error message from the first command is definitely unhelpful.

Drup:
Dammit. I was convinced this one was solved.

@yallop Do you have an idea ? I think it is related to this line: https://github.com/mirage/functoria/blob/master/app/functoria_app.ml#L630

yallop:
I think it'd be good to start thinking how to remove `base_context`.

Drup:
@yallop That means droping `get_mode`. I don't disagree, but it seems quite early.

samoht:
I'm fine with it as long it is documented

yomimono:
Does it still seem quite early?  I'd really like to solve this for 3.0, and that's an excellent time to drop legacy stuff.

samoht:
[...]

PR #560, opened 2016-07-15
Deprecate functions which require functoria base_context

yomimono:
See #461 for motivation, and https://github.com/mirage/functoria/pull/65 for the related upstream change.

This patch removes the following deprecated functions:

* invoking mirage configure --unix and mirage configure --xen to build for UNIX/Xen targets.  Instead, use mirage configure -t unix or -t xen.

* using add_to_opam_packages to install extra required packages not detected by the mirage front-end tool.  Instead, pass the same list to `register` in the ~packages named argument.

* similarly, using add_to_ocamlfind_libraries to link against extra libraries.  Instead, pass the same list to `register` in the ~libraries named argument.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ env:
   - UPDATE_GCC_BINUTILS=1
   - EXTRA_REMOTES="https://github.com/mirage/mirage-dev.git"
   matrix:
-  - PACKAGE=mirage OCAML_VERSION=4.03 PINS="charrua-core:git://github.com/yomimono/charrua-core.git#use_new_marshallers dns:git://github.com/yomimono/ocaml-dns.git#standardize_record_fields functoria:git://github.com/yomimono/functoria.git#no-base-context mirage:git://github.com/yomimono/mirage.git#no-base-context"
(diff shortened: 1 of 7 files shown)


samoht:
Sweet! I'm also tempted to add a combinator for packages (as discussed elsewhere) but let's keep that PR simple and maybe not for mirage 3.0.

.travis.yml
```diff
@@ -5,9 +5,9 @@ script: bash -ex .travis-ci.sh
 env:
   global:
   - UPDATE_GCC_BINUTILS=1
-  - EXTRA_REMOTES="https://github.com/yomimono/mirage-dev.git#remove-console"
+  - EXTRA_REMOTES="https://github.com/mirage/mirage-dev.git"
```
hannesm: since we use the mirage-dev remote, we could just add the PINS below to that remote... instead of using both mechanisms....

yomimono: That remote is pointing at `master` of mirage, which this branch isn't; I could have made a separate branch of the remote, though.  I'll do that in the future.

hannesm: this charrua-core pin also looks fishy... hasn't that now be merged to master (and shouldn't we have a charrua-core package in mirage-dev)? (sorry for nitpicking, I'll go back and code)

Drup:
(I missed the party, but both patches look correct to me. I'm not convinced `--unix` and `--xen` should be removed though)

PR #561, opened 2016-07-19
Warn on use of deprecated functions.

yomimono:
Per @amirmc 's request, this PR proposes a version of `mirage` which warns users of `--unix`, `--xen`, `add_to_ocamlfind_libraries`, and `add_to_opam_packages` that use of these should be replaced, and suggests the appropriate replacements.


lib/mirage.ml
```diff
@@ -1566,10 +1566,22 @@ let get_mode () = Key.(get (get_base_context ()) target)
 
 let libraries_ref = ref []
 let add_to_ocamlfind_libraries l =
+  let pp_v fmt lib = Format.fprintf fmt "\"%s\" ;" lib in
+  let list_pp fmt l = Format.pp_print_list pp_v fmt l in
+  Format.eprintf "add_to_ocamlfind_libraries [%a] is deprecated. \
+      Please instead call `register` with `~libraries:[%a]`.\n"
+      list_pp l list_pp l
+  ;
```
talex5: Not sure if it's an improvement, but I think you can use an `[@@ocaml.deprecated]` annotation for this.

yomimono: Oh, neat, I didn't know about that.  Thanks!

avsm:
LGTM for a point release.

lib/mirage.ml
Drup: Yes, using a deprecation annotation is cleaner.

yomimono: Hm, it looks to me like I can only give `[@@ocaml.deprecated]` a static string, instead of the error message here which lists the elements in question.

talex5: True. On the other hand, the deprecation warning will tell the user the line in their source code that needs removing, and they can copy the list from there.

yomimono: Following (I think) @drup's advice at http://stackoverflow.com/questions/37035225/ocaml-attributes I added some annotations to the corresponding `val`s in `mirage_key.mli` and `mirage.mli`, but I don't get any output as a result of having done so.  Unless someone has time to tutor me in remedial deprecation annotations, I'm probably going to release as is.  (It's temporary code, anyway.)

talex5: Hmm. Is mirage loading the `config.ml` with warnings disabled I wonder?

Drup: @yomimono Can you point me to what you tried ?

yomimono: Recreated here: https://github.com/yomimono/mirage/tree/deprecation_annotations , specifically https://github.com/yomimono/mirage/blob/deprecation_annotations/lib/mirage_key.mli#L41 and https://github.com/yomimono/mirage/blob/deprecation_annotations/lib/mirage.mli#L438 . I get warnings when building the mirage front-end tool itself, but not when running `mirage configure` or `make`.  (I'm using a unikernel which calls both `add_to_ocamlfind_libraries` and `add_to_opam_packages`, and invoking it with `mirage configure --unix`.)

PR #565, opened 2016-07-21
Solo5 changes (new platform `Ukvm and `Virtio).

hannesm:
This is a manual merge of
76575138f27ccfbff2a073ba997c54d4afb84149..6ea671335a71e0f90e2c6b4933a2b830dcb88106
(from djwillia/mirage.git#solo5) and some massaging

Author: Dan Williams <djwillia@us.ibm.com>
Author: Martin Lucina <martin@lucina.net>
Author: Hannes Mehnert <hannes@mehnert.org>

please compare https://github.com/djwillia/mirage/compare/master...djwillia:solo5?expand=1

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -42,7 +42,10 @@ let io_page_conf = object
   method ty = io_page
   method name = "io_page"
   method module_name = "Io_page"
-  method libraries = Key.(if_ is_xen) ["io-page"; "io-page.unix"] ["io-page"]
+  method libraries =
+    Key.match_ Key.(value target) @@ function
+    | `Xen | `Virtio | `Ukvm -> ["io-page"]
+    | `Unix | `MacOSX -> ["io-page"; "io-page.unix"]
   method packages = Key.pure [ "io-page" ]
 end
 
@@ -68,7 +71,10 @@ let clock_conf = object (self)
   method ty = clock
   method name = "clock"
   method module_name = "Clock"
-  method libraries = Key.(if_ is_xen) ["mirage-clock-xen"] ["mirage-clock-unix"]
+  method libraries =
+    Key.match_ Key.(value target) @@ function
+    | `Xen | `Virtio | `Ukvm -> ["mirage-clock-xen"]
+    | `Unix | `MacOSX -> ["mirage-clock-unix"]
   method packages = self#libraries
 end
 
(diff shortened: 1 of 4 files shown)


yomimono:
At first glance this looks great to me.  I'll take some time to try it out once the packages it depends on have landed in `opam-repository`.

hannesm:
btw, this also gets rid of `is_xen` usages inside of mirage... maybe we should delete that function?

lib/mirage.ml
```diff
@@ -1401,16 +1462,39 @@ let configure_makefile ~target ~root ~name ~warn_error info =
               export OPAMVERBOSE=1\n\
               export OPAMYES=1";
   newline fmt;
-  let pkg_config_deps = "mirage-xen" in
+  let pkg_config_deps =
+    match target with
+    | `Xen -> "mirage-xen"
+    | `Virtio -> "mirage-solo5 ocaml-freestanding"
+    | `Ukvm -> "mirage-solo5 ocaml-freestanding solo5-kernel-ukvm"
```
hannesm: not sure whether solo5-kernel-ukvm is really needed here, maybe @mato / @djwillia know details (used below with `pkg-config --static --libs %s`)?

yomimono:
Generally, when there's a configuration incompatibility (e.g. --target=virtio --net=socket) we currently say that the target isn't supported for Xen; we should instead say something like "hypervisor target" or "non-Unix target".

lib/mirage_key.mli
```diff
@@ -27,12 +27,12 @@ include Functoria.KEY with module Arg := Arg
 
 (** {2 Mirage keys} *)
 
-val target: [ `Unix | `Xen | `MacOSX ] key
+val target: [ `Unix | `Xen | `MacOSX | `Virtio | `Ukvm ] key
 (** [-t TARGET]: Key setting the configuration mode for the current project.
-    Is either ["xen"], ["unix"] or ["macosx"].
+    Is either ["xen"], ["unix"], ["macosx"], ["virtio"] or ["ukvm"].
```
yomimono: `s/either/one of/`, I think.

Drup:
I took a quick look, the part where I'm competent looks correct. :)

> btw, this also gets rid of is_xen usages inside of mirage... maybe we should delete that function?

We should certainly change its name. The function is used in mirage-skeleton, but I think most of the intended semantics is "is_something_unixy". Is that a useful enough predicate to have ?

PR #568, opened 2016-07-23
Emit an ocamlfind predicate that matches the target

pqwy:
This relatively innocuous change adds arguments like `-tags 'predicate(xen)'` to the `ocamlbuild` invocation in the `Makefile`. The set of tags is currently `[unix, xen, solo5]`, but this is open for discussion.

The motivation is my long standing frustration: one can do a variety of simple interactions with the system from the code written directly in a unikernel, but it's almost impossible to extract that code to a library.

One example is in the [simplest Mirage demo](https://github.com/mirage/mirage-skeleton/blob/079e37e1b46831e9816c84f19a491356856994e9/console/unikernel.ml#L10). This call is not available from code compiled separately. Of course, I can use a [different one](https://github.com/mirage/mirage/blob/9a52cca64e25feef35099ca2bfd8e4ad52b881ed/types/V1.mli#L76) for that particular purpose, at the cost of functorizing my library, and any bit of code that ever wants to interact with it. But I cannot [peek into the args](https://github.com/mirage/mirage-platform/blob/master/unix/lib/env.mli#L19), [hook into the life-cycle](https://github.com/mirage/mirage-platform/blob/master/unix/lib/lifecycle.mli#L15), or [interact with the event loop](https://github.com/mirage/mirage-platform/blob/master/unix/lib/main.mli).

Actually I _can_ write a bit of OCaml code that does exactly that; the problem is that I need to compile it against at least three implementations of the same interface, and package the three resulting libraries. This is not where the pain stops, it's where it begins: since I now have three separate libraries, either in opam or just in ocamlfind sense, in order to make it seamless for others to use my code, I need to create a pull request to this very repo. In this PR I bake the knowledge of my new libraries into `mirage`, and have it decide which one to use depending on the backend. To add to the fun, I also need to invent semantics of triggering this backend-directed resolution of dependencies. Do I expect the user to depend on the canonical version, intercept it, and inject the specialized version instead? Or maybe just parse the unikernel source, try to detect references to `Twigglet`, and make that trigger my logic.

By exposing the mirage target to ocamlfind, library authors can do surprisingly flexible things just from the `META` files, without touching the `mirage` tool itself. They can also hide this parametricity of their libraries from the users. My motivating example is a completely  trivial [shim](https://github.com/pqwy/mirage-os-shim/blob/master/unix/mirage_OS.ml) over the common intersection of the `OS` API. It needs a bit of work, and having the actual API signature it presents blessed in the official code would help. But it's already solving headaches over library duplication I have elsewhere.

This should of course in no sense be considered a critique of the main design tenet behind Mirage, that interfaces should be functorized over. That technique seems to work very well in most places. It's just that functorizing a library and its six dependants to invoke `sleep` doesn't seem to be one of them.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1454,9 +1454,12 @@ let configure_makefile ~target ~root ~name ~warn_error info =
       append fmt "SYNTAX = -tags \"thread,%s\"\n" default_tags;
       append fmt "FLAGS  = -r -cflag -g -lflags -g,-linkpkg\n"
   end;
+  append fmt "TAGS = \"predicate(%s)\""
+    (match target with
+      `Unix | `MacOSX -> "unix" | `Xen -> "xen" | `Virtio | `Ukvm -> "solo5" );
   append fmt "SYNTAX += -tag-line \"<static*.*>: warn(-32-34)\"\n";
   append fmt "LD?=ld";
-  append fmt "BUILD  = ocamlbuild -use-ocamlfind $(LIBS) $(SYNTAX) $(FLAGS)\n\
+  append fmt "BUILD  = ocamlbuild -use-ocamlfind -tags $(TAGS) $(LIBS) $(SYNTAX) $(FLAGS)\n\
               OPAM   = opam\n\n\
               export PKG_CONFIG_PATH=$(shell opam config var prefix)\
               /lib/pkgconfig\n\n\


talex5:
This might be worth adding, as the linker hack is sometimes useful, but all of the example uses suggested seem bad to me.

We really want to avoid libraries using `OS.Time`. This makes them almost impossible to unit-test. Same goes for peeking the args - libraries should never do this. And if a library wants to hook into my unikernel's event loop, I'd quite like to be involved in that.

On the other hand, perhaps this would help us remove the Xen linkopts hack.

(and we should probably change the hello world example to take `TIME` as an argument)

Drup:
I understand your use case, but one issue I see with this approach is that functoria starts not being able to inform you of the subdependencies. The fact that we can manipulate functors and such without linker tricks also enables more precise documentation.

I wonder if we could emulate GHC's backpack link calculus somehow, that would allow to solve this in a more controlled manner. 
(And please don't peak into argv, we have keys for that.)

That being said, adding the tag sounds like a good idea for a bunch of other reasons (in particular, when unikernel become more complicated, they sometimes requires weird ocamlbuild rules).

pqwy:
@talex5 
A strong stance that all interfaces must be functorized over is certainly a valid position, as much as I don't really agree from practical experience. But we currently have situations where they are not, and the only alternative is to explode the number of packages. I feel that this is a much cleaner solution, and it does not hinder moving into either direction later on.

And yes, I also think this can subsume `linkopts`. Didn't have time to demonstrate that, yet.

@Drup 
Hiding stuff from functoria is by far the worst drawback. But AFAICT adding the extra predicate consistently to all invocations of ocamlfind should fix it. Digging a bit through `mirage` and `functoria` I can't quite pinpoint where exactly ocamlfind is invoked on all `configurable#libraries` to detect the transitive deps... But ultimately the knowledge about the dependencies lies in ocamlfind, and functoria is used to described the dependency graph roots. So if all the invocations of ocamlfind have consistent predicates, functoria can learn what exactly ocamlfind will do. Am I completely off the mark here?

hannesm:
I'm in favour of this change / feature, and believe it'll reduce code duplication and build system hacks (as present in current libraries).  The `mirage-os-shim` library is a nice layer of abstraction for libraries depending on system components, which life outside of a functor.

Drup:
> Digging a bit through mirage and functoria I can't quite pinpoint where exactly ocamlfind is invoked on all configurable#libraries to detect the transitive deps..

Apparently it isn't ... Thinking about it more, I think I removed it because it was too verbose. Regardless, my remark was more due to the fact that we turn things that would have been functor applications (which are shown in the functoria graph) into link magic (which does not).

As long as it's not used for complicated things, I don't think it will be too bad. I would still prefer to have a solution that doesn't depend on ocamlfind though. At the moment, functoria/mirage doesn't do anything intelligent with ocamlfind (except maybe for nocrypto, but that's backcompat).

PR #576, opened 2016-08-09
ukvm target: Correct use of "blk" for block module

mato:
Fixes solo5/solo5#70 (easier doing it here than the other way round) and
mirage/mirage-skeleton#178.


hannesm:
LGTM

samoht:
I've restarted the failing jobs.

PR #577, opened 2016-08-09
Remove extraneous solo5-kernel-ukvm from configure_makefile

mato:
As spotted by @hannesm in #565, this is not required here and will be
pulled in from ocaml-freestanding pkg-config.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1467,7 +1467,7 @@ let configure_makefile ~target ~root ~name ~warn_error info =
     match target with
     | `Xen -> "mirage-xen"
     | `Virtio -> "mirage-solo5 ocaml-freestanding"
-    | `Ukvm -> "mirage-solo5 ocaml-freestanding solo5-kernel-ukvm"
+    | `Ukvm -> "mirage-solo5 ocaml-freestanding"
     | `MacOSX | `Unix -> ""
   in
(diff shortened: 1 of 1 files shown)


hannesm:
hmm, in the `pkg-config` universe, does `mirage-solo5` also depend on `ocaml-freestanding`?  if this would be the case, only `mirage-solo5` would be needed (which would be symmetric to the `mirage-xen` case for xen)!?!  OTOH, explicitly specifying dependencies is nicer, rather than depending on transitivity of dependencies.

mato:
On Tuesday, 09.08.2016 at 08:12, Hannes Mehnert wrote:
> hmm, in the `pkg-config` universe, does `mirage-solo5` also depend on `ocaml-freestanding`?  if this would be the case, only `mirage-solo5` would be needed (which would be symmetric to the `mirage-xen` case for xen)!?!  OTOH, explicitly specifying dependencies is nicer, rather than depending on transitivity of dependencies.

It doesn't at the moment (see mirage-solo5/solo5-bindings/mirage-solo5.pc).
I can make it do that if you think it'll make things clearer.

hannesm:
thanks for the pointer.  I think it is fine as proposed, and once travis is happy this PR can be merged...

mato:
CI has passed on this, so merging.

PR #578, opened 2016-08-12
Add an is_unix key

mato:


yomimono:
This looks good to me.  Do you plan on following with another PR to remove `is_xen`?

mato:
Removing is_xen would mean an API break for anyone using it, right? Do we just spring that on people outright, or is there some deprecation mechanism we can use?

lib/mirage_key.mli
```diff
@@ -38,6 +38,9 @@ val pp_target: [ `Unix | `Xen | `MacOSX | `Virtio | `Ukvm ] Fmt.t
 val is_xen: bool value
 (** Is true iff the {!target} keys takes the value [`Xen]. *)
 
+val is_unix: bool value
+(** Is true iff the {!target} key is a UNIXish system. *)
```
Drup: Maybe provide the exact list of "UNIXish systems" ?

Drup:
I think you could just keep the usage of `(value target)` inside mirage itself. Probably more future-proof (forcing people who add targets to look at all the cases).

PR #585, opened 2016-09-01
Express io-page dependency of crunch in libraries and packages.

yomimono:
This change is needed to get `io-page` included in the set of packages against which a unikernel that includes a `crunch` filesystem builds.  The autogenerated `staticN.ml` requires `io-page`, but installing and running `crunch` itself doesn't, which is why the dependency isn't expressed in the `opam` file for the `crunch` package.

`get_extra_ld_flags` in `mirage.ml` will find `io-page` in the list when compiling for `-t xen` already, because `io-page` is a dependency of the `mirage-xen` package, so we always get it "for free" in that backend -- otherwise, the existing code wouldn't have ever worked.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -169,8 +169,8 @@ let crunch dirname = impl @@ object
     val name = Name.create ("static" ^ dirname) ~prefix:"static"
     method name = name
     method module_name = String.Ascii.capitalize name
-    method packages = Key.pure [ "mirage-types"; "lwt"; "cstruct"; "crunch" ]
-    method libraries = Key.pure [ "mirage-types"; "lwt"; "cstruct" ]
+    method packages = Key.pure [ "mirage-types"; "lwt"; "cstruct"; "crunch"; "io-page" ]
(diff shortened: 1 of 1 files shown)


yomimono:
I'd appreciate a quick look from @drup or maybe @samoht to make sure I haven't misunderstood something here.

Drup:
Seems good. package/libraries is supposed to list "dependencies" in the most general meaning, so both configure and runtime.

hannesm:
I have a question here to @Drup -- the line below the change says `method deps = [ abstract default_io_page ]` -- shouldn't that drag in the `default_io_page` dependency, which depends on the io-page opam package?  Or what exactly is `deps` used for?

Drup:
Huuum, it should, yes. @yomimono, do you remember why it was needed ?

yomimono:
@drup Nope, I don't remember, and neither do my notes.  Sorry :(

hannesm:
I opened #699 to not loose this issue (too hard to keep track of closed issues/PRs in GitHub)... we should find the root of the problem.

PR #586, opened 2016-09-01
also remove Makefile.ukvm when running `mirage clean`

yomimono:
`mirage configure -t ukvm` generates a `Makefile.ukvm`.  Remove it with `mirage clean` alongside `Makefile`.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1620,7 +1620,9 @@ let configure_makefile ~target ~root ~name ~warn_error info =
   append fmt "-include Makefile.user";
   R.ok ()
 
-let clean_makefile ~root = Cmd.remove (root / "Makefile")
+let clean_makefile ~root =
+  Cmd.remove (root / "Makefile");
+  Cmd.remove (root / "Makefile.ukvm")
 
 let check_ocaml_version () =
   (* Similar to [Functoria_app.Cmd.ocaml_version] but with the patch number *)


mato:
Apart from `Makefile.ukvm` it also generates a `_build-ukvm/` directory which we presumably also want to remove? I say presumably because a `make clean` will remove both of those (and also the `ukvm-bin` hypervisor binary, which is probably incorrect since you need it to run the unikernel...?)

yomimono:
I generally expect `make clean` to remove all the artifacts generated by `make`, so my intuition is that anything autogenerated should go, but I don't feel particularly strongly about it.

mato:
Let me rephrase that question -- should `mirage clean` invoke `make clean` if a `Makefile` is present? That way the `mirage` tool doesn't have to care about the details of any artifacts `make` may have generated.

yomimono:
My response clearly didn't quite have enough of my attention in it; I apologize, and thanks for the restatement.  Having `mirage clean` attempt to invoke `make clean` and then remove anything it knows `mirage` is likely to have generated (including the `Makefile`?) that it knows about would probably be more sensible, yeah.

mato:
My comment wasn't clear either, so no worries :-) I don't see any downside in `mirage clean` invoking `make clean` but it's a change to current behaviour, worth checking with others? Any error status from `make clean` should be ignored since it's irrelevant.

hannesm:
The (by mirage configure) generated makefile in the ukvm case contains a `clean` rule which already includes a `rm Makefile.ukvm`, but only after invoking a `make -f Makefile.ukvm ukvm-clean`, which gets rid of the `_build-ukvm` directory...

TL;DR: mirage clean should first execute `$(make) clean`, and then do its clean (including removing the Makefile).

mato:
Does mirage have any way to figure out what $(MAKE) should be?

hannesm:
no. but since our Makefile is any make compatible (some targets won't work with BSD make after #691), we can just call `make clean`... (yeah, this is hackish, the better solution is to solve #700 IMHO)

mato:
If `mirage clean` can do the equivalent of `rm -rf _build_ukvm Makefile.ukvm` then that's also fine (for now) IMO. I for one would be really happy for it to delete those as I find myself constantly doing `make clean && mirage clean`.

hannesm:
in #703 `make clean` and `mirage clean` are the same, and do remove `Makefile.ukvm` and `_build-ukvm`... I suspect this can be closed?

mato:
@hannesm Yes, once #703 is in.

mato:
Superseded by #703

PR #598, opened 2016-09-28
doc: rename odoc->odig and do not install it twice

avsm:


Dockerfile.doc
```diff
@@ -17,23 +17,21 @@ RUN opam depext -uivj 3 mirage tcpip tls cohttp lwt irmin ipaddr uri github alco
     mirage-fs-unix \
     git git-unix mirage-git  \
     core_kernel \
-    topkg-care \
+    todig-care \
```
dbuenzli: find and replace error.

avsm: oops... lazily depending on the autobuilds

avsm: fixed in aff146c49f6faf13b7a4e0ad9e1a785dfb98f9dc

PR #599, opened 2016-09-28
doc: add more libraries to the documentation mix

avsm:


hannesm:
hmm, is this becoming another TROVE?  should they be unified at some point?

avsm:
Yes, this is just an experiment before I do a mass unification before the release...

avsm:
(and link from the homepage etc)

hannesm:
\o/

PR #602, opened 2016-10-01
Error

hannesm:


hannesm:
as propsed on 30. September 2016 on the mailing list (https://lists.xenproject.org/archives/html/mirageos-devel/2016-09/msg00050.html) with ack from @Drup @avsm @talex5 

This is a tracking issue, so far I did
- https://github.com/mirage/functoria/pull/71 (additional: use `Lwt.return` instead of `return`)
- https://github.com/mirage/mirage/pull/602 (additional: opens `Lwt.Infix` for `main.ml` instead of `Lwt`; also removes some dead code (`string_of_network_init_error` sorry dsheets))
- https://github.com/mirage/mirage-platform/pull/171 UNIX part: do main thing in a `Lwt.catch`, and `Logs.err` on exception, xen & solo5 should be updated in the same way; bonus: get rid of `Printf`
- https://github.com/mirage/mirage-clock/pull/18
- https://github.com/mirage/mirage-console/pull/42

missing: netifs, block devices....

hannesm:
the failures I can see (in fat / tar / ...) are due to the usage of a connect from a different library (and installing a released version instead of the git master)... certainly this is one of the changes which affects lots of libraries, but nearly no client code (unikernels don't call out to connect usually)

hannesm:
Travis has happily built mirage-skeleton with this changeset (https://travis-ci.org/mirage/mirage-skeleton/builds/164402978)

There are the set of PRs mentioned above, plus one to extend mirage-dev with mirage-block* etc. (using their master branches, thus the other PRs should be merged before): https://github.com/mirage/mirage-dev/pull/155

PR #607, opened 2016-10-02
expose `direct` as an option for `kv_ro`

mor1:


yomimono:
Thanks for this patch, @mor1 !

PR #590, opened 2016-09-13
[followon #558] Topkg port

avsm:

PR #610, opened 2016-10-04
Random fixes

hannesm:
two changes:
- `connect` does not return a result anymore (somehow this slipped through a rebase)
- a `generator` type in `RANDOM` interface, and use it in `generate`.  this makes the signature the same as Nocrypto.Rng, and thus interoperable.


hannesm:
CI failures: in contrast to before #590, no longer mirage and mirage-types are pinned (this leads to 6 failures); the remaining 2 are due to (in ocaml-dns) `The value `mem' is required but not provided` (fixed in https://github.com/mirage/ocaml-dns/pull/103).

Since both mirage-stdlib-random and (the currently pinned version of nocrypto) implement the new interface, I'll merge this PR (to avoid CI failures in other PRs).

PR #611, opened 2016-10-04
prng key

hannesm:
The goal is to have a `prng` configuration time key to select which PRNG to use by default (for all devices which need random numbers).  A `config.ml` can manually specify `stdlib_random` or `nocrypto_random` if insisting on a specific PRNG (but nearly all use cases should use `default_random ()`).

@Drup I implemented a `prng` key, filling a `default_random ()` value.. now, if multiple `stack`s are used; or I use a stack and `default_random ()` both in `config.ml`, I get an assertion failure from functoria https://github.com/mirage/functoria/issues/70 ...

### DIFF
--- a/lib/mirage.mli
+++ b/lib/mirage.mli
@@ -117,6 +117,12 @@ val stdlib_random: random impl
 val nocrypto_random: random impl
 (** Passthrough to the Fortuna PRNG implemented in nocrypto. *)
 
+val default_random: random impl
+(** Default PRNG device to be used in unikernels.  It defaults to {!stdlib}, but
+    users can override using the {!Key.prng} command line argument.
+*)
+
+
 
 (** {2 Consoles} *)
 
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -173,6 +173,12 @@ end
 
 let nocrypto_random = impl nocrypto_random_conf
 
+let default_random =
+  match_impl (Key.value @@ Key.prng ()) [
+    `Stdlib  , stdlib_random;
+    `Nocrypto, nocrypto_random;
+  ] ~default:stdlib_random
+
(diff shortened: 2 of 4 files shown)


Drup:
Hum, that assert failure is a bit distateful, I'll have to look into what triggers it exactly.

In your case, you are creating multiple keys with the same name. I though it would trigger a clean error message though ...
The simple solution is to create *one* default prng key and to use that key for the optional argument. I would guess that's the behavior you want anyway: by default, if the default random device is used multiple times, you want one key to flip them all.

yomimono:
@hannesm: ping?

hannesm:
I updated this PR to work as expected, there is now a `default_random` provided by mirage.mli, to be used by people programming unikernels! (if someone wants to use `stdlib_random` or `nocrypto_random` explicitly, these are exposed as well)

and a prng key in mirage_key to be filled during configuration (thus someone who wants to compile and run a unikernel can select the value of the default random device).  the default default random device is stdlib.

lib/mirage.mli
```diff
@@ -117,6 +117,13 @@ val stdlib_random: random impl
 val nocrypto_random: random impl
 (** Passthrough to the Fortuna PRNG implemented in nocrypto. *)
 
+val default_random: random impl
+(** PRNG that will choose dynamically between {!stdlib} and {!nocrypto}.
```
yomimono: If I understand the code in `default_random` correctly, this will always be `stdlib` unless the user specifies otherwise -- this should probably be explained here, if I'm correct.

hannesm: hope that's better now!

yomimono:
This definitely looks like an improvement to me!  Thanks for sorting it out.

PR #615, opened 2016-10-06
Network error

hannesm:
see https://github.com/mirage/mirage-dev/pull/166

### DIFF
--- a/pkg/pkg.ml
+++ b/pkg/pkg.ml
@@ -33,7 +33,10 @@ let () =
     Ok [ Pkg.lib "pkg/META.mirage-types" ~dst:"META";
          Pkg.lib "types/V1.mli";
          Pkg.lib "types/V1.cmi";
+         Pkg.lib "types/V1.cmti";
+         Pkg.mllib "types/mirage-types.mllib";
          Pkg.lib ~cond:lwt "types/V1_LWT.mli";
-         Pkg.lib ~cond:lwt "types/V1_LWT.cmi"; ]
+         Pkg.lib ~cond:lwt "types/V1_LWT.cmi";
+         Pkg.lib ~cond:lwt "types/V1_LWT.cmti"; ]
   | other ->
     R.error_msgf "unknown package name: %s" other
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -16,6 +16,8 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
+open Result
+
 (** MirageOS Signatures.
 
     This module defines the basic signatures that functor parameters
@@ -56,6 +58,7 @@ module type DEVICE = sig
   type t
   (** The type representing the internal state of the device *)
 
+  (* XXX: no good reason to have it here anymore! *)
   type error
   (** The type for errors signalled by the device *)
 
@@ -299,17 +302,9 @@ module type BLOCK = sig
 
 end
 
-(** {1 Network stack}
+(** {1 Network stack} *)
 
-    A network interface that serves Ethernet frames. *)
-module type NETWORK = sig
-
-  type page_aligned_buffer
-  (** The type for page-aligned memory buffers. *)
-
-  type buffer
-  (** The type for memory buffers. *)
-
+module Network : sig
   type error = [
     | `Unknown of string (** an undiagnosed error *)
     | `Unimplemented     (** operation not yet implemented in the code *)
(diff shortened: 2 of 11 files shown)


pkg/pkg.ml
```diff
@@ -33,7 +33,10 @@ let () =
     Ok [ Pkg.lib "pkg/META.mirage-types" ~dst:"META";
          Pkg.lib "types/V1.mli";
          Pkg.lib "types/V1.cmi";
+         Pkg.lib "types/V1.cmti";
          Pkg.lib ~cond:lwt "types/V1_LWT.mli";
-         Pkg.lib ~cond:lwt "types/V1_LWT.cmi"; ]
+         Pkg.lib ~cond:lwt "types/V1_LWT.cmi";
+         Pkg.lib ~cond:lwt "types/V1_LWT.cmti";
+         Pkg.mllib "types/mirage-types-pp.mllib"; ]
```
samoht: this should just be `mirage-types.mllib` and we should start adding useful code in there: pretty printers are a good start!

pkg/pkg.ml
```diff
@@ -33,7 +33,10 @@ let () =
     Ok [ Pkg.lib "pkg/META.mirage-types" ~dst:"META";
          Pkg.lib "types/V1.mli";
          Pkg.lib "types/V1.cmi";
+         Pkg.lib "types/V1.cmti";
          Pkg.lib ~cond:lwt "types/V1_LWT.mli";
-         Pkg.lib ~cond:lwt "types/V1_LWT.cmi"; ]
```
dbuenzli: Are these mli only compilation units ? I should introduce an `Exts.types_api` shortcut for these.

dbuenzli: See https://github.com/dbuenzli/topkg/issues/74

samoht: yes and I remember doing similar things in other packages too

pkg/pkg.ml
hannesm: but I couldn't figure out how to put V1 in the mllib without an implementation, and V1_LWT only conditionally...

pkg/pkg.ml
dbuenzli: Aaah @samoht I'm very disappointed you didn't complain.

hannesm:
I updated this PR, now there is a `mirage-types-runtime` (providing atm pretty printers) and a `mirage-types-runtime-lwt` (providing useful combinators for combining result + lwt).

hannesm:
(if there's a way to include `V1.mli` and `M_pp.ml` and `M_pp.mli` (without having to write a `V1.ml`) in a single `mllib` file, please let me know to avoid this library explosion)

types/V1.mli
```diff
@@ -299,38 +302,53 @@ module type BLOCK = sig
 
 end
 
-(** {1 Network stack}
+(** {1 Network stack} *)
+
+module Network : sig
+  type error = [
+    | `Unknown of string (** an undiagnosed error *)
```
yomimono: Perhaps we should change `Unknown` to `Msg` to fit in more nicely with Rresult?

hannesm: sure, and imho remove `` `Disconnected`` (see https://lists.xenproject.org/archives/html/mirageos-devel/2016-10/msg00020.html).  In here I intentionally did not rename or remove anything, but `only` put in result and centralised the error polyvar (and pp).

samoht:
Yes there is ;-)

```
$ cat types/mirage-types.mllib
runtime/M_pp
```

```
$ cat pkg/META.types
version = "%%VERSION%%"
description = "Collection of module signatures for MirageOS"
exists_if = "V1.cmi"
requires = "result"
archive(byte) = "mirage-types.cma"
archive(native) = "mirage-types.cmxa"
plugin(byte) = "mirage-types.cma"
plugin(native) = "mirage-types.cmxs"
exists_if = "mirage-types.cma"
```

etc.

The only thing to remember is that`mli/cmi` never appear in META files (because you only need them at compile time, and they are made visible with `-I`) but they need to be installed properly, so the `cmi/cmti` lines that you added in `pkg/pkg.ml` are indeed needed.

hannesm:
@samoht thanks, so no mention of V1 in the mllib?

hannesm:
force-pushed thx to @samoht (also force-pushed my tcpip branch)

samoht:
exactly, don't mention the mli files in the mllib/META.

types/V1.mli
```diff
@@ -299,38 +302,53 @@ module type BLOCK = sig
...
   (** [write nf buf] outputs [buf] to netfront [nf]. *)
 
-  val writev: t -> buffer list -> unit io
+  val writev: t -> buffer list -> (unit, error) result io
   (** [writev nf bufs] output a list of buffers to netfront [nf] as a
       single packet. *)
 
-  val listen: t -> (buffer -> unit io) -> unit io
+  (* XXX: listen currently runs forever (documentation wrong)! *)
+  val listen: t -> (buffer -> unit io) -> (unit, error) result io
```
yomimono: This `buffer -> unit io` means the function we call can't pass us any errors -- we can only emit those we caused ourselves.  When attempting to revise users to use result types, this presents some difficulties.  Is it intentional, or an opportunity for more discussion?

talex5: (also, the second `unit` should be `'a` if it never returns except on error)

hannesm:
I rebased on master, now use `%%VERSION_NUM%%` in META, and got again rid of the lwt result combinators since [Lwt_result.Infix](https://ocsigen.org/lwt/dev/api/Lwt_result.Infix) exists (plus discussion about naming on the mailing list).  From my perspective this is ready to merge (as said in the Mirage call today), and we can iterate over the API in subsequent commits (as discussed here, rename `` `Unknown of string `` to `` `Msg of string ``, find a reasonable signature for `listen`, etc.).

hannesm:
I restarted the mirage-dev build over [here](https://travis-ci.org/mirage/mirage-dev/builds/165685332) after the last rebase + force-push

hannesm:
which just now turned all green...

yomimono:
In the future I'll make sure to mention that I'm basing my own revisions off of your branches, in hopes of avoiding having to clean up after force-pushes.

I'll merge to make this less messy and because having better error handling for the individual `NETWORK` implementations has some value outside of the higher levels of the TCP/IP stack.  I expect this API (particularly for `listen`) to change further, as @hannesm mentioned.

hannesm:
thx, agreed!  @yomimono sorry for rebasing and force-pushing so often...

PR #619, opened 2016-10-13
correct comment on `listen` and `input` operation in NETWORK and ETHIF

yomimono:
Documentation claimed that `listen` and `input` would start asynchronous threads
and return immediately.  This is incorrect; implementations don't return
until `disconnect` has been called on the underlying interface.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -332,9 +332,9 @@ module type NETWORK = sig
 
   val listen: t -> (buffer -> unit io) -> unit io
   (** [listen nf fn] is a blocking operation that calls [fn buf] with
-      every packet that is read from the interface.  It returns as
-      soon as it has initialised, and the function can be stopped by
-      calling [disconnect] in the device layer. *)
+      every packet that is read from the interface.
+      The function can be stopped by calling [disconnect]
+      in the device layer. *)
 
   val mac: t -> macaddr
   (** [mac nf] is the MAC address of [nf]. *)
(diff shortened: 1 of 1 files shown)


yomimono:
As pointed out by @hannesm in https://github.com/mirage/mirage/pull/615 .

hannesm:
is this the intended behaviour?  or should we fix the implementations and users of the API?

yomimono:
Hm, perhaps @avsm or @samoht can speak to what the intended behavior is?  History on those docstrings goes back to the earliest revision of `V1.mli` I can find with the same text.

samoht:
I am not sure anything was really decided about that. Modifying the docs to reflect reality seems to be a good idea.

hannesm:
sure, fixing documentation is a good thing (thus, please feel free to merge).

The current implementation is nice if you deal with a single network interface.  This is fine since most of our examples deal with only one network interface.   For a second interface you'll have to call listen outside of the main task `Lwt.async (fun () -> Netif.listen t cb)`.

OTOH, maybe we want to make the task coordination explicit?  Thus, `listen` could return a `unit -> Lwt.t` which loops forever (or, considering the discussion in #615 it should return ``(unit, `Disconnected) result Lwt.t``)).  Then a caller would be of the form: `let netif_loop = listen netif cb in netif_loop ()`.

I'd be in favour of making the tasks explicit.  I did not include that in #615 since I wanted to discuss only error handling over there.

yomimono:
When I use more than one network interface, I use `Lwt.join` on the array of `listen` calls, which I don't find to be too cumbersome.  YMMV?

yomimono:
Can we move this discussion to either `mirage-dev` or the mailing list?  This PR is a very small correction to incorrect documentation, and it'd be a shame to lose the discussion after merge.

PR #623, opened 2016-10-17
reword missing ocaml-crunch slightly; use LD from pkg-config for virtio

hannesm:


hannesm:
@mato in case of virtio now pkg-config variable ld is used.

avsm:
This does put the logic for LD detection in an odd place, since it's not clear if other virtio backends will have a pkg-config entry :-/ Is FreeBSD the only known one atm (aside from HyperKit on OSX)?

lib/mirage.ml
```diff
@@ -1489,13 +1489,15 @@ let configure_makefile ~target ~root ~name ~warn_error info =
...
   append fmt "SYNTAX += -tag-line \"<static*.*>: warn(-32-34)\"\n";
-  append fmt "LD?=ld";
   append fmt "BUILD  = ocamlbuild -use-ocamlfind -tags $(TAGS) $(LIBS) $(SYNTAX) $(FLAGS)\n\
               OPAM   = opam\n\n\
               export PKG_CONFIG_PATH=$(shell opam config var prefix)\
               /lib/pkgconfig\n\n\
               export OPAMVERBOSE=1\n\
               export OPAMYES=1";
+  (match target with
+   | `Virtio -> append fmt "LD=$(shell pkg-config solo5-kernel-virtio --variable=ld)"
```
mato: 1. I'd make this case apply to both `Virtio` and `Ukvm` _targets_ for consistency -- solo5/solo5#102 deliberately introduces the `LD` variable for both even though it is only used by `Virtio`.
2. Can you use `LD?=$(shell ...)` instead of `LD=`, i.e. allow user override?

mato:
@avsm solo5/solo5#102 adds the value for consistency for both `virtio` and `ukvm` so it could (should) be used by both Solo5 targets on the Mirage end.

I'm not sure what you mean by "other virtio backends will have a pkg-config entry"?

hannesm:
@avsm so far I only know about the Solo5 virtio work -- I've no clue how `Hyperkit on OSX` would work / works with MirageOS, sorry.

As @mato said, detecting the linker once is a good property, and currently solo5 does so.  This unfortunately introduces another dependency on pkg-config, but I'm sure we'll solve that for MirageOS4.

avsm:
Yeah, this is not the PR on which we can solve all the pkg-config dependencies. This lgtm to merge to me.

mato:
+1

PR #624, opened 2016-10-19
Revert "reword missing ocaml-crunch slightly; use LD from pkg-config for virtio"

mato:
Reverts mirage/mirage#623 /cc @hannesm.

Getting LD this way does not actually work, `export PKG_CONFIG_PATH` does not affect the environment used by `$(shell ...)`. Given that this breaks current master I'm going to revert this PR and work on a proper fix where `mirage` will call out to `pkg-config` itself.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -237,7 +237,7 @@ let crunch dirname = impl @@ object
 
     method configure i =
       if not (Cmd.exists "ocaml-crunch") then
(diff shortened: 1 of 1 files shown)


mato:
Merging w/o waiting for CI since this fixes breakage on master.

hannesm:
thx @mato.  and sorry.  I had `PKG_CONFIG_PATH` set in my shell environment...

PR #625, opened 2016-10-19
Add Solo5 backends to CI

mato:
Now that mirage/mirage CI is building against mirage-dev we should also
be building the Solo5 targets.

/cc @avsm

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -4,3 +4,5 @@ cd mirage-skeleton
 eval `opam config env`
 MODE=unix make
 MODE=xen  make
+MODE=ukvm make
+MODE=virtio make


yomimono:
I've also been working on this, but with the container-based tests and separate matrix entries for the different backends (which I think is required, as all 4 can't complete in time on Travis currently).  See https://travis-ci.org/yomimono/mirage/builds/167688986 which is unfortunately also failing.

mato:
@yomimono Ah, I wasn't aware you were working on it. In that case I'll close this.

yomimono:
Sorry, I made a ticket for it a while ago and was waiting to PR until I had passes for at least some subset of the tests.  I'll hack on it again a bit today.

PR #628, opened 2016-10-19
Build solo5

yomimono:
Get the MODE from the build matrix instead of building with `MODE=unix`, then `MODE=xen`.  Also, add targets for `MODE=virtio` and `MODE=ukvm`. This makes the build matrix significantly larger.  

Some tests consistently fail with symptoms similar to https://gitlab.camlcity.org/gerd/lib-findlib/issues/9 -- `ocamlfind remove tls` attempts to remove `dlltcpip_stubs.so.owner`, which isn't present; this causes the `tls` reinstall to fail, which causes the test to fail entirely.


hannesm:
LGTM.  I'd cut down a random set of build jobs (maybe in total 8?)... 32 is slightly heavyweight (esp since there are only 5 Travis jobs for the whole mirage organisation)

mato:
:tada:

PR #632, opened 2016-10-24
Refresh doc gen to be a sorted list for upcoming trove

avsm:
Also adds a few ppx packages into the mix that we use for web
stuff (like tyxml-ppx)


avsm:
ubuntu mirror failures, rest is green, so merging

PR #633, opened 2016-10-24
disable warning 42; fixes #621; agreeing with @gasche argument; no 4.01 compat

### LINKED ISSUE #621: Allow disabling of warnings in config.ml
As suggested by @gasche in https://github.com/Engil/Canopy/pull/66 -- in this specific case warning 42 is signalled due to Syndic exposing several constructors with the same name from a single module.  There can be nothing done in Canopy to avoid this problem.  It would be great to be able to specify in config.ml which warnings to suppress.

hannesm:


yomimono:
This change looks good to me.  Test failures are Ubuntu failing to reach its package archive.

PR #635, opened 2016-10-26
Pull out the backend predicate

pqwy:
... and *rename* it.

Words like `xen`, and especially `unix`, hog the single `ocamlfind` predicate namespace. As this could become part of a long-lived API, I changed them to something more descriptive.


yomimono:
Thanks for this. :)

PR #639, opened 2016-10-29
generate .xl, .xl.in, .xe, and _libvirt.xml only if we're targeting xen

hannesm:
no use for those when using solo5 afaik

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1684,10 +1684,13 @@ let configure i =
   check_entropy @@ Info.libraries i >>= fun () ->
   Log.info "%a %a" Log.blue "Configuring for target:" Key.pp_target target ;
   Cmd.in_dir root (fun () ->
-      configure_main_xl ".xl" i;
-      configure_main_xl ~substitutions:[] ".xl.in" i;
-      configure_main_xe ~root ~name;
-      configure_main_libvirt_xml ~root ~name;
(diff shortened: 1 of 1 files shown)


yomimono:
Isn't `_libvirt.xml` still potentially useful in a non-Xen context?

hannesm:
I've no clue... from mirage.ml:

```
  append fmt "        <type arch='armv7l' machine='xenpv'>linux</type>";
  append fmt "        <kernel>%s/mir-%s.xen</kernel>" root name;
  ...
  append fmt"         <disk type='block' device='disk'>";
  append fmt "            <driver name='phy'/>";
  append fmt "            <source dev='/dev/loop0'/>";
  append fmt "            <target dev='' bus='xen'/>";
  append fmt "        </disk>";

```

If they would be more general, they might be useful with other backends.  I've no clue who (added and who) uses libvirt support (neither do I know or want to know how to fix it for other backends).

yomimono:
I think it was added by @magnuss for Jitsu.  That block is indeed pretty incorrect for non-Xen targets, so I'm fine with this PR; better to have nothing generated than something broken.

djs55:
IIRC it's not possible to write a generic libvirt xml config file-- the
format is more like the union of all formats with an XML syntax. The KVM
device info always looks different to the xen info :(

PR #641, opened 2016-11-01
Doc refresh

avsm:
Includes my changes and @hannesm


avsm:
merging so that the CI can just build and deploy in trunk

Issue #557, opened 2016-07-13
arp and ip module types should assume one ip address and default gateway


yomimono:
The current state (multiple default gateways; lists for netmasks but not addresses) is inconsistent and confusing.  The types hint at functionality that is not there.

talex5:
The netmask situation is especially confusing because of this, which could perhaps be fixed at the same time:

    and type prefix = Ipaddr.V4.t (* FIXME: Use Ipaddr.V4.Prefix.t *)

hannesm:
IMHO that should be `ip : t -> (Ipaddr.V4.t * Ipaddr.V4.Prefix.t)` (and a `set_` if we need to set the ip dynamically (or could we just have an IP passed at instantiation time, and get away without providing a mutable IP address?) -- if you like a new address, get a new stack

yomimono:
> (or could we just have an IP passed at instantiation time, and get away without providing a mutable IP address?) -- if you like a new address, get a new stack

So for DHCP, we'd construct a stack, get an IP, then destroy the stack and replace it with one with our new information?  I think that's OK at program initiation time, but is it fine if the IP changes while the unikernel is running?

hannesm:
if the IP changes while the unikernel is running, what should happen?  and how will you talk to it then?  if the IP changes, all connections will be broken anyways, thus we can just make a new IP stack!?!?

yomimono:
That sounds like pretty resounding "yes, I think?" to "is it fine?"  (Right now we don't do even close to the right thing with DHCP, just getting a lease once and assuming it's valid forever, so the question is rather hypothetical I admit.)

mor1:
Destroying the stack and creating a new on on address change sounds like
the right thing to do to me, so long as the effect of doing so is contained
to the stack -- ie., higher layer and application state in the unikernel
isn't / doesn't have to be affected.

If someone wants fancier behaviour I guess you'd look to extract the state
from the stack as required, and then recreate the new stack from that --
cf. the Irmin/ARP prototype, or the Irmin/TCP prototype...

On 13 July 2016 at 16:30, Mindy Preston <notifications@github.com> wrote:

> That sounds like pretty resounding "yes, I think?" to "is it fine?" (Right
> now we don't do even close to the right thing with DHCP, just getting a
> lease once and assuming it's valid forever, so the question is rather
> hypothetical I admit.)
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
[...]

PR #643, opened 2016-11-02
improvements to ipv4 configuration

yomimono:
TL;DR : make DHCP configuration of ipv4 a first-class citizen; remove `mode`, many type signature changes.

Details:
* Use charrua-client's Dhcp_ipv4 module to configure an ipv4 impl with dhcp.
Disambiguate between the two ipv4 modules by calling the static one
ipv4_keyed_conf (because it's configured through the key system) instead
of ipv4_conf .  generic_stack will now choose between a keyed
configuration and a DHCP-based one depending on the value of the `dhcp`
key.
* change the "network" key, defining which interface a V1_LWT.NETIF
  should attempt to attach to, to "interface".
* changing V1_LWT.IPV4's prefix type to Ipaddr.V4.Prefix.t,
  addressing a longstanding FIXME.  Ipaddr.V4.Prefix.t defines a
  network, not just a netmask, so change this name and provide the
  expected converters.
* removing the ip_config type, which was a GADT over ipv4 and ipv6.
  This forced ipv4 and ipv6 configurations to assume that the same
  number of addresses, netmasks, or gateways were reasonable, which is
  not the case; see #557
* creating an ipv6_config record type (and associated Mirage_key
  functions) that stores lists of addresses, netmasks, and gateways
* creating an ipv4_config record type (and associated Mirage_key
  functions) that stores only a single address, netmask, and gateway
* exposing both ipv4_config and ipv6_config in V1_LWT
* removing the netmask and gateway getter/setter functions from the
  module type IP.  ipv4 will expect to get this information either via a
  DHCP lease or at connect time via keys, and ipv6 largely determines it
  dynamically.
* removing direct_stack_config, as the stack should receive a configured
  ipv4 argument (whether via DHCP or static configuration isn't the
  stack's concern)
* putting the default values for statically-configured ipv4 modules into
  mirage_key, where they can be nicely exposed to the user via help.
* removing direct_stackv4_with_default_ip, direct_stackv4_with_dhcp, and
  direct_stackv4_with_static_ipv4.  Instead, let functoria figure out
  which implementation should be used based on the keys supplied and
  reasonable defaults via the function direct_stackv4, which requires
  more arguments -- network, ethernet, arpv4, and ipv4, rather than just
  network.
* requiring network, ethernet, and arp implementations as arguments to
  create_ipv4 instead of just a network implementation.  Doing so means
  that an ipv4 implementation can be configured and then handed off to
  stackv4.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -611,7 +611,7 @@ let dhcp time net = dhcp_conf $ time $ net
 let ipv4_of_dhcp dhcp ethif arp = ipv4_dhcp_conf $ dhcp $ ethif $ arp
 
 let create_ipv4
-    ?group _net etif arp =
+    ?group etif arp =
   let address = Key.V4.ip ?group () in
   let network = Key.V4.network ?group () in
   let gateway = Key.V4.gateway ?group () in
@@ -647,8 +647,7 @@ let ipv6_conf ?addresses ?netmasks ?gateways () = impl @@ object
 let create_ipv6
     ?(time = default_time)
     ?(clock = default_monotonic_clock)
-    ?group net { addresses ; netmasks ; gateways } =
-  let etif = etif net in
+    ?group etif { addresses ; netmasks ; gateways } =
   let addresses = Key.V6.ips ?group addresses in
   let netmasks = Key.V6.netmasks ?group netmasks in
   let gateways = Key.V6.gateways ?group gateways in
@@ -823,7 +822,7 @@ let dhcp_stack ?group time tap =
 let keyed_stack ?group tap =
   let e = etif tap in
   let a = arp e in
-  let i = create_ipv4 tap e a in
+  let i = create_ipv4 e a in
   direct_stackv4 ?group tap e a i
 
 let stackv4_socket_conf ?(group="") interfaces = impl @@ object
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -610,11 +610,19 @@ let ipv4_dhcp_conf = impl @@ object
 let dhcp time net = dhcp_conf $ time $ net
 let ipv4_of_dhcp dhcp ethif arp = ipv4_dhcp_conf $ dhcp $ ethif $ arp
 
-let create_ipv4
-    ?group etif arp =
-  let address = Key.V4.ip ?group () in
-  let network = Key.V4.network ?group () in
-  let gateway = Key.V4.gateway ?group () in
+let keyed_ipv4 ?group etif arp =
+  let default_address = Ipaddr.V4.of_string_exn "10.0.0.2" in
+  let default_network = Ipaddr.V4.Prefix.make 24 default_address in
+  let default_gateway = Some (Ipaddr.V4.of_string_exn "10.0.0.1") in
+  let address = Key.V4.ip ?group default_address in
+  let network = Key.V4.network ?group default_network in
+  let gateway = Key.V4.gateway ?group default_gateway in
+  ipv4_keyed_conf ~address ~network ~gateway () $ etif $ arp
+
+let create_ipv4 ?group { address; network; gateway } etif arp =
+  let address = Key.V4.ip ?group address in
+  let network = Key.V4.network ?group network in
+  let gateway = Key.V4.gateway ?group gateway in
   ipv4_keyed_conf ~address ~network ~gateway () $ etif $ arp
 
 type ipv6_config = {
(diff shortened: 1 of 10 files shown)


mato:
Not sure if this is the right place to comment, but since you're hacking in this area: Could we change the naming of the current "default" network interface from `tap0` to something more generic, say, `net0`? On non-UNIX targets the underlying interface (from the Mirage side) has nothing to do with `tap` so calling it `tap0` is confusing.

Separately, is there any need for interfaces to be named with strings? Wouldn't a numeric index suffice?

yomimono:
Whether or not it's the right place to comment, both of those points have been on my mind and I agree with them both.  I'll make separate issues for them.

talex5:
Note that simply renaming `tap0` isn't enough to make it generic. It really doesn't work on Xen (where the default interface is called `0`). Would be nice to have a `default_net` that does the right thing on every platform.

yomimono:
#644 and #645 for interface naming discussions.

hannesm:
highly appreciated! LGTM

lib/mirage.ml
```diff
@@ -866,15 +850,30 @@ let generic_stackv4
...
+  in
   if_impl
     Key.(pure ((=) `Socket) $ net_key)
     (socket_stackv4 ?group [Ipaddr.V4.any])
-    (if_impl
-       dhcp_key
-       (direct_stackv4_with_dhcp ?group tap)
-       (direct_stackv4_with_default_ipv4 ?group tap)
-    )
+    (pick_direct_stack ())
```
Drup: You can remove all the unit arguments in the above code.

lib/mirage.ml
```diff
@@ -866,15 +850,30 @@ let generic_stackv4
...
+  let dhcp_stack () =
+    let time = default_time in
+    let config = dhcp time tap in
+    let e = etif tap in
+    let (a : arpv4 impl) = arp e in
+    let i = ipv4_of_dhcp config e a in
+    direct_stackv4 ?group tap e a i
+  in
+  let pick_direct_stack () =
+    if_impl Key.(pure ((=) true) $ dhcp_key)
```
Drup: I'm sure you can simplify that :3

lib/mirage.ml
```diff
@@ -541,80 +538,107 @@ let ip = Type IP
...
-    gateways = [i "10.0.0.1"];
-  }
 
-let default_ipv4 ?group net = create_ipv4 ?group net default_ipv4_conf
+let dhcp time net = dhcp_conf $ time $ net
+let ipv4_of_dhcp dhcp ethif arp = ipv4_dhcp_conf $ dhcp $ ethif $ arp
 
-type ipv6_config = (Ipaddr.V6.t, Ipaddr.V6.Prefix.t list) ip_config
+let create_ipv4
+    ?group _net etif arp =
```
Drup: Since this is breaking compat, why not drop the `net` argument (and take a `etif` in `create_ipv6`) ?

Drup:
@yomimono very nice. I like the structure of the various devices. 

I don't understand your changes on the default of ipv4 keys. It's inconsistent between ipv6 and ipv4, and moving the default values to keys shouldn't change anything wrt to generated help. Also, it seems like `ipv4_config` is not used anymore at all. 

It seems to me that the `create_*` functions should expose their keys (via optional argument taking an `_ Key.value`, as usual).

yomimono:
Thanks a lot for your comments and review, @drup :) I was hoping you'd get a chance to take a look at this.

> I don't understand your changes on the default of ipv4 keys. 

There were previously defaults specified for ipv4 in many different places, which bothered me so I decided to move it to the place closest to where I'm usually looking for the default values for keys.  I was mistaken about the generated help though -- you're right that this was working properly before.

> Also, it seems like ipv4_config is not used anymore at all.

`ipv4_config` is exposed in `V1_LWT` and used for the DHCP client to return the bits of an obtained lease that are relevant for an ipv4 implementation.

> It's inconsistent between ipv6 and ipv4

See the conversation at https://github.com/mirage/mirage/issues/557 for ipv4 vs ipv6 configuration.  The types for ipv6 seem reasonable given the functionality that's there, but that wasn't the case for ipv4.  It doesn't make sense to do the same thing for these two cases IMO, and we didn't actually use the previous commonality between their types anywhere that I know of.

Drup:
I meant `Mirage.ivp4_config`. `Mirage.ipv6_config` is used in `create_ipv6` but as I said, I think we should just ditch that and take keys as optional parameters everywhere (and keep the functions in `Key.V4` parametrized by their default). It would be more flexible (without exposing non-existing capabilities).

Am I missing something ?

yomimono:
Ah, thanks @drup , that's quite a bit clearer -- the one missing something was me :)

yomimono:
With https://github.com/mirage/mirage/pull/643/commits/74bf3c78f17646cc6f62711d17fe20763e833c55 , it's possible to get ipv4 static config based on a passed `ipv4_config` in `config.ml`:

```
  let ip = Ipaddr.V4.of_string_exn "192.168.3.10" in
  let ipv4_config = {
    address = ip;
    network = Ipaddr.V4.Prefix.make 24 ip;
    gateway = Some (Ipaddr.V4.of_string_exn "192.168.3.1");
  } in
  let stack =
          let e = etif tap0 in
          let a = arp e in
          let ip = create_ipv4 ipv4_config e a in
          direct_stackv4 tap0 e a ip
```

Drup:
Good for me!

hannesm:
looks good for me, pls merge

PR #647, opened 2016-11-03
add solo5 libs now that we can install it together with xen

hannesm:
https://github.com/ocaml/opam-repository/pull/7736 @avsm


avsm:
wooop!

PR #655, opened 2016-11-04
add missing } to `connect` code generation for socket stack

yomimono:
Without this change, attempting to build stacks with `--net=socket` generates a broken `main.ml`.


hannesm:
looks good to me!

PR #657, opened 2016-11-05
fail if the user tries to build tracing with solo5 target

### LINKED ISSUE #650: trying to enable tracing for solo5 targets gets xen tracing stuff
We should error nicely instead.

yomimono:
It would of course be nicer to have tracing on solo5, but we don't.

Fixes #650 .

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1111,8 +1111,9 @@ let mprof_trace ~size () =
     method keys = [ Key.abstract key ]
     method packages = Key.pure ["mirage-profile"]
     method libraries =
-      Key.match_ Key.(value target) @@function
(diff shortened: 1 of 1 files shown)


hannesm:
I wonder whether we should adjust `mirage-solo5` to remove the dependency on `mirage-profile` and the calls... not clear what the benefit is to have those 4 lines calling out to mirage-profile... (see https://github.com/mirage/mirage-solo5/blob/d4f622fb25f7dbd997888242b5d12da75ce60df9/solo5/lib/main.ml#L73 plus one in time.ml).  opinions?  (I am always in favour of removing code which has no use (and since there's no mirage-profile working on solo5 in sight, this code qualifies))

PR #661, opened 2016-11-08
document the listen/_{tcp,udp}v4 behavior of stack

yomimono:
The results of calling `listen` on a` STACKV4.t`, and the interactions
with the listening callbacks registered via `STACKV4.listen_tcpv4` or
`listen_udpv4`, could stand to be better documented.  Here's a start.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -777,19 +777,31 @@ module type STACKV4 = sig
       UDPv4 [port] and immediately return.  If [port] is invalid (not
       between 0 and 65535 inclusive), it raises [Invalid_argument].
       Multiple bindings to the same port will overwrite previous
-      bindings, so callbacks will not chain if ports clash. *)
+      bindings, so callbacks will not chain if ports clash.
+      Note that [listen] must be called again after new callbacks are
+      registered in order to have traffic sent to them.
+      *)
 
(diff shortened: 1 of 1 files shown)


types/V1.mli
```diff
@@ -777,19 +777,31 @@ module type STACKV4 = sig
...
   val listen_tcpv4: t -> port:int -> TCPV4.callback -> unit
   (** [listen_tcpv4 t ~port cb] registers the [cb] callback on the
       TCPv4 [port] and immediatey return.  If [port] is invalid (not
       between 0 and 65535 inclusive), it raises [Invalid_argument].
       Multiple bindings to the same port will overwrite previous
-      bindings, so callbacks will not chain if ports clash. *)
+      bindings, so callbacks will not chain if ports clash.
+      Note that [listen] must be called again after new callbacks are
+      registered in order to have traffic sent to them.
+      *)
```
talex5: When did this change? You never used to have to do this...

yomimono: Let me make sure I'm right about this in the first place before I attempt to answer that question.

yomimono: I'm not right about this!  I'm clearly too confused to be attempting to resolve anyone else's confusion.

PR #666, opened 2016-11-12
rename M_pp to M_util (there'll be more than pp)

hannesm:
M_pp is a pretty bad name since I plan to move e.g. functions for Stats there (into submodules).  Thus rename early before we have too many dependencies... A travis build is good at https://github.com/mirage/mirage-dev/pull/208

Best be merged with https://github.com/mirage/mirage-tcpip/pull/264


samoht:
I am not very fond of the `M_util` name. What about keeping the same conventions as `Functiora` and:
- rename `V1` into `Mirage`
- rename `M_util` into `Mirage_runtime`

hannesm:
hmm, this might require some more thinking
- the `mirage` package provides a `mirage-runtime` module (located in `lib_runtime`)
- mirage and mirage-types are independent atm (on the opam level, there's no dependency between them)

--> does it sound better to leave mirage-types as the two interface files as before, and have runtime code in `mirage.runtime` (then mirage will depend on mirage-types, but that is fine I guess)?

hannesm:
turns out I do not understand the difference between `lib_runtime/mirage_runtime.ml{i}` and `lib/mirage_key.ml{i}` - the former contains the `log_threshold` (included in the latter), and both contain converters for ipv4/ipv6 addresses and prefixes... any insight (likely @Drup knows)?

PR #669, opened 2016-11-13
Use both lib/ and share/ for PKG_CONFIG_PATH

hannesm:
Thanks to @mato investigation [in mirage-solo5](https://github.com/mirage/mirage-solo5/pull/15), pc files should be installed into `share/pkgconfig` rather than `lib/pkgconfig` (which is not possible with `.install` files, but only with shell scripts and hacks).

The `lib/pkgconfig` path was hardcoded by several opam packages:
- [ ] https://github.com/mirage/mirage-solo5/pull/15 https://github.com/mato/mirage-solo5/pull/1
- [ ] https://github.com/mirage/ocaml-freestanding/pull/13
- [ ] https://github.com/Solo5/solo5/pull/122
- [x] https://github.com/pqwy/ocb-stubblr/pull/6
- [ ] https://github.com/talex5/mini-os/pull/2
- [ ] https://github.com/mirage/mirage-xen-minios/pull/21
- [ ] https://github.com/mirage/mirage-platform/pull/176
- [ ] https://github.com/mirage/mirage-dev/pull/210 (including: minios-xen, mirage-xen-minios, gmp-xen, gmp-freestanding, ocb-stubblr, mirage-xen-posix)

The xen part (minios-xen or mirage-xen-minios) somewhat breaks, and I cannot get much debug information from travis... someone with a test setup please test and fix it, should be straightforward.

A mirage-dev travis run with working unix + solo5 builds is available at https://github.com/mirage/mirage-dev/pull/209

Update: none of the PRs needs to be merged, ocb-stubblr and this PR here now use both `share/pkgconfig` and `lib/pkgconfig` as search PATH.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1395,10 +1395,13 @@ let rec expand_name ~lib param =
 
 (* Invoke pkg-config and return output if successful. *)
 let pkg_config pkgs args =
-  match Cmd.read "PKG_CONFIG_PATH=$(opam config var lib)/pkgconfig \
-    pkg-config %s %s" pkgs args with
-  | Ok s -> String.trim s
+  match Cmd.read "opam config var prefix" with
   | Error e -> failwith e
+  | Ok prefix ->
+    match Cmd.read "PKG_CONFIG_PATH=%s/lib/pkgconfig:%s/share/pkgconfig \
+                    pkg-config %s %s" prefix prefix pkgs args with
+    | Ok s -> String.trim s
+    | Error e -> failwith e
 
 (* Get the linker flags for any extra C objects we depend on.
  * This is needed when building a Xen/Solo5 image as we do the link manually. *)


talex5:
Compiled libraries are arch-specific, so I think the pc files should go in `lib`, not `share`. e.g. a 32-bit binary is not compatible with a 64-bit one (maybe it should even have it's own `lib` directory, like `lib64xen` or something). Users should probably check both `lib` and `share`, like the default search path does.

dbuenzli:
> which is not possible with .install files,

This is incorrect use `Pkg.lib` and a relative path to move up.

mato:
On 2016-11-13 20:00, Daniel Bünzli wrote:
>> which is not possible with .install files,
> 
> This is incorrect use `Pkg.lib` and a relative path to move up.

If that works then based on my reading of the OPAM documentation it's 
certainly not intended and probably a security issue.

dbuenzli:
> If that works then based on my reading of the OPAM documentation it's
> certainly not intended and probably a security issue.

As a matter of fact it currently works. Of course we'd like to get rid of these relative tricks, this was discussed here https://github.com/ocaml/opam/issues/2153 with a few proposals but then it was closed without mentioning any solution to the problem.

hannesm:
Thanks for all the feedback - I updated this PR to include both `lib/pkgconfig` and `share/pkgconfig` into the `PKG_CONFIG_PATH` -- thus other PRs can be dropped at your inconvenience.  This is in sync with ocb-stubblr (at https://github.com/pqwy/ocb-stubblr/commit/d30e86ac2fb1134feba75519293d98ea64a3b845)

hannesm:
I earlier failed to use my git thingies correctly, but _now_ the change is in that mirage will use both `share/pkgconfig` and `lib/pkgconfig`.  From the responses gathered here and on the mailing list, I guess this is the least controversial behaviour.  Thus, please merge once travis is happy (will enable the @maot `mirage-solo5` topkg PR to go through).  KTHXBYE

PR #672, opened 2016-11-14
need a String.trim after `opam config var prefix`

hannesm:
this fixes the `mirage configure` step you see failing in Travis.  sorry again for this breakage and the mirage-console-* breakages...


avsm:
thanks!

Issue #670, opened 2016-11-14
move ocamlfind package mirage-types.lwt to mirage-types-lwt


hannesm:
followup of https://github.com/mirage/mirage/pull/666#issuecomment-260281645
- @samoht says use `Mirage_utils` instead of `M_util`
- also proposes `V1` to `Mirage` (I second that name!)
- we now have mirage.runtime (for keys), and mirage-types (providing pp via M_util), should unify (into mirage.runtime, makes mirage depend on mirage-types (sounds fine!))
- unclear (at least to me, awaiting a brief description from @Drup) what should end up in `lib_runtime/mirage_runtime` and what in `lib/mirage_key`

hannesm:
via IRC drup says: mirage_runtime is only printers and parsers.  the logging stuff should be in the logging library.  code sharing is not meant to be between the modules accordin to drup.

hannesm:
the (more or less) obvious solution is as follows:
- make `functoria.runtime` an independent (to/from functoria) opam package
- make `mirage.runtime` an independent (to/from mirage) opam package (depends on mirage-types and functoria-runtime)
- get rid of `mirage-types.runtime`, has then moved to `mirage.runtime`

I plan to do this (hopefully tonight).  the mirage.runtime package is what ends up in every mirage unikernel (and has support for argument processing, pp, etc.).  Some of this is also present in the mirage package (since you want both at configuration and at runtime have converters which can extract an ipv4/6 address of a string).

by moving it into separate opam package, the intention will become more clear, and potential user packages have a much finer-grained dependency.

hannesm:
- mirage-types-lwt should install V1_LWT, rather than mirage-types.lwt

hannesm:
I updated the title to reflect what is left to do here (mirage-types.lwt -> mirage-types-lwt)

hannesm:
since #735 has landed, closing this.

PR #673, opened 2016-11-14
minor nits for pkg.ml

hannesm:


avsm:
looks like a `mirage-types` specific failure:

```
-> installed mirage-http.dev~mirage
Done.
[ERROR]
  command terminated with exit code 2
stderr: ocamlfind: Package `tls.mirage' not found
```

hannesm:
TLS recently switched to topkg (see https://github.com/mirage/mirage-dev/commit/addf3bf118f04059a88af8f9ca07b57ad85a78cd) (but the conditional when to compile with mirage support did not change -- mirage-types-lwt is still required (ipaddr is a dependency of mirage-types-lwt, and thus should have no effect)) -- mirage-dev builds fine, I cannot reproduce this issue... it only happens here in the mirage repo in the `PACKAGE=mirage-types` cases...

as mentioned yesterday on IRC, the `MODE` in the travis scripts here is also ignored (all runs are build for unix).

If @samoht (and/or others, @avsm) agrees to these nits here, I'd be happy to merge (since I've some more changes based on top this which will be in a separate PR at some point)

samoht:
LGTM

ideally we should remove the conditionals and build V1_LWT only when building `mirage-types-lwt` so I guess it's a step in the right direction.

hannesm:
this will require updates to lots of packages (but may solve the travis issue we currently see) -- atm V1_LWT is distributed as mirage-types.lwt, and mirage-types-lwt is a virtual package (not installing anything).  I put it as an action item onto #670 to not loose the idea.

PR #676, opened 2016-11-14
pkg-config: share before lib, also only request mirage-solo5 (which h…

hannesm:
…as ocaml-freestanding as dependency)

/cc @mato

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1399,7 +1399,9 @@ let pkg_config pkgs args =
   | Error e -> failwith e
   | Ok prefix ->
     let prefix = String.trim prefix in
-    match Cmd.read "PKG_CONFIG_PATH=%s/lib/pkgconfig:%s/share/pkgconfig \
+    (* the order here matters (at least for ancient 0.26, distributed with
+       ubuntu 14.04 versions): use share before lib! *)
+    match Cmd.read "PKG_CONFIG_PATH=%s/share/pkgconfig:%s/lib/pkgconfig \
                     pkg-config %s %s" prefix prefix pkgs args with
     | Ok s -> String.trim s
     | Error e -> failwith e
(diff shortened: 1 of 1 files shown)


mato:
LGTM

pqwy:
Are you *sure* that going against the normal behavior of `pkg-config` is a good solution to whatever problem you were having?

mato:
Sorry for the rather terse PRs. It's not normal behaviour. We're working around a bug in `pkg-config` which AFAICT was exposed by the use of multiple directories in `PKG_CONFIG_PATH` and resulted in the link order for Solo5 unikernels being wrong (`libmirage-solo5_bindings.a` was placed _after_ the ocaml-freestanding `asmrun.a`).

Best as I can tell(*) the bug only affects `pkg-config` versions prior to `0.28` from January 2013, unfortunately Ubuntu 14.04 (used by Travis) ships `0.26`(!). @hannesm has also verified that this does not break on FreeBSD which uses it's own `pkg-conf` reimplementation of `pkg-config`.

(*) I _think_ this is the commit that fixes the `pkg-config` bug: https://cgit.freedesktop.org/pkg-config/commit/?id=6ecf318c927ffe89a028b8b9cb70d0535673c550

`<rant>`
Our use of `pkg-config` is a house of cards in any case, since we depend on the _system_ `pkg-config` and then manipulate it's search path via `PKG_CONFIG_PATH`.

A possible improvement might be to ship our own build of `pkg-config` (or the FreeBSD implementation thereof) as part of [conf-pkg-config](https://github.com/ocaml/opam-repository/tree/master/packages/conf-pkg-config) with the correct paths already set and call it something like `opam-pkg-config`. This could then be replaced with a thin wrapper over per-package variables for OPAM 2.0.
`</rant>`

/cc FYI @avsm @yomimono @talex5

hannesm:
Thanks for this explanation @mato.  I would suggest to use it as is for now, and make a switch over to opam 2.0 once that is available and stable.  No need to work on improving the pkg-config situation in the meantime (as it is right now it seems to work on the platforms we're interested in).

mato:
I agree, unless it breaks again in some unforeseen way, in which case we'll need to fix it properly.

PR #678, opened 2016-11-15
pass MODE to container when testing

yomimono:
In the mainline `.travis-docker.sh`, the `MODE` environment variable isn't passed to the container which runs the tests, resulting in `make` in `mirage-skeleton` defaulting to `unix`.  (See https://github.com/ocaml/ocaml-ci-scripts/issues/136 .)  Use a branch which passes `MODE` to the container.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,4 +1,3 @@
 eval `opam config env`
-rm -rf mirage-skeleton
 git clone -b mirage-dev https://github.com/mirage/mirage-skeleton.git
-make -C mirage-skeleton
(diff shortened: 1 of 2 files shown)


yomimono:
This is fiddlier than expected; tests keep failing on `ocamlbuild` because of the presence of compiled objects from the `mirage-skeleton` test.  I'm iterating to bring the false negative rate down.

yomimono:
OK, there are still some test failures but they appear to be genuine problems building the project within the given container which we can try to address.  Merging, since authentically failing tests are more useful than inauthentically passing ones...

PR #681, opened 2016-11-15
mirage-runtime is now a custom opam package

hannesm:
This PR is best merged together with:
- [ ] https://github.com/mirage/mirage-tcpip/pull/266
- [ ] https://github.com/mirage/mirage-console-solo5/pull/6
- [ ] https://github.com/mirage/mirage-console/pull/50
- [ ] https://github.com/mirage/functoria/pull/80
- [ ] https://github.com/mirage/mirage-dev/pull/213

A travis build here (and in the other PRs) will fail.  But a successful one (where all those packages were pinned to my `#runtime` branches) is available at https://travis-ci.org/hannesm/mirage-dev/builds/176165464

The reasoning behind this change is motivated by @samoht (see #670 for more details).  In short:  instead of having a `mirage.runtime` and `mirage-types` providing code, use a real opam package.  While at it, @Drup noticed that functoria and functoria.runtime should be independent (thus I split them into two separate packages as well).

When I introduced code into the mirage-types package, I did there (instead of into mirage.runtime) because I don't want e.g. tcpip to depend on mirage (where mirage.runtime is a sublibrary of).  Now, having runtime code in both mirage-types (which is types only, hence the name) and in mirage.runtime is not good.  The code from `mirage-types/runtime` now moved (as `Mirage_pp` instead of `M_util`) into mirage-runtime.  Mirage-runtime depends on mirage-types.

The mirage package currently depends on the mirage-runtime package due to code reuse (from `lib_runtime/mirage_runtime.{ml,mli}` into `lib/mirage_key.{ml,mli}`).  It might be worth to get rid of this dependency/code sharing.

Additionally, this PR passes `dontlink unix str num threads` to unikernels (compiled for ukvm/virtio/xen), functoria-runtime, and mirage-runtime.


samoht:
That looks great, thanks!!

hannesm:
NB: mirage-types-lwt is _still_ a virtual package (and mirage-types.lwt is around), but this is the topic of a different PR.

PR #682, opened 2016-11-15
re-apply 077d852ef14cc5e00620a22a16b75f1152761f22, lost during qubes merge

hannesm:
I looked through a7cff94d354af6c0824f02a34718000369fd09d9 -- looks good otherwise (I couldn't find any other unintended merges)


hannesm:
reverted the `if_ is_unix` back to `match_` to avoid exceptions, and also removed `is_xen` from functoria_key (not used anymore, doesn't make too much sense in a world with solo5, there is `is_unix`)

PR #683, opened 2016-11-16
no need to call String.length before String.is_prefix since we're using Astring

hannesm:


avsm:
investigating some of these CI failures in https://github.com/mirage/ocaml-gnt/pull/24

PR #684, opened 2016-11-16
Rename mirage-clock-xen to -freestanding

mato:
Depends on mirage/mirage-clock#21.


hannesm:
this is a very sensible change in my opinion!

avsm:
yes, agreed

PR #690, opened 2016-11-18
add result types for FLOW & most `write`-like functions in networking components

yomimono:
Confession: I'm not very happy with this PR, but it and the corresponding upstream universe at https://github.com/yomimono/mirage-dev/tree/higherlevel-error are probably the best I'm going to do without more outside input.  Things it does:

* continues the pattern set by @hannesm in the earlier work adding result types to `write` functions in the `NETWORK` module type for `ETHERNET`, `IP`, `ARP`, `ICMP`, `UDP`, and `TCP`.  `TCP.create_connection` also now returns a result type.  As in the earlier PR, error types are defined in `V1` and printers in `Mirage_pp`.
* refactors the `FLOW` module type to use result types for both `read` and `write`.

Things it doesn't do:
* anything with the storage stack
* provide nice combinators
* anything with `input` types

PRs for users of these interfaces:
- [ ] channel - https://github.com/mirage/mirage-channel/pull/13
- [ ] conduit - https://github.com/mirage/ocaml-conduit/pull/166
- [ ] dns - https://github.com/mirage/ocaml-dns/pull/108
- [ ] mirage-console - https://github.com/mirage/mirage-console/pull/51
- [ ] mirage-flow - https://github.com/mirage/mirage-flow/pull/21
- [ ] mirage-net-unix - https://github.com/mirage/mirage-net-unix/pull/32
- [ ] mirage-qubes - https://github.com/mirage/mirage-qubes/pull/4
- [ ] tcpip - https://github.com/mirage/mirage-tcpip/pull/267
- [ ] vchan - https://github.com/mirage/ocaml-vchan/pull/90

Packages that are known to be affected, but for which there is no PR yet:
* ocaml-tls

### DIFF
--- a/lib_runtime/mirage_pp.ml
+++ b/lib_runtime/mirage_pp.ml
@@ -1,8 +1,37 @@
-
 let pp_console_error pp = function
   | `Invalid_console str -> Format.fprintf pp "invalid console %s" str
 
+let unspecified pp m = Format.fprintf pp "unspecified error - %s" m
+
 let pp_network_error pp = function
-  | `Msg message     -> Format.fprintf pp "unspecified error - %s" message
-  | `Unimplemented   -> Format.fprintf pp "operation not yet implemented"
-  | `Disconnected    -> Format.fprintf pp "device is disconnected"
+  | `Msg message   -> unspecified pp message
+  | `Unimplemented -> Format.fprintf pp "operation not yet implemented"
+  | `Disconnected  -> Format.fprintf pp "device is disconnected"
+
+let pp_ethif_error = pp_network_error
+
+let pp_ip_error = pp_ethif_error
+
+let pp_icmp_error pp = function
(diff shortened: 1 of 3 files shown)


hannesm:
The downside of this is that, as @talex5 remarked in https://github.com/mirage/mirage-console/pull/50#issuecomment-260770648, the errors are now closed.

Especially for `FLOW` this is a bit sad, since TLS has two kinds of failures: recoverable (such as `please reconfigure your required TLS version, or add some ciphersuite the other host understands`) and irrecoverable (no shared implemented ciphersuite).  When writing client code, I might want to handle the first by reconfiguring the stack and re-try to establish a connection, whereas the second is a hard failure.

Now, `FLOW` dictates which errors may happen: `` `Msg of string ``.  This needs some more consideration and a followup PR before 3.0 IMHO (I wouldn't count on myself having the time to work on that, unfortunately).

PR #627, opened 2016-10-19
Invoke pkg-config directly when generating Makefile

mato:
Rather than invoking pkg-config from within the generated Makefile,
invoke it directly at "configure" time.

Pros:
- This allows us to set variables like "LD" (see #623) sensibly.
- Generated Makefile contains the actual commands which will be run,
rather than call-outs to pkg-config whose output may change at build
time.

Cons:
- "mirage configure" needs to be re-run if dependent packages are
updated, to reflect any new pkg-config variables in the generated
Makefile.

@avsm @hannesm @yomimono Please review, taking into account the above pros/cons.

PR #691, opened 2016-11-19
[RFC] No opam, constrain package versions in unikernels

hannesm:
see https://github.com/mirage/functoria/pull/82 ... very much WIP (but compiles + runs :)

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1629,7 +1629,9 @@ let configure_makefile ~target ~root ~name ~opam_name ~warn_error info =
       append fmt "Makefile.ukvm:";
       append fmt "\t ukvm-configure $$(PKG_CONFIG_PATH=$(PKG_CONFIG_PATH) pkg-config --variable=libdir solo5-kernel-ukvm)/src/ukvm $(UKVM_MODULES)";
       newline fmt;
-      append fmt "include Makefile.ukvm";
+      append fmt "ukvm-bin:: Makefile.ukvm";
+      append fmt "\t$(MAKE) -f Makefile.ukvm $@";
+      newline fmt;
       append fmt "build:: main.native.o ukvm-bin";
       tst_pkg_config pkg_config_deps;
       append fmt "\t$(LD) $(PRE_LD_FLAGS) \\\n\
@@ -1639,7 +1641,8 @@ let configure_makefile ~target ~root ~name ~opam_name ~warn_error info =
         name ;
       append fmt "\t@@echo Build succeeded";
       newline fmt;
-      append fmt "clean:: ukvm-clean\n\
+      append fmt "clean:: Makefile.ukvm\n\
+                  \t$(MAKE) -f Makefile.ukvm ukvm-clean\n\
                   \tocamlbuild -clean\n\
                   \t$(RM) Makefile.ukvm";
       R.ok ()
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -45,8 +45,7 @@ let qrexec_qubes = impl @@ object
   val name = Name.ocamlify @@ "qrexec_"
   method name = name
   method module_name = "Qubes.RExec"
-  method packages = Key.pure ["mirage-qubes"]
-  method libraries = Key.pure ["mirage-qubes"]
+  method packages = Key.pure [package "mirage-qubes"]
   method configure i =
     match Key.(get (Info.context i) target) with
     | `Qubes -> Result.Ok ()
@@ -71,8 +70,7 @@ let gui_qubes = impl @@ object
   val name = Name.ocamlify @@ "gui"
   method name = name
   method module_name = "Qubes.GUI"
-  method packages = Key.pure ["mirage-qubes"]
-  method libraries = Key.pure ["mirage-qubes"]
+  method packages = Key.pure [package "mirage-qubes"]
   method configure i =
     match Key.(get (Info.context i) target) with
     | `Qubes -> Result.Ok ()
@@ -95,8 +93,7 @@ let qubesdb_conf = object
   method ty = qubesdb
   method name = "qubesdb"
   method module_name = "Qubes.DB"
-  method libraries = Key.pure ["mirage-qubes"]
-  method packages = Key.pure [ "mirage-qubes" ]
+  method packages = Key.pure [ package "mirage-qubes" ]
   method connect _ modname _args =
      Fmt.strf
 "@[<v 2>\
@@ -114,8 +111,8 @@ let io_page_conf = object
   method ty = io_page
   method name = "io_page"
   method module_name = "Io_page"
-  method libraries = Key.(if_ is_unix) ["io-page.unix"] ["io-page"]
-  method packages = Key.pure [ "io-page" ]
+  method packages =
+    Key.(if_ is_unix) [package ~sublibs:["unix"] "io-page"] [package "io-page"]
 end
 
 let default_io_page = impl io_page_conf
@@ -135,14 +132,15 @@ let default_time = impl time_conf
 type pclock = PCLOCK
 let pclock = Type PCLOCK
 
-let posix_clock_conf = object (self)
+let posix_clock_conf = object
   inherit base_configurable
   method ty = pclock
   method name = "pclock"
   method module_name = "Pclock"
-  method libraries =
-    Key.(if_ is_unix) ["mirage-clock-unix"] ["mirage-clock-freestanding"]
-  method packages = self#libraries
+  method packages =
+    Key.(if_ is_unix)
+      [package "mirage-clock-unix"]
+      [package "mirage-clock-freestanding"]
   method connect _ modname _args =
     Printf.sprintf "%s.connect ()" modname
 end
@@ -152,14 +150,15 @@ let default_posix_clock = impl posix_clock_conf
 type mclock = MCLOCK
 let mclock = Type MCLOCK
 
-let monotonic_clock_conf = object (self)
+let monotonic_clock_conf = object
   inherit base_configurable
   method ty = mclock
   method name = "mclock"
   method module_name = "Mclock"
-  method libraries =
-    Key.(if_ is_unix) ["mirage-clock-unix"] ["mirage-clock-freestanding"]
-  method packages = self#libraries
+  method packages =
+    Key.(if_ is_unix)
+      [package "mirage-clock-unix"]
+      [package "mirage-clock-freestanding"]
   method connect _ modname _args =
     Printf.sprintf "%s.connect ()" modname
 end
@@ -174,8 +173,7 @@ let stdlib_random_conf = object
   method ty = random
   method name = "random"
   method module_name = "Stdlibrandom"
-  method packages = Key.pure ["mirage-stdlib-random"]
-  method libraries = Key.pure ["mirage-stdlib-random"]
+  method packages = Key.pure [package "mirage-stdlib-random"]
   method connect _ modname _args =
     Printf.sprintf "Lwt.return (%s.initialize ())" modname
 end
@@ -191,6 +189,7 @@ let enable_entropy, is_entropy_enabled =
   let g () = !r in
   (f, g)
 
+(* XXX: needs to move into the makefile (depend and/or build)!
 let check_entropy libs =
   Cmd.OCamlfind.query ~recursive:true libs
   >>| List.exists ((=) "nocrypto")
@@ -202,7 +201,7 @@ let check_entropy libs =
        device. You can do so by adding ~deps:[abstract nocrypto] \
        to the arguments of Mirage.foreign."
   else R.ok ()
-
+*)
 let nocrypto = impl @@ object
     inherit base_configurable
     method ty = job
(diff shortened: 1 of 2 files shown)


hannesm:
there's some stuff which now needs (in contrast to what I thought in #627) to be delayed until build time, such as nocrypto check (NB: the only user of OCamlfind.query/installed from Functoria, which can thus go away as well -- shouldn't be a concern to functoria)... at configuration time we cannot promise that all dependent packages are installed (thus calling out to ocamlfind is not very robust).

I'd suggest to leave the fat and crunch at the configuration step for now (requires the user to have crunch/fat around when configuring the unikernel).  This has at least the property that there are not more source files generated at build time (and mirage really needs to run only once).

But stuff like pkg-config should be defered to later (into the makefile for when build is called).  also, the depend target should be updated to do sth along the lines `opam install --deps-only <myopam>`.

see https://github.com/mirage/mirage-skeleton/pull/198

hannesm:
Version constraints of a unikernel: mirage configure emits automatically constraints into the opam file for `mirage-types` and `mirage-runtime` (always: `>= current-version & < next-major`).  When you change a types or runtime API, if it is backwards compatible (extension of the current things), just bump the minor version.  For non-backwards compatible changes, we can just bump the major version.  old configured unikernels will not even been tried to compile with the new types or runtime. 

If you like, you can also specify constraints in your very own `config.ml` on mirage-types or mirage-runtime.  Version constraints get merged: maximum version of all mins,  minimum version of all max versions.

Thus, `package ~min:"3.0.5" ~max:"3.5.0" "mirage-types"` and `package ~min:"3.2.0" ~max:"4.0.0" "mirage-types"` will result into `package ~min:"3.2.0" ~max:"3.5.0" "mirage-types"`.

yomimono:
First off, this is excellent and I really appreciate this PR.  Thank you very much for putting this (very badly needed) set of patches together!

I think there's some room for improvement in the user experience, though:

```
o-opam🐫  ((detached from hannesm/no-opam)) mirageos:~/mirage-skeleton/console$ opam pin
functoria.dev~mirage            git  /home/user/functoria#no-opam
functoria-runtime.dev~mirage    git  /home/user/functoria#no-opam
mirage.dev~mirage               git  /home/user/mirage#no-opam   
mirage-runtime.dev~mirage       git  /home/user/mirage#no-opam   
mirage-types.dev~mirage         git  /home/user/mirage#no-opam   
mirage-types-lwt.dev~mirage     git  /home/user/mirage#no-opam   
no-opam🐫  ((detached from hannesm/no-opam)) mirageos:~/mirage-skeleton/console$ mirage configure -t virtio
Fatal error: exception (Failure
   "command terminated with exit code 1\
  \nstderr: Package solo5-kernel-virtio was not found in the pkg-config search path.\
  \nPerhaps you should add the directory containing `solo5-kernel-virtio.pc'\
  \nto the PKG_CONFIG_PATH environment variable\
  \nNo package 'solo5-kernel-virtio' found\
  \n")
```

I get a `mirage-console.opam` but an incomplete `main.ml` and `Makefile`.

The automatic upper bound interacts poorly with pins that specify a high package number (coincidentally, anything pinned in a switch that knows about `mirage-dev` at the moment, as those version numbers are all `dev~mirage` which is higher than any number).

Basic `opam` question -- is `opam install --deps-only <file>` supposed to work?   For a generated `console-virtio.opam`, I get a complaint that there's no such package and failure when trying to run that command.  I skimmed TFM for `opam install` but maybe missed what would've helped me.  

`opam pin add console-virtio .` in the directory containing the `opam` file works, but then I have a `console-virtio` package installed locally.  It doesn't hurt anything, but it's rather nonsensical IMO, and this introduces the potential for namespace collisions between the generated name of the opam file and the opam package universe.

samoht:
The pb with crunch/fat is that it is actually a configuration that you want to make available at deployment time (e.g. if you want to embed some secrets). So it makes sense that they don't fit very well in the refactoring, as we used to put together stuff which were supposed to happen a configuration/build/deployment time.

Not sure how to solve that properly, maybe we need to had an other stage in the compilation pipeline, but we can see later. The current simplification is way better than it was before ;-) Maybe just generate the script to generate the fat/crunch images so that they can be run later?

hannesm:
@yomimono thanks for your feedback.  sorry for pushing an unfinished changeset which doesn't completely work yet (and all I tested was `opam lint` on the generated opam file)

- incomplete main.ml/Makefile:  as mentioned in this PR, we'll need to defer pkg-config calls again to the Makefile (rather than doing them in mirage during configure time) - will update
- version constraints: there is no need for mirage to emit any version constraints.  reasonable unikernels will specify their mirage-types lower and upper bounds - will update
- opam troubles: I haven't tried, but my idea was to do (in `make depend`) `opam pin -n add <name> <name.opam>; opam install --deps-only <name>; opam pin remove <name>` - will update
- crunch/fat  (@samoht): I have two workflows in mind: *dev* where it would be nice as it is atm (generate the stuff during mirage configure, leave it there forever, requires fat/crunch being available at configure time); *deployment* where the build host does not have access to secret data.  there should be a vm image produced which is postprocessed (`ld`) to add arbitrary binary data (as an elf section).  while mirage currently uses *dev*, we've to find a good solution for *deployment* (while keeping the *dev* working for development)

hannesm:
I updated this PR:
- no more upper bound emitted by mirage
- the name for the opam file is now mirage-unikernel-<myunikernel>-<target>`
- `depend` got tweaked: `opam pin add <name> . ; opam install --deps-only <name> ; opam pin remove <name>`
- the `pkg_config` calls went back into the Makefile (instead of having them in OCaml code).

This works pretty much for me.  I don't have any clue whether something (and what and how) is supposed to call `opam depext`... but others (@avsm?) will shed some light into that.

With BSD make, the `sed -e s|@|$(OPREFIX)/lib/|g` does not work (it does with GNU make), where `OPREFIX=$$(opam config var prefix)`.  The reason is that `sed` does not evaluate the opam expression... maybe @mato has an answer?

hannesm:
and this lets me run `mirage configure -t xen` on my FreeBSD laptop, which will never be able to compile and execute such unikernels ;)

lib/mirage.ml
```diff
@@ -1814,19 +1757,21 @@ module Project = struct
...
-        | `Ukvm -> "solo5-kernel-ukvm" :: "mirage-solo5" :: l
-        | `Unix | `MacOSX -> "mirage-unix" :: l
-
-      method libraries =
-        let l = [ "mirage-runtime"; "mirage-types"; "mirage-types.lwt" ] in
+        let l = [
+          (* XXX: use %%VERSION_NUM%% here instead of hardcoding a version? *)
+          package "lwt";
+          package ~ocamlfind:[] "mirage-types-lwt";
+          package ~sublibs:["lwt"] ~min:"3.0.0"  "mirage-types";
```
Drup: Both `mirage-types-lwt` and `mirages-types.lwt` ?

Drup:
The opam file generation should be in functoria directly. It's not mirage specific and it's not doing any extra work.

> there's some stuff which now needs (in contrast to what I thought in #627) to be delayed until build time, such as nocrypto check (NB: the only user of OCamlfind.query/installed from Functoria, which can thus go away as well -- shouldn't be a concern to functoria)... at configuration time we cannot promise that all dependent packages are installed (thus calling out to ocamlfind is not very robust).

Yes++. You have all my support on that one. :D

lib/mirage.ml
hannesm: yes. no.  soon to be changed... atm there is mirage-types which provides V1 and V1_LWT in its lwt subpackage, and mirage-types-lwt is a virtual package which gathers all deps to get mirage-types.lwt.  pretty insane atm, should be better soon...

hannesm:
the ocamlfind code from functoria is already removed in functoria's PR...

hannesm:
I don't know when to merge (rough times ahead), an example unikernel over at logs-syslog works https://travis-ci.org/hannesm/logs-syslog/jobs/177822764 ci-script: https://github.com/hannesm/logs-syslog/blob/dev/.travis-test.sh (basically `mirage configure -t XX` followed by `make depend` and `make`) -- I'm pretty certain mirage, mirage-dev, mirage-skeleton need updates in their ci scripts

unfortunately I won't be able to spend much time on this in the upcoming weeks, so please go ahead and do what is good.

hannesm:
in the bigger picture, I guess we should get rid of the `Makefile`, and just reimplement the bits and pieces we need in OCaml (using topkg?).  Benefits would be: types! no need to think about escaping too much, etc. :)  There'd also be no need to generate code, only some bits of input for the build system (basically the PKG and LIBS lists).

hannesm:
There's a full travis build at https://travis-ci.org/hannesm/mirage-dev/builds/177981102 which has some changes:
- https://github.com/mirage/mirage-skeleton/pull/198 calls depend as well
- https://github.com/hannesm/mirage-dev/commit/a7b67fade1069b3440f3c770c68bc9bfd4f22e30 now installs `crunch` and `fat-filesystem` upfront (otherwise e.g. `static_website` fails due to executing crunch at configure phase)
- mirage and functoria PRs

Unix/Xen/virtio works fine, ukvm does not:

```
make[1]: Entering directory `/home/travis/build/hannesm/mirage-dev/mirage-skeleton/console'
Makefile:46: Makefile.ukvm: No such file or directory
ukvm-configure $(PKG_CONFIG_PATH=/home/travis/.opam/4.03.0/share/pkgconfig:/home/travis/.opam/4.03.0/lib/pkgconfig pkg-config --variable=libdir solo5-kernel-ukvm)/src/ukvm 
/bin/sh: 1: ukvm-configure: not found
```

can we delay the ukvm thingies until after make depend finished (I'm not able to test ukvm due to lack of systems, also will be in transit for the next few days)?

hannesm:
update: travis is happy, see comment above

yomimono:
I probably don't have time to review this properly today but I'm very excited to see it.  Thanks a ton for this, @hannesm! :D

mato:
Same here, this is great but still need to test and review it fully. Hopefully tomorrow, fingers crossed. Thanks @hannesm!

mato:
Initial comments from a first try of this, using `ukvm` as a target:
1. The pinning command in `make depend` asks for interactive confirmation ("Package mirage-unikernel-... does not exist, create as a NEW package?"). Can we remove this check?
2. One of the better things about generating all of the `Makefile` before this change was that it was obvious what the actual commands invoked are. Now we're back to stuff like this:
  ````
ld $(PKG_CONFIG_PATH=/home/mato/.opam/mirage-dev-4.04.0/share/pkgconfig:/home/mato/.opam/mirage-dev-4.04.0/lib/pkgconfig pkg-config --variable=ldflags solo5-kernel-ukvm) \
  _build/main.native.o \
  $(ocamlfind query -r -format '-L%d %(freestanding_linkopts)' -predicates native conduit conduit.mirage functoria-runtime io-page lwt magic-mime mirage-bootvar-solo5 mirage-clock-freestanding mirage-http mirage-logs mirage-net-solo5 mirage-runtime mirage-solo5 mirage-stdlib-random mirage-types mirage-types.lwt nocrypto nocrypto.mirage tcpip tcpip.arpv4 tcpip.ethif tcpip.icmpv4 tcpip.ipv4 tcpip.stack-direct tcpip.tcp tcpip.udp tls tls.mirage uri | sort -u | sed -e 's|@|/home/mato/.opam/mirage-dev-4.04.0/lib/|g') $(PKG_CONFIG_PATH=/home/mato/.opam/mirage-dev-4.04.0/share/pkgconfig:/home/mato/.opam/mirage-dev-4.04.0/lib/pkgconfig pkg-config --static --libs mirage-solo5) \
  -o mir-https.ukvm
````
  which IMO makes it much harder to understand what is going on.
3. Running `make` a second time after a successful build gives me the following errors:
````
SANITIZE: a total of 2 files that should probably not be in your source tree
  has been found. A script shell file
  "/home/mato/solo5/mirage-skeleton/static_website_tls/_build/sanitize.sh" is
  being created. Check this script and run it to remove unwanted files or use
  other options (such as defining hygiene exceptions or using the -no-hygiene
  option).
IMPORTANT: I cannot work with leftover compiled files.
ERROR: Leftover object files:
  File ukvm-net.o in _build-ukvm has suffix .o
  File ukvm-core.o in _build-ukvm has suffix .o
Exiting due to hygiene violations.
Compilation unsuccessful after building 0 targets (0 cached) in 00:00:00.
Makefile:40: recipe for target 'main.native.o' failed
make: *** [main.native.o] Error 1

````

hannesm:
thanks for testing and your feedback, @mato

1. we can add the `-y` parameter to the calls to opam (or export OPAMYES=1) in the generated Makefile.
2. I don't have a solution for this.  we do agree that `configure -> install dependencies -> build` is the right workflow (in contrast to earlier times), don't we?
3. there are three solutions (AFAICS):
- call ocamlbuild without the recursive flag (see #537) -- I don't agree with the argument in there (and if the warning is misleading, we should find another way to disable it)
- require a `-$(MAKE) clean` in the build rule in the Makefile
- build ukvm stuff in `_build` as well

hannesm:
superseeded by #703

PR #692, opened 2016-11-20
Do not use LD from pkg-config, but just the system one

hannesm:
This hack was needed at some time on FreeBSD, but is no longer -- cleaning up

This complements https://github.com/Solo5/solo5/pull/123 waiting for @mato


mato:
Merging for solo5/solo5#123 despite broken Travis (acked by @hannesm)

PR #701, opened 2016-11-22
docs: add lru-cache and notty

avsm:


Dockerfile.doc
```diff
@@ -50,6 +50,7 @@ RUN opam depext -uivj 3 \
   jenga \
   logs \
   logs-syslog \
+  lru-cache \
```
pqwy: This thing is called [`lru`](https://github.com/pqwy/lru) since a couple of minutes ago. :smile:

PR #702, opened 2016-11-23
[WIP] Refactor CHANNEL types to be exception-less and use `result`

avsm:
Changes:
- remove `read_stream` and `io_stream` as Lwt_stream is difficult to
  combine with error handling
- remove `read_until` and only expose the line handling functions.
  I'm also considering moving the line-handling functions into a
  separate signature since most consumers of CHANNEL only want the
  buffering logic.
- no more exceptions, just use `result` and propagate `FLOW.error`
  in return values.

Dependent PRs:
- [ ] mirage/mirage-channel#14
- [ ] mirage/mirage-http#30
- [x] yomimono/charrua-client#6
- [ ] mirage/ocaml-git#172

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: c
-install: wget https://raw.githubusercontent.com/yomimono/ocaml-ci-scripts/pass-mode/.travis-docker.sh
+install: wget https://raw.githubusercontent.com/avsm/ocaml-ci-scripts/extra-env/.travis-docker.sh
 script: bash -ex .travis-docker.sh
 sudo: required
 services:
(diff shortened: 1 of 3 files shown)


avsm:
related: https://github.com/mirage/mirage-channel/pull/14

now fixing up the other CHANNEL consumers like mirage-http

avsm:
Also https://github.com/mirage/mirage-http/pull/30

avsm:
requires https://github.com/yomimono/charrua-client/pull/6 on osx to work

avsm:
https://github.com/mirage/ocaml-git/pull/172

samoht:
I am not super fond of using polymorphic variants (mainly for the performance impact of allocating blocks in the successful case) but it is not worth that it was before so LGTM.

Issue #700, opened 2016-11-22
generation of Makefile for building unikernels


hannesm:
is brittle, replace with some OCaml code based on bos/topkg?  not needed for 3.0, but a nice to have afterwards...  would also ease integration of other things, such as

![bytes](https://www.cl.cam.ac.uk/~hm519/mirage-console-bytes.svg)

![deps](https://www.cl.cam.ac.uk/~hm519/mirage-console-xen.svg)

(via https://hannes.nqsb.io/Posts/BottomUp)

mor1:
(All IMO.)

No objection so long as invocation is as straightforward (`make` or `build` or similar; `mirage build` if you must), and it doesn't end up involving a twisty-turny maze of scripts (or `_tags` and `myocamlbuild.ml` or all the other many things that the many OCaml build systems seem to end up introducing) and and other such goop. Being able to see reasonably directly which commands are being invoked, as with a `Makefile`, is really useful -- moving to a simple `make` was such an improvement over the thousand and one ways that were explored prior to that...

(Ok, maybe I exaggerate slightly :) but please make it as simple to use. Also, what's actually brittle about the current practice?)

hannesm:
evaluation order, differences between BSD make and GNU make, see https://github.com/mirage/mirage/pull/691#issuecomment-261815803.  The Makefile is just a string from the OCaml perspective - there's no type checking happening.  I'm pretty sure if we have whitespaces or unicode characters in anything, it will all break.

The build tool in my mind should be very independent from mirage code wise.  Just calling out to `pkg-config`, `ocamlbuild`, and `ocamlfind`.  I guess in history the configure and build phases were not strictly separated, as is with #691, thus leading to confusion and unneeded complexity.

mato:
A partial step to improving this would be to use some kind of template engine type thingy for the Makefile generation rather than just appending to a string with `fmt`.

hannesm:
now that #703 is in, what is left is to integrate generation of pictures (but these are new feature enhancements, the core of this issue is done)

Issue #596, opened 2016-09-15
mirage configure should not install depext by default


avsm:
Also depext invocation should be `-ui` so that the package manager is updated, or it'll probably not be consistent for Debian/Ubuntu.

The common case for configuration is to not install depexts and it is quite slow, so reversing the default makes sense for 3.0

hannesm:
done by #703 AFAICS

yomimono:
I think this isn't quite right, as it seems we don't do the right thing when we need depexts - see https://travis-ci.org/yomimono/mirage/jobs/180149763#L1607 .

mato:
Confirmed -- I also saw this while updating unikernel-runner to the new order. Have worked around it by installing depext packages manually for now.

hannesm:
then the question arises whether the generated `Makefile` should include an `opam install depext` in its `depend` rule?

mato:
No, it needs to do the equivalent of `opam depext -i` but for the `--deps-only` that Mirage wants. I'll see if I can figure out what that is.

mato:
See #718

PR #703, opened 2016-11-25
New order

### LINKED ISSUE #696: enforce the right minimum OCaml version
We need to enforce a proper minimum OCaml version in MirageOS 3 before it comes out of beta:

- [ ] we no longer test <4.03 on Travis so if it is 4.02 then we need to add it back
- [ ] `-ocaml-version-check`in `check_ocaml_version` still asks for 4.02.3+
- [ ] the `opam` files also require 4.02.

So if we leap to 4.03 as the minimum version we need to fix up the ocaml version checks, and if we stay on 4.02 we need to restore test cases.  Personally I am in favour of shifting up to 4.03 given the rapid pace of OCaml releases recently.

CC #695 also for 4.02 compilation fixes

hannesm:
requires https://github.com/mirage/functoria/pull/84, on top of #691

relies on https://github.com/Solo5/solo5/pull/137 which emits ukvm build products into `_build/ukvm` instead of `_build-ukvm`

The generated Makefile is much more simplicistic.  There is a new `build` subcommand, which calls out to ocamlbuild, and in a link step to ld (apart from Unix target).

~~~As mentioned in https://github.com/mirage/functoria/pull/84, `clean` is somehow broken, but `configure` and `build` are tested on unix.~~~

Superseeds #691 fixes #696 (we use opam, use its `available` and `ocaml-version`!) #700 #596 superseeds #586

Feedback welcome.


hannesm:
There is a working mirage-skeleton travis job at https://travis-ci.org/hannesm/mirage-dev/builds/179095513

Depends on the following PRs:
- #703 
- https://github.com/mirage/functoria/pull/84
- https://github.com/mirage/mirage-dev/pull/230
- https://github.com/mirage/mirage-skeleton/pull/200

and is ready to.  as @mato mentioned in the previous PR, ukvm make (multiple times) looks a bit flaky, not sure what should be done there.

as mentioned by @drup, the info stuff needs some more thoughts (now that we have packages, libraries, and version constraints).  since there's now a `build` phase, that should be straightforward to integrate.

lib/mirage.ml
```diff
@@ -20,13 +20,14 @@ open Astring
...
-module Log = Functoria_app.Log
 module Codegen = Functoria_app.Codegen
 
+let src = Logs.Src.create "mirage" ~doc:"mirage library"
+module Log = (val Logs.src_log src : Logs.LOG)
+
 include Functoria
 
-let get_target i = Key.(get (Info.context i) target)
+let target i = Key.(get (Info.context i) target)
```
Drup: Please keep `get_target` for that one, it overlaps with the key otherwise. :)

lib/mirage.ml
```diff
@@ -1399,393 +1395,381 @@ let configure_main_xl ?substitutions ext i =
...
+        newline fmt;
+        append fmt "-include Makefile.user";
+        R.ok ()) ()
+  with
+  | Ok b -> b
+  | Error _ -> R.error_msg "couldn't open output channel for Makefile"
+
+let clean_makefile () = Bos.OS.File.delete Fpath.(v "Makefile")
+
+let configure_myocamlbuild () = Bos.OS.File.write Fpath.(v "myocamlbuild.ml") ""
```
Drup: Why create an empty myocamlbuild.ml ? It prevents the user from having one.

hannesm: to avoid getting `ocamlbuild` error that it might be not an OCaml project and called with `-r` (see https://github.com/ocaml/ocamlbuild/blob/756fe48e76a874e5d7e20de637024c0379355d1b/src/options.ml#L373), discussion at https://github.com/mirage/mirage/pull/537 and https://github.com/mirage/mirage/pull/691#issuecomment-262559815.  But sure, that should be guarded.

lib/mirage.ml
hannesm: done

hannesm:
customisation:
- `Makefile.user` is included (not sure what could be in there, pls tell me ;)
- `_tags`
- `myocamlbuild.ml`
- custom build devices (see the block example https://github.com/hannesm/mirage-skeleton/blob/61533f871b69dadb61915607b5d48292d8b3fd62/block/config.ml#L9)

Which code to move? since functoria doesn't have a concept of targets, and both compile and link steps are target specific, I don't know which pieces you'd want to move over.  I've the feeling that we might want to phase configure out (since no user-devices should need to specify it, they'll much more likely want to use the build phase).  But certainly `mirage` generates a lot of files during configure (which have mirage-specific stuff in there), opam, Makefile, xe, xl, xl.in, libvirt.xml, ...

Drup:
> Which code to move?

The concept of makefile that you have currently it's mostly generic. `make depend`, opam stuff, etc. That's not dependent on the target. Anyway, as long as nobody else uses functoria, it's a bit abstract. :)

mato:
I have a possible fix for the problem with `ukvm` not building the 2nd time round. Adding the following to `_tags` makes it work for me:

````
<_build-ukvm>: -traverse
````

Perhaps we can generate a `_tags` file as part of `mirage configure`? If we want to make it user-customizable then we can include `_tags.user` from it.

One nit: With the current implementation I don't see the final link command `ld ...` output when running `make`, is this intentional?

mato:
Another nit -- running `make` multiple times seems to always rebuild everything?

hannesm:
On 28/11/2016 16:00, Martin Lucina wrote:
> I have a possible fix for the problem with `ukvm` not building the 2nd time round. Adding the following to `_tags` makes it work for me:
> 
> ````
> <_build-ukvm>: -traverse
> ````

ah, nice. I added this (unfortunately it seems to have to be in both
functoria and mirage ocamlbuild invocation, is there another solution?)

> Perhaps we can generate a `_tags` file as part of `mirage configure`? If we want to make it user-customizable then we can include `_tags.user` from it.
> 
> One nit: With the current implementation I don't see the final link command `ld ...` output when running `make`, is this intentional?

did you try mirage build -v -v ?

> Another nit -- running `make` multiple times seems to always rebuild
everything?

fixed in functoria (only delete config.*, not entire _build)

mato:
> ah, nice. I added this (unfortunately it seems to have to be in both functoria and mirage ocamlbuild invocation, is there another solution?)

Doesn't seem to have had any effect here...did you try with `_tags` removed? None gets generated.

> did you try mirage build -v -v ?

I did not. Works, thanks. Possible improvement would be to display the commands executed in a more cut-and-paste format, would help with debugging/tweaking.

> fixed in functoria (only delete config.*, not entire _build)

Confirmed fixed.

Minor nit: `ukvm-configure` gets invoked on every `mirage build`, which regenerates `Makefile.ukvm`, which recompiles `ukvm`. It _might_ be a better idea to only invoke `ukvm-configure` if `Makefile.ukvm` is not present? However, if you reinstall `ukvm` then you probably want to reinvoke it but that seems an edge case...?

Also, how hard would it be for `mirage build` (and indeed any other commands) take the default value of `-t` from that passed to a previous run of `mirage configure`? Right now `-t` always defaults to `unix`.

mato:
Clarification: If I ran `mirage configure -t X`, does it make any sense for `mirage build -t Y` to run at all? Shouln't it just error out?

hannesm:
On 28/11/2016 16:42, Martin Lucina wrote:
>> ah, nice. I added this (unfortunately it seems to have to be in both functoria and mirage ocamlbuild invocation, is there another solution?)
> 
> Doesn't seem to have had any effect here...did you try with `_tags` removed? None gets generated.

Could you try to, instead of using a _tags file, pass -tag-line
"<_build-ukvm>: -traverse" to ocamlbuild directly?

>> did you try mirage build -v -v ?
> 
> I did not. Works, thanks. Possible improvement would be to display the commands executed in a more cut-and-paste format, would help with debugging/tweaking.

done.

>> fixed in functoria (only delete config.*, not entire _build)
> 
> Confirmed fixed.
> 
> Minor nit: `ukvm-configure` gets invoked on every `mirage build`, which regenerates `Makefile.ukvm`, which recompiles `ukvm`. It _might_ be a better idea to only invoke `ukvm-configure` if `Makefile.ukvm` is not present? However, if you reinstall `ukvm` then you probably want to reinvoke it but that seems an edge case...?
> 
> Also, how hard would it be for `mirage build` (and indeed any other commands) take the default value of `-t` from that passed to a previous run of `mirage configure`? Right now `-t` always defaults to `unix`.

That's why I still generate a Makefile.  The build step now needs a
target.  To solve this, some persistent data (the target of the last
configure run) needs to be stored somewhere and read back in.  Didn't
want to go through that hazzle right now.  But it might be worth doing.
Maybe open a separate issue for this?

mato:
On Monday, 28.11.2016 at 08:53, Hannes Mehnert wrote:
> Could you try to, instead of using a _tags file, pass -tag-line
> "<_build-ukvm>: -traverse" to ocamlbuild directly?

Nope, nothing :-( Checked and re-checked, only putting the line directly in
`_tags` works for me.

> > Also, how hard would it be for `mirage build` (and indeed any other commands) take the default value of `-t` from that passed to a previous run of `mirage configure`? Right now `-t` always defaults to `unix`.
> 
> That's why I still generate a Makefile.  The build step now needs a
> target.  To solve this, some persistent data (the target of the last
> configure run) needs to be stored somewhere and read back in.  Didn't
> want to go through that hazzle right now.  But it might be worth doing.
> Maybe open a separate issue for this?

Will do -- there are other uses for consuming "what target was/is this
unikernel configured for?"

hannesm:
On 28/11/2016 17:24, Martin Lucina wrote:
> On Monday, 28.11.2016 at 08:53, Hannes Mehnert wrote:
>> Could you try to, instead of using a _tags file, pass -tag-line
>> "<_build-ukvm>: -traverse" to ocamlbuild directly?
> 
> Nope, nothing :-( Checked and re-checked, only putting the line directly in
> `_tags` works for me.

that's sad.  I asked on OCaml IRC, and that `-tag-line "<_build-ukvm>:
-traverse"` is expected to have the same behaviour as putting it into a
`_tags` file.  Is there any reason ukvm does put its built products into
`_build-ukvm` instead of using e.g. `_build/ukvm`?

> Will do -- there are other uses for consuming "what target was/is this
> unikernel configured for?"

`clean` - atm for deleting the right opam file (but could only
conditionally remove Makefiles/build products (see
mir-X.virtio/mir-X.xen/mir-X.ukvm/mir-X/mir-X.elf).

mato:
On Monday, 28.11.2016 at 10:41, Hannes Mehnert wrote:
> that's sad.  I asked on OCaml IRC, and that `-tag-line "<_build-ukvm>:
> -traverse"` is expected to have the same behaviour as putting it into a
> `_tags` file.  Is there any reason ukvm does put its built products into
> `_build-ukvm` instead of using e.g. `_build/ukvm`?

Well, we needed a directory that was unlikely to conflict with anything
else in the downstream build and `_build-ukvm` was inspired by the Mirage
`_build`. We could add an option to specify the "build" directory to
`ukvm-configure`, but need to check that it would actually help first.

Having said that, if `-tag-line` is intended to work but doesn't then it's
probably a bug in `ocamlbuild`?

hannesm:
@mato the ocamlbuild problem needs further investigation and should be reported upstream.

what I still don't understand is why we don't encounter this issue with mirage&functoria master (or do we?).

what the actual question for me is how to work around the ocamlbuild bug (since I wouldn't want to wait for an ocamlbuild bugfix and release).  the options I can see so far are:
- emit a `_tags` file -- which would to be generated by functoria (since `mirage configure` leads to functoria calling ocamlbuild on config.ml)
- switch for ukvm to a different build directory below `_build`

I prefer the second option.  I don't think the complexity to set the output directory explicitly via command-line argument is needed.

hannesm:
@mato I went ahead and opened https://github.com/Solo5/solo5/pull/137 which puts build products into `_build/ukvm`.  if this is fine for you, it'd make me happy (and this PR mergeable)

hannesm:
(travis is happy according to https://travis-ci.org/hannesm/mirage-dev/builds/179629868 with the most recent changes here and in solo5)

yomimono:
I want to take this for a test drive before I click any green buttons, but feel free to click for yourself if you'd like.  This set of PRs is great! :D

hannesm:
I'm waiting for @mato approval for the `_build-ukvm` to `_build/ukvm` change

mato:
@hannesm I'm not 100% happy with using `_build/ukvm` -- thinking of other users of Solo5 and ukvm especially, `_build` could be used by other build systems which is why I suggested adding an option to specify the build directory.

Regarding the ocamlbuild problem, why don't you want to emit a `_tags` file?

hannesm:
@mato an option to `ukvm-configure` is fine as well.

Emitting a `_tags` file would need to be done in `functoria` (since this is the first caller to `ocamlbuild`).  I dislike the idea of hardcoding something special about solo5-ukvm in functoria.  Users should be able to tweak the builds by writing their own `_tags` files if needed, and I rather dislike how e.g. `oasis` handles this.

I still don't know whether the behaviour in this branch is different from master (and if so, what is the difference in master?)

mato:
@hannesm I think I have a working solution without the need for `_tags` or a change to the ukvm build directory. Invoking ocamlbuild with `-X _build-ukvm` makes it correctly ignore the directory and everything in it. Where do I need to add it to try this with your changes? (My OCaml is not up to the task...)

As for why it's not causing a problem on master, I had a look, no idea ...

hannesm:
updated those

mato:
Seems to work, hooray!

mato:
@hannesm One more `ukvm`-specific bug which got lost somewhere in this change: `ukvm-configure` needs to be called with `net` and/or `blk`, depending on whether or not the unikernel intends to use a network/block device.

mato:
i.e. this code got lost somewhere:
https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L1687
https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L1697

PR #706, opened 2016-11-29
use a result type for arp query

yomimono:
Requires changes in `tcpip`, `arp`, and any other implementors of `V1.ARP`.


yomimono:
Closing this for the moment, as on second thought I'm not quite happy with the `tcpip` side.

Issue #506, opened 2016-03-07
IP.write can fail


talex5:
According to https://github.com/mirage/mirage-tcpip/pull/166:

> Calls to `Ipv4.write` and `Ipv4.writev` can fail if a link-level address for the destination IP can't be found via ARP.

This should be either:
- Documented, and the exception exposed in the module type, or
- Turned into an error return

talex5:
(It should almost certainly be an error return, because code calling this will often want to handle network errors e.g. by retrying. I only looked quickly, but I didn't see this being handed in the TCP code; transmitting a TCP segment and getting this error will likely call `async_exception_hook` and exit the unikernel, which is unlikely to be the desired result. Unfortunately, this does mean another API change.)

hannesm:
isn't it `connect` which can fail, but `write` which shouldn't (at least in POSIX).

I'm still undecided and curious what `write` semantics people want, I'd prefer in several scenarios to have an API which allow me to check whether the segment has been acknowledged (certainly in a non-blocking fashion, otherwise bandwidth is poor).

But, in general, a `FLOW.write`(`v`) is a fire and forget operation... (and thus, the TCP/IP implementation should ensure that `write` does not depend on ARP [which should IMHO be cached for a connection])...

talex5:
Note that `IP` isn't a `FLOW`.

I don't think you can cache the next-hop ARP address, because it could easily change during a long-running TCP stream. TCP doesn't have to send every packet in a stream along the same path.

Checking the progress of a segment may be useful, but that's at a different layer (TCP) to this issue (IP).

I think that if the IP code is able to report an error then the API should allow it to do that. If TCP wants to ignore transmission errors and rely on the retransmission logic instead, it can.

hannesm:
oh, sorry for my layer violation. maybe we should talk about the API in more detail in Morocco?

talex5:
Sure :-) Actually, I'd like to discuss lots of things about error handling in Mirage there too...

yomimono:
`IP.write` now returns an error variant, but the specific case mentioned here isn't yet addressed; routing failures still generate an exception rather than a proper error.

yomimono:
Fix merged!

Issue #601, opened 2016-09-30
uncaught exceptions in the networking stack


rixed:
I have a (XEN) program that dies (uncaught exception, exit 2) when the peer it
is connected to disappear.  It will first retransmit some TCP segments (`TCP
retransmission on timer seq ...`) and then crash:

```
Fatal error: exception Ipv4.Make(Ethif)(Arpv4).Routing.No_route_to_destination_address(_)
Mirage exiting with status 2
```

The exception is thrown from some asynchronous thread spawned by
mirageos itself. Indeed, it seems the networking stack rely a lot on the
Lwt.async call without always trying to catch the exceptions.
I understand that the TCP stack will encounter errors asynchronously, but I
believe the unix model of error reporting is more convenient: the error should be saved in the "socket" internal state and be reported (synchronously) next time the socket is used for
anything.

Currently to avoid the termination of the program one can set the Lwt `async_exception_hook` but it is passed only the faulty IP; the channels depending on this IP still have to be closed.

(Note: I would have written to mirageos-devel@lists.xenproject.org instead of opening a bug but the ML still drops my emails it seems)

hannesm:
see also https://github.com/mirage/mirage-tcpip/issues/240 ... and https://github.com/mirage/mirage/issues/506 .. I don't think someone is actively solving this atm

yomimono:
This is fixed by the just-merged #711 and companion commit in `tcpip` at https://github.com/mirage/mirage-tcpip/pull/269 .

PR #711, opened 2016-11-30
Arp result types

yomimono:
Include an ARP error variant; it can be returned as an error result from `query` in module type `ARP`.  (For reals this time.)

Requires changes to `tcpip` which you can see at https://github.com/mirage/mirage-tcpip/pull/269 .

To see passing tests, check out the results for the `mirage-dev` universe including this branch and the matching `tcpip` one: https://travis-ci.org/yomimono/mirage-dev/builds/179917304


yomimono:
Fixes #506 , #601.

hannesm:
this looks fine to me

PR #714, opened 2016-11-30
docs: add more libraries from recent releases and fixes

avsm:


Dockerfile.doc
```diff
@@ -26,14 +28,17 @@ RUN opam depext -uivj 3 \
...
   dns \
+  dns-forward \
   dockerfile \
   duration \
   ezxmlm \
   fat-filesystem \
   functoria \
   git \
   git-mirage \
+  depyt \
```
yomimono: duplicate; depyt is already above

avsm: fixed in 51cd889dca18e000c69699c871533103ab303294 thanks!

PR #715, opened 2016-11-30
replace tap0 with default_network

### LINKED ISSUE #644: default network interface should be named something more generic
The current "default" network interface is named tap0, which is pretty cryptic.  Something more generic like `net0` or `default_interface` might be better.

### LINKED ISSUE #645: should the type for interfaces be a numerical index, not a string?
See discussion at https://github.com/mirage/mirage/pull/643 .  Currently, the value is a string which some implementations then attempt to interpret as a number, which can be a runtime exception (see https://github.com/mirage/mirage-net-xen/blob/master/lib/frontend.ml#L238 ).

yomimono:
Fixes #644 .

Does not fix #645.

Needs changes in unikernels that previously used `tap0`; there's a [matching PR for mirage-skeleton](https://github.com/mirage/mirage-skeleton/pull/203).  To see a consistent universe, check [this fork of mirage-dev](https://github.com/yomimono/mirage-dev/tree/default-network).  Its passing tests are visible [here](https://travis-ci.org/yomimono/mirage-dev/builds/180195689).

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -562,7 +562,11 @@ let network_conf (intf : string Key.key) =
   end
 
 let netif ?group dev = impl (network_conf @@ Key.interface ?group dev)
-let tap0 = netif "tap0"
+let default_network =
+  match_impl Key.(value target) [
+    `Xen   , netif "0";
+    `Qubes , netif "0";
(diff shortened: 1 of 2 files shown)


hannesm:
this looks fine to me, thanks!

PR #718, opened 2016-12-02
Install depexts for unikernel during make depend

### LINKED ISSUE #596: mirage configure should not install depext by default

mato:
Fixes #596.

@avsm I've deliberately not used `--update` in the call to `depext`; I don't think we did that before and it can modify the user's package manager state which IMO is not good for default behaviour.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1517,6 +1517,7 @@ let configure_makefile ~opam_name =
                   \n\
                   depend::\n\
                   \t$(OPAM) pin add --no-action --yes %s .\n\
+                  \t$(OPAM) depext --yes %s\n\
                   \t$(OPAM) install --yes --deps-only %s\n\
                   \t$(OPAM) pin remove --no-action %s\n\
                   \n\
@@ -1525,7 +1526,7 @@ let configure_makefile ~opam_name =
                   \n\
                   clean::\n\
                   \tmirage clean\n"
-        opam_name opam_name opam_name;
+        opam_name opam_name opam_name opam_name;
       newline fmt;
       append fmt "-include Makefile.user";
       R.ok ())


hannesm:
LGTM if this solves the problem (I never used depext, has been in a slightly broken state for me). thanks!

avsm:
@mato: The `--update` call is primarily for CI use, since it ensures that security updates rotate in (or in Debians case, broken archives are not attempted).  It's fine to remove it in this case as the consumer of the Makefile can ensure that their package manager is in a working state.

@hannesm is there something I can do to improve the state of depext on FreeBSD upstream in opam-depext?  Or are you referring to incomplete metadata in opam-repository?

hannesm:
@avsm it didn't work multiple times, and then I lost motivation to look into it (instead used fake installs of depext and conf-*)... will re-evaluate/investigate once I have my FreeBSD CI in a working state

mato:
@hannesm @avsm `depext` worked fine for me on FreeBSD when I last tried it...

hannesm:
@mato great to hear :)  then maybe it just works now

mato:
I don't understand what's going on with the Travis failures here...

yomimono:
I also have travis failures I don't understand on https://travis-ci.org/yomimono/mirage-dev/jobs/180729709 - I'm investigating.

yomimono:
https://github.com/mirage/mirage-skeleton/pull/205 should've fixed the travis errors here, rerunning...

yomimono:
This improves the state of the world greatly.  Thanks @mato!

PR #719, opened 2016-12-02
Make "0" the default network interface on non-Unix targets

mato:
Followup to #715. Solo5 does not currently care about the string passed
to Netif, so just use "0" for all non-Unix targets.


yomimono:
Thanks for this, @mato!

PR #721, opened 2016-12-05
move fat generation to build phase

yomimono:
If this is done in `configure`, we haven't yet had a chance to generate the list of dependencies and install them.  The user (especially if they're a CI system) might not have had a chance to install `fat-filesystem` yet.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -466,7 +466,8 @@ let fat_block ?(dir=".") ?(regexp="*") () =
   impl @@ object
     inherit block_conf block_file as super
 
-    method configure i =
+    method packages = Key.pure [ package ~build:true "fat-filesystem" ]
+    method build i =
       let root = Info.root i in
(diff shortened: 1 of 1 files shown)


hannesm:
LGTM... might conflict with #722 in a trivial way

lib/mirage.ml
```diff
@@ -466,7 +466,8 @@ let fat_block ?(dir=".") ?(regexp="*") () =
   impl @@ object
     inherit block_conf block_file as super
 
-    method configure i =
+    method packages = Key.pure [ package "fat-filesystem" ]
```
yomimono: in this `fat_block` stanza buildtime, to make sure we have `fat` to invoke for making the filesystem.  (there's also `fat_conf` which is the section for building the fs for use by the running unikernel, which has its own dependency)

hannesm: when it is only required during build time, you can use the `~build:true` argument to `package` (see example usage in crunch). if also needed at runtime, do as above.

PR #723, opened 2016-12-07
ocamlify names used in fat block construction & reference

yomimono:
Attempting to use some of the filenames generated by slapping `dir` and `regexp` together isn't going to go too well.

Also, `fpath`'s `/` threw an exception when `dir` was `../other-subdir` in the previous implementation.  `Fpath.append` on the result of `Fpath.v dir` works fine.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -518,6 +518,9 @@ let kv_ro_of_fs_conf = impl @@ object
     method name = "kv_ro_of_fs"
     method module_name = "Fat.KV_RO.Make"
     method packages = Key.pure [ package "fat-filesystem" ]
+    method connect _ modname = function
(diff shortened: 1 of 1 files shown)


dbuenzli:
`Fpath.(/)` is for adding segments you can use `Fpath.(//)` for appending paths.

yomimono:
In this case it could be any random user-generated garbage.

dbuenzli:
Then you should use `Fpath.of_string` not `Fpath.v`. See the [doc string](http://erratique.ch/software/fpath/doc/Fpath#VALv)

yomimono:
Thanks @dbuenzli .  `mirage` and `functoria` both generally need better output when the user calls them with something unexpected.

PR #725, opened 2016-12-09
remove "mir-" prefix from build products.

hannesm:
some systems have limited space for naming, and this "mir-" looks strange to me

see e.g. https://github.com/Solo5/solo5/issues/141 -- but if there are any good arguments to keep a `mir-` prefix, please discuss.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -320,7 +320,7 @@ let crunch dirname = impl @@ object
     method deps = [ abstract default_io_page ]
     method connect _ modname _ = Fmt.strf "%s.connect ()" modname
 
-    method build i =
+    method build _i =
       match query_ocamlfind  ["crunch"] with
       | Ok _ ->
         let dir = Fpath.(v dirname) in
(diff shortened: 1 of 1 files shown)


yomimono:
I see no reason to keep it, but notifying @avsm and @mor1 is a good idea since I think these names are hard-coded into the deployment scripts for our self-hosted Xen unikernels, so we'll need to change as well.

mor1:
Yes-- at a quick glance, would need to update:
https://github.com/mirage/mirage-decks-deployment/blob/master/scripts/post-merge.hook#L47
https://github.com/mirage/mirage-decks-deployment/blob/master/scripts/create-vm.sh#L29
https://github.com/mirage/mirage-decks-deployment/blob/master/scripts/prepare-config.sh#L34

...and similarly for mirage/mirage-www-deployment and mirage/mirage-tutorial-deployment

hannesm:
From my track record of breaking changes, I suspect it is not surprising that I'm happy with breaking deploy scripts... all the config.ml need to be adapted anyway ;) (here's a mir btw):

yomimono:
A+ mir 11/10 would space again

PR #726, opened 2016-12-09
Udp random

hannesm:


hannesm:
see https://github.com/mirage/mirage-tcpip/pull/272

yomimono:
I'd prefer a solution that tries not to conflict with any listeners, but the patch in mirage/mirage-tcpip#272 is better than failing outright.  The previous behavior has the annoying property of working fine for sockets and then generating a runtime failure in the direct stack.

Issue #698, opened 2016-11-22
error handling


hannesm:
as post-remarked on #690

The downside of this is that, as @talex5 remarked in https://github.com/mirage/mirage-console/pull/50#issuecomment-260770648, the errors are now closed.

Especially for `FLOW` this is a bit sad, since TLS has two kinds of failures: recoverable (such as `please reconfigure your required TLS version, or add some ciphersuite the other host understands`) and irrecoverable (no shared implemented ciphersuite).  When writing client code, I might want to handle the first by reconfiguring the stack and re-try to establish a connection, whereas the second is a hard failure.

Now, `FLOW` dictates which errors may happen: `` `Msg of string ``.  This needs some more consideration and a followup PR before 3.0 IMHO (I wouldn't count on myself having the time to work on that, unfortunately).

Having said this, I'd guess it depends very much on the specific device where we need flexible (user-extensible) errors and where not. My assumption is that there is no need for `NETIF` or `CONSOLE` to be flexible (apart from if we want to have uniformity across everything), whereas there are incentives for `FLOW` being flexible.

yomimono:
I was thinking about this particular case and it seems there's generally a useful distinction between transient errors (or recoverable ones) and errors where nothing the caller can do (including wait) has a chance of making the call succeed.

I'd love to see followup PRs on the current errors work before 3.0.

yomimono:
I've exacerbated this with #705 .

lpw25:
I happened to arrive at this from skim reading mirageOS-devel. I have a suggestion that might help with what @talex5 mentioned in the linked comment. You could do something like:

```
(* In mirage-types -- or whatever your central place is *)
type flow_error = [
  `Some_error
  `Some other_error
]

val pp_flow_error : fmt -> flow_error -> ()

module type Flow = sig
  type error

yomimono:
Belated thanks, @lpw25 , this was exactly the kind of "um just do this" advice I needed to face ever looking at this again.  Patches are coming but as always, I won't be heartbroken in the case that someone else gets to it first.

hannesm:
I spend some hours in order to get this working, but failed so far.

What I discovered on the way:
- `Kv_ro.error` now contains `` `Unknown_key`` (previously: `` `Unknown_key of string `` -- what was the reason to drop the string (IMHO it was useful for error messages)
[...]

PR #729, opened 2016-12-12
[WIP; RFC] error types restated privately in flow implementations

yomimono:
A universe is in progress at https://github.com/yomimono/mirage-dev/tree/lpw-error-suggs ; it currently contains `mirage`, `tcpip`, and `mirage-flow`.

Attempts to address some concerns in #698.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -165,7 +165,19 @@ module type FLOW = sig
   (** The type for flows. A flow represents the state of a single
       reliable stream that is connected to an endpoint. *)
 
-  val read: flow -> (buffer Flow.or_eof, Flow.error) result io
+  type error = private [> Flow.error]
+  (** the type for read errors. *)
+
+  type write_error = private [> Flow.write_error]
+  (** the type for write errors. *)
+
+  val pp_error : Format.formatter -> error -> unit
+  (** a printer for errors. *)
+
+  val pp_write_error : Format.formatter -> write_error -> unit
+  (** a printer for write errors. *)
+
+  val read: flow -> (buffer Flow.or_eof, error) result io
   (** [read flow] blocks until some data is available and returns a
       fresh buffer containing it.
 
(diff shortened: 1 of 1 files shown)


yomimono:
We have a module (`Tls_mirage.Make`)which wants take a FLOW as a module argument and then reexport the functions of that FLOW, but with its own error types.  This design doesn't really work for that case, because it's not possible for us to include the module argument's error types directly (like this):

```
module Make (F : V1_LWT.FLOW) : sig
  type error  = [ `Tls_alert   of Tls.Packet.alert_type
                | `Tls_failure of Tls.Engine.failure
                | F.error
                ]
```

because `F.error` is private:

```
File "mirage/tls_mirage.mli", line 13, characters 18-25:
Error: The type F.error is not a polymorphic variant type
```

And we also can't wrap the error like this:

```
  type error  = [ `Tls_alert   of Tls.Packet.alert_type
                | `Tls_failure of Tls.Engine.failure
                | `Flow of F.error
                ]
```

because then our error type is no longer `> V1.Flow.error`, as the polyvars in V1.Flow.error are wrapped.

We had some discussion around this in the OCamlLabs Slack around this pattern.  I'm not convinced that it can accommodate the desired use in ocaml-tls.

talex5:
Maybe the problem is that `Msg of string` shouldn't be in `Flow.error` in the first place?

This forces users of a flow to handle the cases of both `Msg` and private, unknown errors. We should be able to remove all `Msg` or `Unknown` cases I think. They don't add anything to the interfaces.

For a meaningful case (e.g. `Eof`) you would want it in the TLS type, and you'd want TLS to generate TLS `Eof` errors from its underlying flow in some way. It is slightly annoying that the type would then allow both `Eof` and `Flow Eof`, even in the case where all eof errors in the underlying flow should turn into TLS eofs, but maybe that wouldn't be true in the general case (e.g. an Eof half way through a TLS record might not be a clean Eof at the TLS layer).

hannesm:
If it would just be a polyvar, not private, would it work?  Is there any value in making it private?

In the end, as a client of the `Tls_mirage` FLOW, I might want to act upon some concrete `` `Tls_failure `` (by reconfiguring my stack, and retry a connection) -- this means boiling it down into a single `` `Msg of string `` won't solve this problem.

yomimono:
Closing; https://github.com/mirage/mirage/pull/737 exists and I don't expect to push here again.

PR #730, opened 2016-12-13
mirage-runtime depends on mirage-types

samoht:
I guess it's added by the mirage tool at some point, but it's better to keep track of it there.


hannesm:
thx, the upstream `mirage-dev` repository needs an update as well, then...

hannesm:
(ehm, no, it is already in the opam files)

PR #733, opened 2016-12-13
FS: remove `format` and `Format_unknown`

djs55:
`format` is similar to `connect` which we removed previously from the common interface: it is likely to be implementation-specific (e.g. number of inodes, number of superblocks etc etc)

Furthermore it seems strange to have to acquire a `FS.t` handle in order to be able to format a filesystem. This requires that implementations track whether the `FS.t` is formatted or not (e.g. with a Some or None) and then all operations may return `Format_unknown` (which they currently can't). Removing `format` and `Format_error` fixes this problem.

With this change, `Fs.error` is a proper subset of `Fs.write_error`.

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -883,7 +883,6 @@ module Fs : sig
     | `Is_a_directory      (** Cannot read or write the contents of a directory *)
     | `No_directory_entry  (** Cannot find a directory entry *)
     | `Not_a_directory     (** Cannot create a directory entry in a file *)
-    | `Format_unknown      (** The block device appears to not be formatted *)
   ]
   type write_error = [
     | `Msg of string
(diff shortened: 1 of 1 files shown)


djs55:
I'm still fixing up `ocaml-fat` as a Mirage 3 FS proof of concept. I'll send a pointer when I've got something.

djs55:
For reference the `ocaml-fat` main signature now looks like:
```
module Make (B: V1_LWT.BLOCK)(M: S.IO_PAGE) : sig
  include V1_LWT.FS

  val connect : B.t -> (t, V1.Block.error) Result.result io
  val format : B.t -> int64 -> (t, [> `Msg of string]) Result.result io
end
```
- Note how both `connect` and `format` take a block device (rather than a filesystem)

yomimono:
Thanks for this patch and the ocaml-fat one, @djs55 .  This change seems sensible to me.

djs55:
@yomimono thanks! I'll merge this and the mirage-dev PR.

PR #737, opened 2016-12-15
WIP: No more error

hannesm:
on top of #729,  also https://github.com/mirleft/ocaml-tls/pull/348 https://github.com/mirage/mirage-console/pull/53 https://github.com/mirage/mirage-console-solo5/pull/9 https://github.com/mirage/mirage-tcpip/pull/275 https://github.com/mirage/mirage-flow/pull/24

still not very happy: the compilation of `tls` succeeds, but I'm unable to use it:
```OCaml
File "src/logs_syslog_mirage_tls.ml", line 6, characters 31-34:
Error: Signature mismatch:
       ...
       The type `error' is required but not provided
       File "types/V1.mli", line 217, characters 7-42: Expected declaration
```

where line 6 is: `module TLS = Tls_mirage.Make(TCP)`

And lots of `lift_XX` in `tls_mirage.ml`, which IMHO shouldn't be necessary.  I suspect the fundamental problem is that `error` and `write_error` in `FLOW` are unrelated (and furthermore, the `` `Msg of string `` should not be in `FLOW`)!?!  In TLS, since a `read` can lead to `write`, and `write` can lead to `read`, the distinction of reader and writer errors is rather brittle.

Maybe @yallop knows a quick way out or someone else (@talex5 @samoht @avsm) has some time to get rid of `` `Msg ``!?

client code: https://github.com/hannesm/logs-syslog/tree/no-more-error

### DIFF
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -141,7 +141,7 @@ module Flow : sig
 end
 
 (** {1 Connection between endpoints} *)
-module type FLOW = sig
+module type RFLOW = sig
 
   type +'a io
   (** The type for potentially blocking I/O operations. *)
@@ -153,10 +153,10 @@ module type FLOW = sig
   (** The type for flows. A flow represents the state of a single
       reliable stream that is connected to an endpoint. *)
 
-  type error = private [> Flow.error]
+  type error
   (** the type for read errors. *)
 
-  type write_error = private [> Flow.write_error]
+  type write_error
   (** the type for write errors. *)
 
   val pp_error : Format.formatter -> error -> unit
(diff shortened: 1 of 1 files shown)


types/V1.mli
```diff
@@ -141,7 +141,7 @@ module Flow : sig
 end
 
 (** {1 Connection between endpoints} *)
-module type FLOW = sig
+module type RFLOW = sig
```
hannesm: according to @yallop there's no need to have `RFLOW`, but instead magic runes in implementations of `FLOW` can be used...

samoht:
Yes we should get rid of the `'Msg of string` as we discussed this morning with @avsm. This is not needed as we can just use `pp_error`. I'll try to have a look at that tomorrow.

types/V1.mli
samoht: which magic runes?

PR #740, opened 2016-12-18
mainly cosmetics

hannesm:
- remove all raise from `method package`
- in case a device is not suitable for the target, `configure` emits an error
- don't Log.err _and_ R.error_msg, only use the latter
- unify connect error message
- use Fmt.strf all over instead of partially Printf.a/sprintf
- crunch has a build-dep on crunch, no need to verify manually that crunch is
  around during build time
- when matching target, don't use wildcards
- massage code to 80 columns

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -22,7 +22,7 @@ module Key = Mirage_key
 module Name = Functoria_app.Name
 module Codegen = Functoria_app.Codegen
 
-let src = Logs.Src.create "mirage" ~doc:"mirage library"
+let src = Logs.Src.create "mirage" ~doc:"mirage cli tool"
 module Log = (val Logs.src_log src : Logs.LOG)
 
 include Functoria
@@ -34,6 +34,9 @@ let with_output ?mode f k err =
   | Ok b -> b
   | Error _ -> R.error_msg ("couldn't open output channel for " ^ err)
 
+let connect_err name number =
+  Fmt.strf "The %s connect expects exactly %d argument%s"
+    name number (if number = 1 then "" else "s")
 
 (** {2 OCamlfind predicates} *)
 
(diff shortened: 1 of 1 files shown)


lib/mirage.ml
```diff
@@ -1064,11 +1029,14 @@ let resolver_unix_system = impl @@ object
...
-      | `Unix | `MacOSX ->
+      Key.(if_ is_unix)
         [ package ~ocamlfind:[] "mirage-conduit" ;
           package ~sublibs:["mirage";"lwt-unix"] "conduit" ]
-      | _ -> failwith "Resolver_unix not supported on unikernel"
+        []
+    method configure i =
+      match get_target i with
+      | `Unix | `MacOSX -> R.ok ()
+      | _ -> R.error_msg "Resolver_unix not supported on unikernel"
```
avsm: 'on unikernel' -> 'on non-Unix targets'?

hannesm:
@avsm thx for your comment, I did another commit which takes that into account

lib/mirage.ml
```diff
@@ -1237,42 +1202,37 @@ let mprof_trace ~size () =
...
-             MProf.Trace.Control.start trace_config;@ \
-             MProf_xen.share_with (module Gnt.Gntshr) (module OS.Xs) ~domid:0 trace_pages@ \
-             |> OS.Main.run@]"
+           @[<v 2> let trace_pages = MProf_xen.make_shared_buffer ~size:%a in@ \
+           let buffer = trace_pages |> Io_page.to_cstruct |> Cstruct.to_bigarray in@ \
+           let trace_config = MProf.Trace.Control.make buffer MProf_xen.timestamper in@ \
+           MProf.Trace.Control.start trace_config;@ \
+           MProf_xen.share_with (module Gnt.Gntshr) (module OS.Xs) ~domid:0 trace_pages@ \
+           |> OS.Main.run@]"
           Key.serialize_call (Key.abstract key)
```
talex5: Yes, under Xen we need to grant dom0 permission to read the trace buffer. On Unix, we mmap a file. I think Travis tests that tracing works (or at least compiles).

hannesm: ok, thanks for the clarification.

Issue #349, opened 2015-01-23
remove need to pass "xvda1" to block_of_file for xen backends


heidi-ann:
issue discussed at https://github.com/mirage/mirage-skeleton/pull/69#issuecomment-71226567 and https://github.com/mirage/xen-arm-builder/issues/47

talex5:
It's not clear to me what you could do about this. The Unix file and the Xen block device genuinely have different names, so you need to give both.

The only options I can think of are:

1. Name the Unix file `xvda1` so it just happens to match. This would work for the examples (except that there's currently a bug in `mirage` preventing it - if the file exists, it converts to an absolute path, even if it's being used as a Xen device name).

2. Make a convenience function that handles both. e.g. `block ~xen:"xvda1" ~unix:"disk.img" ()`.

samoht:
I think that has been fixed by @djs55

Issue #273, opened 2014-07-21
Add a combinator for list of states


samoht:
> How a list of network devices with arbitrary length, for example [tap0;
tap1; tap2], that is defined in config, can be passed as a parameter to
the Main module in unikernel.ml?

Drup:
That should be solved too, by creating a `list string` key.

Issue #182, opened 2014-01-05
cross-compilation to ARM/Android


avsm:
Some notes from @cgreenhalgh about cross-compiling Mirage to Android:

http://lists.ocaml.org/pipermail/opam-devel/2013-September/000198.html

All of the notes are found here:
https://github.com/cgreenhalgh/ocaml-crosscompiling

@dsheets, could you add anything more you have to this issue on your OPAM cross compilation experiences as well?

avsm:
This will be dealt with in Mirage 4 via https://github.com/mirage/mirage/issues/1195

Issue #181, opened 2014-01-03
fix support for multiple xen block devices


avsm:
The current heuristic is to select the first one, which breaks multiple block devices.  We need this for mirage-decks, as a full crunch results in a 40MB kernel.

samoht:
I think this is now fixed.

PR #743, opened 2016-12-20
Refactors errors and split signatures into their own opam packages

samoht:
Following the discussions we had with @talex5, @yallop and @hannesm last week, I implemented the new error scheme using private rows. After finding [2](https://caml.inria.fr/mantis/view.php?id=7437) [bugs](https://caml.inria.fr/mantis/view.php?id=7438) in the compiler, I finally managed to get something working. Here is the full set of patches:

### new repos with new signatures
- mirage-device (DEVICE): [GitHub - mirage/mirage-device: Mirage device](https://github.com/mirage/mirage-device)
- mirage-time (TIME):  [GitHub - mirage/mirage-time: Time signatures for MirageOS](https://github.com/mirage/mirage-time)
- mirage-random (RANDOM) : [GitHub - mirage/mirage-random: Random-related devices for MirageOS](https://github.com/mirage/mirage-random)

### move V1 signatures in existing repo:
- mirage-console (CONSOLE) + errors: [Update the new error scheme in mirage-types and import CONSOLE sigs from V1 by samoht · Pull Request #55 · mirage/mirage-console · GitHub](https://github.com/mirage/mirage-console/pull/55)
- mirage-block (BLOCK) + errors:  [Modernize build, import V1.BLOCK from mirage-types by samoht · Pull Request #25 · mirage/mirage-block · GitHub](https://github.com/mirage/mirage-block/pull/25)
- mirage-flow (FLOW) + errors: [modernize the build, split into 3 packages + adapt to the new error scheme by samoht · Pull Request #25 · mirage/mirage-flow · GitHub](https://github.com/mirage/mirage-flow/pull/25)
- mirage-clock (MCLOCK, PCLOCK): [Import V1.MCLOCK and V1.PCLOCK from mirage-types by samoht · Pull Request #24 · mirage/mirage-clock · GitHub](https://github.com/mirage/mirage-clock/pull/24)

### Use the new split sigs only
- mirage-logs: [Use the new mirage-clock instead of mirage-types by samoht · Pull Request #8 · mirage/mirage-logs · GitHub](https://github.com/mirage/mirage-logs/pull/8)

### Adapt to the new error scheme + use the new split sigs	
- mirage-vnetif: [Use topkg, use the new mirage-time and mirage-clock module, adapt to the new error scheme by samoht · Pull Request #11 · MagnusS/mirage-vnetif · GitHub](https://github.com/MagnusS/mirage-vnetif/pull/11)
- mirage-net-unix: [Adapt to the new error scheme in mirage-types by samoht · Pull Request #35 · mirage/mirage-net-unix · GitHub](https://github.com/mirage/mirage-net-unix/pull/35)
- mirage-net-macosx: [Adapt to the new error scheme by samoht · Pull Request #16 · mirage/mirage-net-macosx · GitHub](https://github.com/mirage/mirage-net-macosx/pull/16)
- mirage-channel: [Adapt to the new error scheme by samoht · Pull Request #17 · mirage/mirage-channel · GitHub](https://github.com/mirage/mirage-channel/pull/17)
- ocaml-dns: [Adapt to the new error scheme by samoht · Pull Request #116 · mirage/ocaml-dns · GitHub](https://github.com/mirage/ocaml-dns/pull/116)
- ocaml-tls: [mirage: update the the new error scheme by samoht · Pull Request #349 · mirleft/ocaml-tls · GitHub](https://github.com/mirleft/ocaml-tls/pull/349)
- ocaml-conduit: [Adapt to the new error scheme + depends on mirage-flow instead of mirage-types by samoht · Pull Request #182 · mirage/ocaml-conduit · GitHub](https://github.com/mirage/ocaml-conduit/pull/182)
- mirage-http: [Use the new error scheme by samoht · Pull Request #32 · mirage/mirage-http · GitHub](https://github.com/mirage/mirage-http/pull/32)
- ocaml-crunch: [Adapt to the new error scheme by samoht · Pull Request #31 · mirage/ocaml-crunch · GitHub](https://github.com/mirage/ocaml-crunch/pull/31)
- mirage-console-solo5: [Adapt to the new error scheme by samoht · Pull Request #10 · mirage/mirage-console-solo5 · GitHub](https://github.com/mirage/mirage-console-solo5/pull/10)
- mirage-net-solo5: [Adapt to the new error scheme by samoht · Pull Request #15 · mirage/mirage-net-solo5 · GitHub](https://github.com/mirage/mirage-net-solo5/pull/15)
- mirage-net-xen: [Adapt to the new error scheme by samoht · Pull Request #54 · mirage/mirage-net-xen · GitHub](https://github.com/mirage/mirage-net-xen/pull/54)
- ocaml-vchan: [Adapt to the new error scheme, use mirage-flow and mirage-console instead of mirage-types by samoht · Pull Request #97 · mirage/ocaml-vchan · GitHub](https://github.com/mirage/ocaml-vchan/pull/97)
- mirage-block-solo5: [Adapt to the new error scheme, use mirage-block instead of mirage-types by samoht · Pull Request #7 · mirage/mirage-block-solo5 · GitHub](https://github.com/mirage/mirage-block-solo5/pull/7)
- mirage-block-xen: [Adapt to the new error scheme by samoht · Pull Request #59 · mirage/mirage-block-xen · GitHub](https://github.com/mirage/mirage-block-xen/pull/59)
- mirage-block-ramdisk: [Import the tests from mirage-block, use the new Mirage_block.S by samoht · Pull Request #12 · mirage/mirage-block-ramdisk · GitHub](https://github.com/mirage/mirage-block-ramdisk/pull/12)

### Fix pre-existing errors
- charrua-client: [Link with mirage-types-lwt instead of mirage-types.lwt by samoht · Pull Request #7 · yomimono/charrua-client · GitHub](https://github.com/yomimono/charrua-client/pull/7)
- mirage-platform: [xen: really remove the dependency to mirage-types by samoht · Pull Request #181 · mirage/mirage-platform · GitHub](https://github.com/mirage/mirage-platform/pull/181)

### Meta
- mirage: [Refactors errors and split signatures into their own opam packages by samoht · Pull Request #743 · mirage/mirage · GitHub](https://github.com/mirage/mirage/pull/743)
- mirage-dev: [do not merge Adapt to the new error scheme by samoht · Pull Request #273 · mirage/mirage-dev · GitHub](https://github.com/mirage/mirage-dev/pull/273)
- mirage-skeleton: [Update to new error scheme in mirage-types by samoht · Pull Request #213 · mirage/mirage-skeleton · GitHub](https://github.com/mirage/mirage-skeleton/pull/213)

### Add new constraints

- ocaml-dns-forward: [Add a constraint for the version of mirage-flow to use by samoht · Pull Request #50 · djs55/ocaml-dns-forward · GitHub](https://github.com/djs55/ocaml-dns-forward/pull/50)

### DIFF
--- a/lib_runtime/mirage_pp.ml
+++ b/lib_runtime/mirage_pp.ml
@@ -17,6 +17,10 @@ let pp_tcp_error ppf = function
   | `Timeout -> Fmt.string ppf "connection attempt timed out"
   | `Refused -> Fmt.string ppf "connection attempt was refused"
 
+let pp_tcp_write_error ppf = function
+  | #Mirage_flow.write_error as e -> Mirage_flow.pp_write_error ppf e
+  | #V1.Tcp.error as e            -> pp_tcp_error ppf e
+
 let pp_fs_error ppf = function
     | `Is_a_directory      -> Fmt.string ppf "is a directory"
     | `Not_a_directory     -> Fmt.string ppf "is not a directory"
--- a/lib_runtime/mirage_pp.ml
+++ b/lib_runtime/mirage_pp.ml
@@ -1,79 +1,33 @@
-open Result
-let pf pp = Format.fprintf pp
+open V1
 
-let pp_console_error pp = function
-  | `Invalid_console str -> pf pp "invalid console %s" str
+let pp_network_error = Mirage_device.pp_error
+let pp_ethif_error = Mirage_device.pp_error
 
-let unspecified pp m = pf pp "unspecified error - %s" m
+let pp_arp_error ppf = function
+  | `Timeout -> Fmt.string ppf "Dynamic ARP timed out"
 
-let pp_network_error pp = function
-  | `Msg message   -> unspecified pp message
-  | `Unimplemented -> pf pp "operation not yet implemented"
-  | `Disconnected  -> pf pp "device is disconnected"
+let pp_ip_error ppf = function
+  | #Mirage_device.error as e -> Mirage_device.pp_error ppf e
+  | `No_route -> Fmt.string ppf "no route to destination"
 
-let pp_ethif_error = pp_network_error
+let pp_icmp_error ppf = function
+  | `Routing message -> Fmt.pf ppf "routing error: %s" message
 
-let pp_arp_error pp = function
-  | `Timeout -> Format.fprintf pp "Dynamic ARP timed out"
+let pp_tcp_error ppf = function
+  | `Timeout -> Fmt.string ppf "connection attempt timed out"
+  | `Refused -> Fmt.string ppf "connection attempt was refused"
 
-let pp_ip_error pp = function
-  | `Msg message   -> unspecified pp message
-  | `Unimplemented -> Format.fprintf pp "operation not yet implemented"
-  | `Disconnected  -> Format.fprintf pp "device is disconnected"
-  | `No_route      -> Format.fprintf pp "no route to destination"
+let pp_fs_error ppf = function
+    | `Is_a_directory      -> Fmt.string ppf "is a directory"
+    | `Not_a_directory     -> Fmt.string ppf "is not a directory"
+    | `No_directory_entry  ->
+      Fmt.string ppf "a directory in the path does not exist"
 
+let pp_fs_write_error ppf = function
+  | #Fs.error as e       -> pp_fs_error ppf e
+  | `Directory_not_empty -> Fmt.string ppf "directory is not empty"
+  | `File_already_exists -> Fmt.string ppf "file already exists"
+  | `No_space            -> Fmt.string ppf "device has no more free space"
 
-let pp_icmp_error pp = function
-  | `Msg message   -> unspecified pp message
-  | `Routing message -> pf pp "routing error: %s" message
-
-let pp_udp_error pp (`Msg message) = unspecified pp message
-
-let pp_tcp_error pp = function
-  | `Msg message   -> unspecified pp message
-  | `Timeout       -> pf pp "connection attempt timed out"
-  | `Refused       -> pf pp "connection attempt was refused"
-
-let pp_flow_error pp = function
-  | `Msg message   -> unspecified pp message
-
-let pp_flow_write_error pp = function
-  | `Msg message   -> unspecified pp message
-  | `Closed        -> pf pp "attempted to write to a closed flow"
-
-let pp_block_error pp = function
-  | `Msg message   -> unspecified pp message
-  | `Unimplemented -> pf pp "operation not yet implemented"
-  | `Disconnected  -> pf pp "a required device was disconnected"
-
-let pp_block_write_error pp = function
-  | `Msg message   -> unspecified pp message
-  | `Unimplemented -> pf pp "operation not yet implemented"
-  | `Disconnected  -> pf pp "a required device was disconnected"
-  | `Is_read_only  -> pf pp "attempted to write to a read-only disk"
-
-let pp_fs_error pp = function
-    | `Msg message         -> unspecified pp message
-    | `Is_a_directory      -> pf pp "is a directory"
-    | `Not_a_directory     -> pf pp "is not a directory"
-    | `No_directory_entry  -> pf pp "a directory in the path does not exist"
-    | `Format_unknown      -> pf pp "the device is not formatted for this filesystem"
-
-let pp_fs_write_error pp = function
-  | `Msg message         -> unspecified pp message
-  | `Is_a_directory      -> pf pp "is a directory"
-  | `Not_a_directory     -> pf pp "is not a directory"
-  | `Directory_not_empty -> pf pp "directory is not empty"
-  | `No_directory_entry  -> pf pp "a directory in the path does not exist"
-  | `File_already_exists -> pf pp "file already exists"
-  | `No_space            -> pf pp "device has no more free space"
-
-let pp_kv_ro_error pp = function
-  | `Msg message   -> unspecified pp message
-  | `Unknown_key   -> pf pp "key not present in the store"
-
-let reduce = function
-  | Ok () -> Ok ()
-  | Error (`Msg _) as e -> e
-  | Error `Unimplemented -> Error (`Msg "unimplemented functionality discovered")
-  | Error `Disconnected -> Error (`Msg "a required device was disconnected")
+let pp_kv_ro_error ppf = function
+  | `Unknown_key k -> Fmt.pf ppf "key %s not present in the store" k
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -442,6 +442,7 @@ end
 (** TCP errors. *)
 module Tcp : sig
   type error = [ `Timeout | `Refused]
+  type write_error = [ error | Mirage_flow.write_error]
 end
 
 (** A TCP stack that can send and receive reliable streams using the
@@ -451,7 +452,7 @@ module type TCP = sig
   type error = private [> Tcp.error]
   (** The type for TCP errors. *)
 
-  type write_error = private [> Mirage_flow.write_error | Tcp.error]
+  type write_error = private [> Tcp.write_error]
   (** The type for TCP write errors. *)
 
   type buffer
--- a/types/V1.mli
+++ b/types/V1.mli
@@ -48,271 +48,36 @@ open Result
 
     {e Release %%VERSION%% } *)
 
-(** {1 Abstract devices}
+module type DEVICE = Mirage_device.S
 
-    Defines the functions to define what is a device state and
-    how to disconnect such a device. *)
-module type DEVICE = sig
+(** {1 Time-related devices} *)
 
-  type +'a io
-  (** The type for potentially blocking I/O operation *)
-
-  type t
-  (** The type representing the internal state of the device *)
-
-  val disconnect: t -> unit io
-  (** Disconnect from the device.  While this might take some time to
-      complete, it can never result in an error. *)
-end
-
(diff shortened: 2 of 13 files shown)


types/V1.mli
```diff
@@ -225,32 +252,32 @@ module type CONSOLE = sig
...
+
+(** Block device errors. *)
 module Block : sig
-  type error = [
-    | `Msg of string     (** an unspecified error *)
-    | `Unimplemented     (** operation not yet implemented in the code *)
-    | `Disconnected      (** the device has been previously disconnected *)
-  ]
+
+  type error = Device.error
```
hannesm: wouldn't we want to be able to extend for a specific `Block` to include a `` `Msg of string ``!?

hannesm: ups, nevermind.  I should first read until the end

hannesm:
I got the same warning, also think it is harmless (maybe just put `-34` in the `_tags` file?).

This PR looks good to me!

lib_runtime/mirage_pp.ml
```diff
@@ -1,79 +1,47 @@
-open Result
+open V1
 let pf pp = Format.fprintf pp
```
avsm: We could just put a dependency on `Fmt` here, as it defines the same function.

types/V1.mli
```diff
@@ -48,10 +48,23 @@ open Result
 
     {e Release %%VERSION%% } *)
 
-(** {1 Abstract devices}
+type 'a pp = Format.formatter -> 'a -> unit
```
avsm: and likewise here, this is `Fmt.t` -- while we are trying to minimise dependencies here, Fmt is standalone and has a number of other useful functions to sensibly compose formatters.  I'm not against this patch either though, just a suggestion.

avsm:
```
# File "lib_runtime/mirage_pp.mli", line 3, characters 42-55:
# Error: Unbound module Console
# Command exited with code 2.
```

types/V1.mli
samoht: Today `mirage-types` does not have any dependency.

samoht:
@avsm all the reverse builds are broken because I haven't do the work yet :-) I'm doing that next.

lib_runtime/mirage_pp.ml
```diff
@@ -1,79 +1,47 @@
...
 
-let pp_console_error pp = function
-  | `Invalid_console str -> pf pp "invalid console %s" str
-
-let unspecified pp m = pf pp "unspecified error - %s" m
+let pp_device_error pp = function
```
samoht: Yes, good point.

types/V1.mli
hannesm: I do appreciate mirage-types not having any dependency (since we require 4.03+, result must be present in any case)

avsm: Fine by me.

avsm: One thing that would help beginners (as a documentation pass) is some tips on where to find functions that use the defined types -- for example, a pointer to Ptime for the time functions, Fmt for the pretty printers, Rresult for result combinators.  This can be be done later once the code compiles :)

lib_runtime/mirage_pp.ml
samoht: The `_pp.pp_` is indeed pretty annoying. Two options:
- I can move the function to `Mirage_runtime`
- I can just drop the `pp_` prefix, and tell people to not open `Mirage_pp`.

I tend towards the later, but both options seems ok.

hannesm: I prefer the latter as well.

samoht:
Finally after discussing with @yomimono and @avsm I am going to the full ride: split up `V1` into pieces, and move the error/pretty-printers into the sub-packages instead of having to deal centrally with all the boiler-plate.

This is a rather large change, but I suspect that it's possible to do it in a few days (and I'm very motivated to do it :p) so I started to do it (it's not strictly speaking part of this PR, but this is somehow related, so I will continue working here).

So far I have:
- created https://github.com/mirage/mirage-device and put the `DEVICE` signature there
- created https://github.com/mirage/mirage-random and put the `RANDOM` signature there
- created https://github.com/mirage/mirage-time and put the `SLEEP`, `MCLOCK` and `PCLOCK` signatures there
- refactored `mirage-flow` to use `topkg` and to define 3 subpackages (`mirage-flow`, `mirage-flow-lwt` and `mirage-flow-uniz): https://github.com/mirage/mirage-flow/pull/25 This is a rather big change, but the tests still pass so I am sure that's fine

I'll move out the other signatures packages tomorrow, and I hope to have something mergeable before the end of the week.

dbuenzli:
> created https://github.com/mirage/mirage-time and put the SLEEP

While `SLEEP` and clocks can be related (e.g. wait for a clock interrupt to continue), it seems to me that the current `SLEEP`, being tied to cooperative concurrency has nothing to do in there.

samoht:
@dbuenzli any suggestion where should I should put it? `mirage-sleep`?

dbuenzli:
> @dbuenzli any suggestion where should I should put it? mirage-sleep?

Wherever your run loop abstraction is defined. Also `mirage-time` could be renamed to `mirage-clock`

samoht:
> Wherever your run loop abstraction is defined

It's already duplicated on every platform and tests:
- https://github.com/mirage/mirage-platform/tree/master/unix/lib 
- https://github.com/mirage/mirage-platform/blob/master/xen/lib/time.mli#L50
- https://github.com/mirage/mirage-solo5/blob/master/lib/oS.mli#L75
- https://github.com/mirage/mirage-tcpip/blob/master/lib_test/vnetif_common.ml#L21
- https://github.com/mirage/ocaml-qcow/blob/master/lib_test/utils.ml#L112
- ...

Which IMHO seems to indicate that we have to either remove it (and assume an ambient `sleep` function) or have a central place to define this helper and make people use it.

talex5:
I often want my sleeps to depend on a clock, rather than on the run loop, for unit tests (e.g. [mirage-clock-test](https://github.com/talex5/mirage-clock-test)) . It seems reasonable that a clock might want to offer a way to get a notification at a certain time, or after some period of time has elapsed according to that clock.

samoht:
@MagnusS just pointed me to https://github.com/MagnusS/mirage-event-clock/blob/master/lib/event_clock.ml#L74

dbuenzli:
> I often want my sleeps to depend on a clock, rather than on the run loop, for unit tests (e.g. mirage-clock-test) .

It seems you rather want to configure which clock you use for your run loop. 

> It seems reasonable that a clock might want to offer a way to get a notification at a certain time, or after some period of time has elapsed according to that clock.

While a clock could provide you this under the form of interrupts, you wouldn't want cooperative programs to use this directly. Rather a run loop would use this with a priority queue to implement a cooperative `sleep`.

Drup:
@samoht I feel like this would be simpler if you had all the packages in the same repository (still as separate packages).

hannesm:
as discussed several times, I'd be happy when the whole [mirage-os-shim](https://github.com/pqwy/mirage-os-shim/blob/master/src/mirage_OS.mli) would just be there, no need to functorise over basic OS functionality which should be there exactly once (or is anyone seriously interested in running two clocks at different speed in the same unikernel?) -- and I still think if you use `time` or `clock` in your tests, you made a mistake (see for example the thread on arp - if you make time explicit to your pure protocol implementation (which you should), there's no need to rely on a concrete `Time.sleep_ns` in your tests).

but I'm too tired to discuss this again.

samoht:
Let's say that this won't be resolve for Mirage3 :p 

My current goal is just to split the types in separate repos, while keeping the signatures as similar as possible as what we have today in `V1`. I think this is pretty non-controversial for now :-) We can revisit that in 6 months.

avsm:
I agree with splitting up the types into separate OPAM packages, but is there any reason not to keep them in the same Git repo with multiple .opam files in the repo?  We would avoid a preponderance of repositories and make future name refactoring easier.

samoht:
A quick update: I have made some progress and I am able to compile a few examples in `mirage-skeleton` with the new scheme: I managed to port [tls](https://github.com/mirleft/ocaml-tls/compare/master...samoht:fs-errors) with relatively few changes, I thing the last blocker (hopefully) is `conduit`. The packages that I am porting are tracked in https://github.com/samoht/mirage-dev/commits/master 

My goal is to make all the examples compile in `mirage-skeleton` and then polish the patches and merge. I realise that it is a large patcher, which is blocking progress for other people, so I am trying to be quick ...

samoht:
See https://github.com/mirage/mirage-skeleton/pull/213 for the changes needed to mirage-skeleton to compile all of the example on the unix backend. Polishing up all the patches/PRs now.

One thing that maybe went wrong is that I notice that the TCP/IP stack seems a bit slow on OSX (using vmnetd).  ICMP packets seems to get lost pretty often, and the HTTP request takes `>1s` (so maybe we lost the first SYN too). Not sure where the error comes from, I will try to reproduce without my patches.

yomimono:
I'm working on the `tcpip` problem @samoht reported.  I suspect I broke this in #643 and it's only now been discovered.

samoht:
I have updated the issue body with all the current PRs. My plan is to continue fixing the last few remaining issues, and do a merge tomorrow.

The second-half of splitting the Mirage signature could (and probably should) be done separately. I plan to look at that next week.

samoht:
Rebased, with hopefully more green ticks this time.

samoht:
mirage-skeleton is green on unix, xen and solo5. Most of the individual PRs are green too. I'll start to merge everything, expect a little bit of breakage for a few hours...

PR #744, opened 2016-12-22
fixes newline in bootvar connect

yomimono:
Without this fix, `mirage configure -t qubes` generates a `main.ml` that does not compile.


hannesm:
thanks for fixing this, sorry for breaking it

Issue #694, opened 2016-11-20
workflow: provision after build


hannesm:
~~~This superseeds https://github.com/mirage/mirage/issues/631 and addresses https://github.com/mirage/mirage/pull/691#issuecomment-261734044~~~

(updated to reflect changes merged via #703)

At the very moment in master:
- `mirage configure` won't call out to opam.  It will configure the unikernel:  generate a main.ml and a Makefile.
- `make depend` will install the missing dependencies
- `make` compiles the OCaml code as before
- crunch is done during build

After XXX, we should be able to do:
- `mirage configure` - same as above
- `make depend` - install forgotten packages
- `make` - compile
- `provision` - link/add secret keys to existing unikernel image.  The build host does not need to be trusted anymore!

There are still some missing pieces (how to actually do the provisioning -- easy with block devices: just provide the device during startup, but unclear how to do crunch here).

I'd be happy to use for MirageOS3 the intermediate solution, but if there are people who have strong ideas and know how to do the rest, we can also target the full provisioning workflow.

hannesm:
now that #703 is merged, I'd leave this open until we have a plan for the provisioning step (esp regarding Crunch to be applied/linked after the unikernel is built)

hannesm:
#204 and #423 are related to this as well.. I guess we'll need a clean proposal how the MirageOS-X (for X > 3.0.0) should work (as mentioned above, I'd hope to have a `configure - build - provision` workflow -- currently the provision happens while building)

mato:
@dinosaure asked me today about implementation approaches for embedding secrets (or any arbitrary data) in a unikernel binary in a separate `provision` step. There are several options:

1. Appending extra data to the binary during `provision` and getting access to a handle to the running binary at run time e.g. (on Linux) via `/proc/self/fd`, then reading this data in OCaml. Unfortunately this won't work across all targets, e.g. at least `virtio`, `muen` and probably also `xen`.
2. We could have `build` produce relocatable objects and do a final link at `provision` time, but I'd stay away from that as it requires having most of the toolchain around for `provision` and would be somewhat fragile.
3. There is a compromise option. If we can live with the requirement that the _maximum_ size of embedded data would _always_ be reserved in a `.data` segment, and this size would be defined at `build` time, then we could essentially reserve a `static char __embedded_data[size];` during `build` and patch the "real" data in at `provision` time. This should work across all targets, since the data will get loaded by "whichever mechanism the target uses to load the binary".

[...]

PR #749, opened 2016-12-24
Syslog

hannesm:
this adds syslog support to mirage (via logs-syslog), the CLI.  an example of its usage can be seen [here](https://github.com/mirleft/nqsb.io/blob/cf962bc700a16049fb0ee1933e0a11e9b7c3f3a1/config.ml#L14-L27).

This PR leaves the syslog configuration solely in config.ml (with code produced by the mirage tool), the unikernel itself does not need to care about setting up the connections.  Log messages before the network stack is available are output to console (via mirage-logs).

While I'd appreciate to have this in Mirage3, I'm not yet happy with logging (but those issues can wait until 3.1 or 4.0):
- mirage-logs does two things: a console reporter, and a ring buffer for "debug on crash" -- we should move the ring buffer code to mirage-runtime (and have a `--debug-on-crash` command line key -- atm since `Logs.set_reporter` is called in `connect`, the ring buffer will be rather useless in the syslog case (it will only contain startup log messages))
- when logging via syslog is requested, we should avoid printing to the console (unless in error conditions, as done in logs-syslog) -- this also means we might want to preserve the boot messages in some ring buffer and send them over the wire once we have a UDP/TCP/TLS stack
- logging target should be a deployment time key (thus, a boot parameter: no-log | console | syslog config), the log level is already a command-line (configure/boot time) parameter (related to #694)
- nobody has a udp/tcp in their hand (this is why I now use stackv4 in logs-syslog instead of udp/tcp), the common monolithic thing you have is a stackv4 -- which is very v4 specific, and also not very modular (no clue how a ppp or ipsec link should look like in a stackv4 (it demands an ipv4 stack which requires arp))
- the tls vs tcp syslog should instead use a flow (in logs-syslog), and leave connection establishment up to the caller -- but then the reconnect code needs to be somewhere (not sure what the API should be here) -- I suspect conduit has this somehow in mind, but my brain fails to understand conduit


samoht:
Thanks! Having syslog combinators looks like a good idea. Just please revert 11df4d9 here too :p 

We will devise a way to distribute devices in separate packages in Mirage4 :p

hannesm:
I reverted the commit you don't like, and did the reverse in 8e92179d

PR #750, opened 2016-12-25
being able to choose arp implementation

hannesm:
PLEASE only look at the commit 74a751c (PR on top of #749)

This enables someone to use a slightly different ARP implementation (named `arp'`) without any hashtables, and with other timeout and retry values.  The source of this arp library is https://github.com/hannesm/arp

The common functions to construct a stack, `dhcp_ipv4_stack`, `static_ipv4_stack`, and `qubes_ipv4_stack` now have an `?arp:(?clock -> ?time -> ethernet -> arp)` argument, and a new `arp'` value is constructed in `mirage.ml`.

This again shows to me how hard it is to use mirage in a modular way -- both tcpip.arpv4 and arp.mirage have the very same interface, but nevertheless we need to specify two devices, and in order to make it usable these devices need to live in the mirage cli.  This means that the mirage cli has global (closed) knowledge over all devices, and using custom bits and pieces is hard (without modifying mirage.ml).  Earlier I used to modify main.ml, but this is very painful if you run a handfull of unikernels.

Another issue, while being here, is optional arguments to the mirage functions: not all respect all, thus deciding _globally_ on a specific arp/udp/time/clock is rather hard.  I suspect we should follow the `random` approach: single key, and expose a `default_random` inside of mirage, used by all the implementations (or is there demand to use a different `clock` implementation for your different pieces of the stack (X.509 / TCP)?  I'd suspect that not, well, at least I haven't come across any use case.

I currently use this branch for my unikernels (nqsb.io/ns.nqsb.io/btc-pinata/...) (was tired of way too many ARP timeout log messages).


lib/mirage.mli
```diff
@@ -267,6 +267,9 @@ val arpv4 : arpv4 typ
 val arp : ?clock:mclock impl -> ?time:time impl -> ethernet impl -> arpv4 impl
 (** Default ARPv4 implementation provided by the tcpip library *)
 
+val arp' : ?clock:mclock impl -> ?time:time impl -> ethernet impl -> arpv4 impl
```
samoht: I am really not fond of `'`.

lib/mirage.mli
```diff
@@ -376,18 +379,20 @@ val direct_stackv4 :
 val socket_stackv4 :
   ?group:string -> Ipaddr.V4.t list -> stackv4 impl
 
+type arp_sig = ?clock:mclock impl -> ?time:time impl -> ethernet impl -> arpv4 impl
+
 (** Build a stackv4 by looking up configuration information via QubesDB,
  *  building an ipv4, then building a stack on top of that. *)
-val qubes_ipv4_stack : ?group:string -> ?qubesdb:qubesdb impl -> network impl -> stackv4 impl
+val qubes_ipv4_stack : ?group:string -> ?qubesdb:qubesdb impl -> ?arp:arp_sig -> network impl -> stackv4 impl
```
samoht: why this isn't `?arp:(ethernet impl -> arpv4 impl)`? Does the stack really care about your arp device needing a clock/time device?

hannesm: this is related to https://github.com/mirage/mirage/pull/743#issuecomment-268390620 - I personally don't think there is much value in clock/time, but I'd want to use only a single time&clock in a unikernel.  if we provide `?clock` for some combinators, but use `default_clock` hardcoded in others, this does not allow to use a custom clock.

the other reason is that ocaml complained that the signatures don't match (I suspect I've to partially apply the functions with `clock` and `time` being `None`).

samoht: you can let people build their own `arp` implementation (using the clock they want), I don't think we need to thread that through all the arguments.

hannesm: ok. done in a455761

lib/mirage.mli
```diff
@@ -292,20 +292,20 @@ type ipv4_config = {
 
 type ipv6_config = {
   addresses : Ipaddr.V6.t list;
-  netmasks : Ipaddr.V6.Prefix.t list;
-  gateways : Ipaddr.V6.t list;
+  netmasks  : Ipaddr.V6.Prefix.t list;
+  gateways  : Ipaddr.V6.t list;
```
samoht: not sure why these are needed

samoht:
LGTM, thanks for the changes.

PR #755, opened 2016-12-28
Use mirage-kv and mirage-fs

samoht:


samoht:
Travis is green, so merging.

@avsm datakit/CI needs more disk space,

PR #756, opened 2016-12-28
Use mirage-fs instead of ocaml-fat to transform FS into KV_RO

samoht:


samoht:
Restarted TravisCI after https://github.com/mirage/ocaml-crunch/pull/32

samoht:
Restarted after https://github.com/mirage/mirage-http/pull/33

PR #767, opened 2017-01-08
mirage-runtime: requires mirage-device to build

avsm:


avsm:
now CI breakage is on dns.mirage package renaming

PR #769, opened 2017-01-10
mirage-types-lwt: minor fixes

hannesm:
- adjust merlin
- add mirage-stack{-lwt} mirage-protocols{-lwt} to META of mirage-types{-lwt}
- export Mirage_types_lwt.{PCLOCK, MCLOCK} (provided by mirage-clock-lwt)

imho straightforward @yomimono

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -7,6 +7,8 @@ PKG astring logs bos mirage-runtime
 PKG mirage-device mirage-time mirage-time-lwt mirage-random
 PKG mirage-flow mirage-flow-lwt mirage-console mirage-console-lwt
 PKG mirage-block-lwt mirage-net-lwt mirage-fs-lwt mirage-kv-lwt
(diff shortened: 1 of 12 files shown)


hannesm:
travis failures include that `dns.mirage` is not installed!? and some internal solver errors.. all not related to this PR.

hannesm:
the `dns.mirage` is likely fixed by https://github.com/mirage/ocaml-dns/pull/124

hannesm:
(CI still failing with dns.mirage, likely because the dns + mirage-dns opam file in `mirage-dev` remote are not yet updated)

hannesm:
(to make this explicit here: this change won't introduce a strict 4.03 dependency -- it removes `open Result` and the `result` package from the dependencies, but the `open` was unused (since the signatures moved to different repositories + opam packages), and thus `mirage-types{-lwt}` can happily be compiled with 4.02 and without result.

PR #779, opened 2017-01-17
refer to tls logs-syslog module by proper name

yomimono:


yomimono:
@hannesm, I'd appreciate confirmation that this fix is correct.  Thanks!

hannesm:
thanks!

PR #783, opened 2017-01-19
Add an "ldpostflags" variable for Solo5 targets

mato:
Accounts for future flexibity in final link runes for Solo5 targets by
adding an "ldpostflags" pkg-config variable, the contents of which (if
present) are appended to the end of the final link command.

Also fixes a potential bug in the `pkg_config` callout function where
the result should be an empty list if the output is empty.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1723,7 +1723,7 @@ let pkg_config pkgs args =
   Bos.OS.Env.set_var var (Some value) >>= fun () ->
   let cmd = Bos.Cmd.(v "pkg-config" % pkgs %% of_list args) in
   Bos.OS.Cmd.run_out cmd |> Bos.OS.Cmd.out_string >>| fun (data, _) ->
-  String.cuts ~sep:" " data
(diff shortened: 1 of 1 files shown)


mato:
The Travis failures here appear to have something to do with a dependency on `libpcre`, perhaps this needs to be added to the Travis setup explicitly?

avsm:
Pcre should normally be installed via `opam depext` which will find the OS package in the OPAM `conf-pcre` package. Is depext not being called for some reason...

mato:
@avsm I've no idea. There's also a failure building `dns`. Both failures are unrelated to this change. 

Example failures:

Points where it fails:
https://travis-ci.org/mirage/mirage/jobs/193477073#L3443
https://travis-ci.org/mirage/mirage/jobs/193477073#L3619

Stderr:
https://travis-ci.org/mirage/mirage/jobs/193477073#L3634

Failing command:
https://travis-ci.org/mirage/mirage/jobs/193477073#L3735

yomimono:
Merging to get this in before tag; I'm looking at `dns` and a few others today.

PR #785, opened 2017-01-30
explain why we make an empty myocamlbuild.ml

yomimono:


hannesm:
we're talking about this piece of ocamlbuild: https://github.com/ocaml/ocamlbuild/blob/0eb62b72b5abd520484210125b18073338a634bc/src/options.ml#L375-L387 -- maybe worth including for future reference (including commit id, otherwise it soon will point somewhere in space)?

PR #793, opened 2017-02-24
remove EXTRA_REMOTE for mirage-dev from travis config

yomimono:


yomimono:
CI still fails for `ukvm` targets, but tests `mirage-skeleton` and passes for other targets.  Merging for now and will continue to work on fixing the ukvm builds.

PR #795, opened 2017-03-01
Add a --output CLI option to overwrite the unikernel name

samoht:


samoht:
This needs https://github.com/mirage/functoria/pull/108 to be merged first

samoht:
With these patches, you can do:

```
$ mirage configure -o foo
$ make depends && make
$ ./foo
```

Which makes it simpler to embed MirageOS unikernels in scripts.

yomimono:
`foo` needs to be stripped of troublesome characters - try `mirage configure -o foo_oh_wow_amazing; make depend`.

samoht:
@yomimono not sure to see your point, `mirage configure -o foo_oh_wow_amazing;` works fine :-)

yomimono:
Sorry, try something like `-o plart.blorf` - the `make depend` step goes into a recurring failure, asking you to edit the generated opam metadata repeatedly.  `.` needs to be stripped to avoid that.

samoht:
Ha yes, so only the name of the opam file should be sanitized. I'll add that.

samoht:
Done.

hannesm:
Looks like a sensible thing to do and a reasonable PR.

yomimono:
It's functionally good (and a good idea!) and the PR to mirage looks reasonable to me, but I don't understand the functoria side well enough to merge it.  I'm happy to merge this once the functoria change is in.

lib/mirage.ml
```diff
@@ -1466,6 +1466,31 @@ module Substitutions = struct
...
+    ) s;
+  let s' = Buffer.contents b in
+  if String.length s' = 0 then raise (Invalid_argument s);
+  s'
+
+let long_name name target =
+  Fmt.strf "mirage-unikernel-%s-%a" (opamify name) Key.pp_target target
+
+let unikernel_name i target = match Info.output i with
+  | Some o -> opamify (Filename.basename o)
```
hannesm: curious why you take the `basename` here? AFAICS, output should never contain any `/` characters (or are they handled well, is `mirage configure -o /tmp/foobar` supposed to work?) -- this invariant should be checked in functoria while parsing that option.

lib/mirage.ml
```diff
@@ -1762,20 +1783,23 @@ let ldflags pkg = pkg_config pkg ["--variable=ldflags"]
 
 let ldpostflags pkg = pkg_config pkg ["--variable=ldpostflags"]
 
-let link info name target =
+let link info target =
   let libs = Info.libraries info in
+  let output = output info in
+  Bos.OS.Cmd.run Bos.Cmd.(v "mkdir" % "-p" % Filename.dirname output)
```
hannesm: `Bos.OS.Dir.create` (see http://erratique.ch/software/bos/doc/Bos.OS.Dir.html#VALcreate) will work on more platforms and do the same AFAICS

lib/mirage.ml
samoht: Yes, `mirage configure -o bin/foo` is supposed to work.

hannesm: ic, and its opam file will be `foo.opam`!?  Maybe worth documenting somewhere (but unclear to me where to do this best).

yomimono: When I configure with `-t ukvm` and `-o /otherdir/elsewhere`, should I get the `ukvm-bin` in the destination directory too?  I'll need it to do anything with the artifact, but it stays in the build directory currently:

```
~/mirage-skeleton/applications/static_website_tls $ mirage configure -t ukvm -o /home/user/unikernels/website
~/mirage-skeleton/applications/static_website_tls $ mirage configure -t ukvm -o /home/user/unikernels/website && make depend && make
~/mirage-skeleton/applications/static_website_tls$ ls ~/unikernel_apiary/
website.ukvm
```

hannesm:
what is the status of this PR?  could you rebase and address the comments above, so we can merge this? :)

samoht:
I don't have a need for this anymore and I don't feel very confortable adding this feature without tests (due to the various possible ways this could fail), so I am closing that PR. I'll maybe come back to it at one point in the future.

yomimono:
>  I don't feel very confortable adding this feature without tests 

If you have time, I'd be curious to hear which tests you'd want to see.  I want to expand the test wishlist beyond `tcpip`; if we can think of good tests for `mirage` itself, that would be a great place to head next.

samoht:
@yomimono basically it was not clear to me what `-o` would mean for the different targets (especially when multiple files are generated). Having similar end-to-end tests like on functoria for the `noop` unikernel and check that the right things are done for some mirage-specific CLI calls could be useful.

Issue #798, opened 2017-03-03
Mirage module should support mirage-block-ramdisk & generic_block


yomimono:
A generic_block which allows the user to specify the block implementation at compile-time (as is possible with generic_kv_ro) would be useful; also we should be able to get a mirage-block-ramdisk from some function in the Mirage module.

Suggested by @g2p - thanks!

PR #800, opened 2017-03-03
Add a --block configure flag for picking ramdisk or file-backed block…

g2p:
… devices

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -406,11 +406,10 @@ class block_conf file =
 let block_of_file file = impl (new block_conf file)
 
 class ramdisk_conf rname =
-  let name = Name.create rname ~prefix:"ramdisk" in
-  object (self)
+  object
     inherit base_configurable
     method ty = block
-    method name = name
+    method name = "ramdisk"
     method module_name = "Ramdisk"
     method packages =
       Key.pure [ package "mirage-block-ramdisk" ]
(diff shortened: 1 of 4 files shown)


yomimono:
Fixes #798 .

g2p:
Hold on, my test case was off, the --block=ramdisk case isn't really building a ramdisk yet.

yomimono:
@g2p: try https://github.com/yomimono/mirage/tree/blockier - the method `name` has to be unique among the classes, otherwise there can be strange collisions when Functoria reduces the graph (see https://github.com/mirage/functoria/issues/79).

g2p:
Thank you for digging this up!

I will revise this. There needs to be a way to specify the ramdisk size.

g2p:
I've experimented with `--block-size-sectors` and `--block-sector-size` keys, but in the end, I think the size is best specified programmatically using [`Ramdisk.resize`](https://mirage.github.io/mirage-block-ramdisk/). Having flags that control the size in the ramdisk case and not in the block-file case would be confusing. Command-line flags also can't handle differing sizes for multiple devices.

So, please merge this branch as-is.

yomimono:
@g2p do you not need something like https://github.com/yomimono/mirage/commit/b16848b1c1761a7513b7b10a6ae8da1c971a620f ?  I don't recall the specific surprising behavior we got with this patch or else I'd test myself.

g2p:
@yomimono You're right, I had forgotten your patch when going back on my desktop.
The patch is necessary to pick the right module name in main.ml.
Thank you.

PR #801, opened 2017-03-03
IPv6 needs Random

hannesm:


yomimono:
Thanks @hannesm and @olleolleolle !

yomimono:
Thanks @hannesm and @olleolleolle !

lib/mirage.ml
```diff
@@ -729,13 +729,13 @@ let ipv4_qubes db ethernet arp = ipv4_qubes_conf $ db $ ethernet $ arp
 
 let ipv6_conf ?addresses ?netmasks ?gateways () = impl @@ object
     inherit base_configurable
-    method ty = ethernet @-> time @-> mclock @-> ipv6
+    method ty = ethernet @-> random @-> time @-> mclock @-> ipv6
     method name = Name.create "ipv6" ~prefix:"ipv6"
     method module_name = "Ipv6.Make"
     method packages = Key.pure [ package ~sublibs:["ipv6"] "tcpip" ]
```
hannesm: if `opam` would know better about dev packages, we could constrain this here... certainly there should be a `min:"3.0.1"`, but this will make the `ipv6_conf` unusable right now (until a tcpip-3.0.1 is released).

another idea would be to revive `mirage-dev` for this, and use `dev~mirage` as version (or `3.0.1`).  any suggestions?

yomimono: My plan was:
0) merge `make depends` fix to mirage
1) constrain mirage 3.0.0 to tcpip = 3.0.0
2) release tcpip 3.1.0 with mtu, improved random in ipv6, bugfixes (tcpip is already constrained to mirage-protocols* >= 1.1.0)
3) release mirage 3.0.1 depending on tcpip 3.1.0 (+mirage-protocols, mirage-protocols-lwt 1.1.0), and constraining the minimum tcpip version in generated unikernels to 3.1.0

I intend to do this today so not sure it's worth reviving mirage-dev, but if I take longer than that it probably is.

hannesm: it is not clear to me what exactly `1` means.  mirage does not depend on tcpip.  tcpip does not depend on mirage.   Instead, the unikernels generated by mirage configure should be constrained to tcpip = 3.0.0.  But since mirage3.0.0 is already out, this cannot be done (easily).

with `3`, we should be careful which interfaces/devices need these constraints.  Functoria's `package` mechanism can deal with lower and upper bounds, and merges thereof.

yomimono: Re `1`: You're right (and my thinking was too muddy on this), but I think we can help the situation by introducing a conflict -- mirage 3.0.0 conflicts with tcpip > 3.0.0 .  This still isn't quite right (as you say, the dependency is actually in the generated unikernel, but it's too late to do anything about that), but it should avoid the situation where a user has installs `mirage` and generates a unikernel that the `tcpip` they already have will be unable to use.

Your point re: `3` is well-taken, thanks for that.

yomimono:
Merging this and the companion tcpip PR; I'm planning to cut releases for mirage, tcpip, and charrua-core shortly.

PR #803, opened 2017-03-06
don't add mirage-dev remote for docs.mirage.io generation

hannesm:


avsm:
+1 if this builds (there might be straggling unreleased packages)

PR #805, opened 2017-03-06
add a depends target to the generated Makefile

### LINKED ISSUE #804: Add alias 'make depends'
I routinely type `make depends` when I mean `make depend`.  I don't think I'm alone in this. Could we add that as an alias?

yomimono:
Fixes #804 .

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1594,7 +1594,7 @@ let configure_makefile ~opam_name =
                   .PHONY: all depend clean build\n\
                   all:: build\n\
                   \n\
-                  depend::\n\
+                  depend depends::\n\
                   \t$(OPAM) pin add --no-action --yes %s .\n\
                   \t$(DEPEXT)\n\
                   \t$(OPAM) install --yes --deps-only %s\n\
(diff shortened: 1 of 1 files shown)


avsm:
This also needs an entry to `PHONY`.  Honestly, I would just pick either `depend` or `depends` and stick with it rather than having aliases in Makefiles, which just complicates them.

yomimono:
The current situation where `make depends` doesn't work is very unfriendly; I saw people type that, grumble a bit, and fix it at least six times yesterday.  `make depend` is more natural for at least the person who initially named it and probably some others, and is all over the documentation now besides.  I would really like to support both rather than add a stumbling block for our users which they'll encounter every time they use our tool.

hannesm:
I don't understand.  Should we then also support `Depend` (uppercase), `dep` (even shorter), and all other potential spellings done by a user?

Would users be happy if `mirage configure` (on success) would output what to do next?

yomimono:
If I saw users type `make Depend` or `make dep` a huge percentage of the time (and myself typed that quite frequently), I would say we should support that too, yes.  But note that I didn't propose that, because I haven't seen it; I'm attempting to address the actual problem that we've seen users make.

I don't know what makes users happy, but I know what makes them unhappy is the perception that their tools are fighting them.

yallop:
I don't have a strong view on the alias issue, but I think it's worth mentioning a nice usability feature that not everyone may be aware of: `make` has tab completion support for targets in some shells (`bash`, `zsh`, perhaps others), so you can write
```
make dep<TAB>
```

to input `make depend`

PR #806, opened 2017-03-07
doc: Temporarily disable some unreleased packages

avsm:
They will be restored once released; as they now must be in stable
opam for this to work since #803


hannesm:
`git-mirage` is actually released to opam, and should work..

avsm:
I'm stuck in constraint hell thanks to cmdliner 1.0 -- fixing that first then will revisit the "to fix" comments in this Dockerfile..

PR #807, opened 2017-03-07
Allow qubesdb to be used under Xen

talex5:
`-t qubes` is useful to run generic unikernels on Qubes, but for Qubes-specific unikernels we need to use the Xen target so that we can control the agents directly.

See e.g. https://github.com/talex5/qubes-mirage-skeleton/pull/2

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -104,7 +104,7 @@ let qubesdb_conf = object
   method packages = Key.pure [ package "mirage-qubes" ]
   method configure i =
     match get_target i with
-    | `Qubes -> R.ok ()
+    | `Qubes | `Xen -> R.ok ()
     | _ -> R.error_msg "Qubes DB invoked for non-Qubes target."
   method connect _ modname _args = Fmt.strf "%s.connect ~domid:0 ()" modname
 end
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -105,7 +105,7 @@ let qubesdb_conf = object
   method configure i =
     match get_target i with
     | `Qubes | `Xen -> R.ok ()
-    | _ -> R.error_msg "Qubes DB invoked for non-Qubes target."
+    | _ -> R.error_msg "Qubes DB invoked for an unsupported target; qubes and xen are supported"
   method connect _ modname _args = Fmt.strf "%s.connect ~domid:0 ()" modname
 end
 


yomimono:
thanks @talex5 !

hannesm:
it looks weird to me to have a qubes target and nevertheless use `-t xen` for unikernels running on qubes.  is it worth having a qubes target?  can the qubes target be massaged to allow controllng the agents explicitly?

yomimono:
>  is it worth having a qubes target?

I was kind of on the fence about `-t qubes`, but the only other way I could think of to get the thing I needed (automatic building of unikernels bootable on Qubes without changing the config or unikernel) was to add something like `--with-qubes`, which would then be invalid for all targets but `-t xen`.  `-t qubes` seemed more honest.

This PR is for users who are trying to accomplish a different goal than I was -- they're willing to forego being able to run their unikernels unmodified on other targets and have explicitly put the dependency in their configs.

> can the qubes target be massaged to allow controllng the agents explicitly?

Generally, we should provide methods for overriding the various stubs that `-t qubes` gives you by default.  This was left as a TODO in the original PR... further PRs welcome ;)

hannesm:
what about `qrexec` and `gui_qubes` then (both in mirage.mli)?

instead of including more workarounds for xen and qubes, why don't finish the TODO items then?

talex5:
Mirage doesn't seem to have a good way to override defaults. We had the same problem with the log reporter (if you specify a logger you should get that, but if you don't then you should get a default instead). We solved that with a custom argument to `register`.

We could expose `qrexec` and `gui_qubes`, but there's little benefit since if you want to target Qubes then you can create these manually. `qubesdb` is special only because we need it to set up networking and that's a pain to do manually.

hannesm:
On 09/03/2017 14:29, Thomas Leonard wrote:
> Mirage doesn't seem to have a good way to override defaults. We had the same problem with the log reporter (if you specify a logger you should get that, but if you don't then you should get a default instead). We solved that with a custom argument to `register`.

I suspect then someone should work on this by gathering demands and
implementation possibilities.  Adding another (completely different)
workaround (custom argument to register vs using wrong target) to
'solve' a similar problem might work in the short term, but does not
lead to clean reusable code, but rather to more messy code with
undocumented invariants and assertions.

The logging 'solution' is rather inconvenient, since I miss the
possibility to not use mirage-logs, but keep config.ml unmodified.  A
command-line (configuration-time) parameter would be much nicer.

Issue #761, opened 2017-01-02
Imposition of `-classic-display`


mor1:
Given there's no way to override `-classic-display` whereas we do have a means for users to inject their own flags via `Makefile.user`, please could we not impose `-classic-display` in the invocation of `ocamlbuild`?

hannesm:
In the meantime (in mirage master), ocamlbuild is not invoked via make anymore.  This is of course a downside, but there is an upside: there is no more too much string concatenation in mirage, and no dependency on make.

While I don't specifically care about `classic-display`, it got added since a large amount of our ecosystem uses it.  If you would like to make something configurable with third-party tools, there may be a generic solution to customise your `ld` and flags, as well as `ocamlbuild` and other tools invoked.  I'd dislike to have another ad-hoc solution such as `warn-error` for all possible `ocamlbuild` command line flags (why is `quiet` not exposed?).

`classic-display` is especially useful in unattended builds to see in the log (console output) what actually happened (such as bulk builds with opam).

just my two cents, feel free to submit a PR removing `classic-display` or making it user-configurable.  IMHO there are more pressing issues than this.

avsm:
Conversely: why do you want to get rid of `-classic-display` @mor1?

talex5:
Don't know about Mort, but I want to get rid of it because it's hard to see real warnings in between all the verbose output about compiler commands.

Also, it's a bit silly having the different layers of our build stack fighting with each other. e.g. the ocamlbuild authors decided to turn this off by default, then some build tool (e.g. topkg) decides to override that and turn it on, then the actual application decides to override it again to disable it, then maybe some user decides to override it again because they want it on in some particular case. It should be set once, under the ultimate control of the user.

avsm:
> Also, it's a bit silly having the different layers of our build stack fighting with each other

Yes, agreed.

dbuenzli:
> because it's hard to see real warnings in between all the verbose output about compiler commands.

A good editor allows you to directly jump to the warning and error locations.

> then some build tool (e.g. topkg) decides to override that and turn it on,

The reason is that this option makes it much easier to debug build problems (there is `_build/log` but it's interspread with much more information I'm usually not interested in).

mor1:
Basically -- what @talex5 said. 

To precis an IRC chat with @hannesm yesterday:
* This isn't fantastically important, but it was the first thing I noticed trying out Mirage3
[...]

PR #811, opened 2017-03-14
don't add "-color always" to ocamlbuild invocation, since it breaks some scripts

### LINKED ISSUE #797: Mirage build forces ANSI colour escapes
This breaks e.g. Vim's quickfix feature because it doesn't strip the escape codes before matching error patterns.

Looks like this was added in https://github.com/mirage/mirage/commit/85b6932f61f25d50c5f878518089534174da083e#diff-7727aac76b5d1f16253b3426f63177d4R1592.

/cc @hannesm

hannesm:
OCaml (since 4.03.0) attempts to colorise warnings and errors (in a "auto" mode
by default, which checks whether stdout is a tty).  ocamlbuild does not execute
ocaml in a tty, this is why by default there is no color.

An upstream fix to consider the environment variable OCAML_COLOR in case no
"-color" command line flag was provided is proposed, and may be included in
4.05.0.  In the meantime, revert to no colored output (as in Mirage <3.0.0).

Fixes #797 (and @haesbaert mentioned the same in some emacs script https://github.com/mirage/mirage/issues/761#issuecomment-280501942)

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1671,7 +1671,6 @@ let compile libs warn_error target =
   let tags =
     [ Fmt.strf "predicate(%s)" (backend_predicate target);
       "warn(A-4-41-42-44)";
-      "color(always)";
       "debug";
       "bin_annot";
       "strict_sequence";


yomimono:
Thanks for this and the upstream work, @hannesm!

PR #820, opened 2017-03-30
ukvm-configure expects a list of modules (net/blk), and is (at least …

hannesm:
…here)

not happy with lots of empty arguments.

previously:
 'ukvm-configure' '/src/ukvm' '' '' '' '' '' ''
after:
 'ukvm-configure' '/src/ukvm'

//cc @mato

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1832,12 +1832,11 @@ let link info name target =
                %% of_list ldpostflags)
     in
     let ukvm_mods =
-      let ukvm_filter = function
-        | "mirage-net-solo5" -> "net"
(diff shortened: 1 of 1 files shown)


mato:
AFAIK that's only the new multi-backend/arch version of ukvm-configure, but worth fixing anyway. +1

hannesm:
CI breakage due to mirage-skeleton dns example, fix is over in the mirage-skeleton repository:  https://github.com/mirage/mirage-skeleton/pull/229

yomimono:
Thanks for this, @hannesm!

PR #822, opened 2017-04-09
fallback to system $PKG_CONFIG_PATH

timbertson:
In not-strictly-opam environments (e.g. opam2nix), `opam config var prefix` may fail or not be where dependencies actually are. This patch allows falling back to `pkg-config` libraries in the system's `$PKG_CONFIG_PATH`, if set.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1732,12 +1732,15 @@ let opam_prefix =
 
 (* Invoke pkg-config and return output if successful. *)
 let pkg_config pkgs args =
+  let var = "PKG_CONFIG_PATH" in
+  let pkg_config_fallback = match Bos.OS.Env.var var with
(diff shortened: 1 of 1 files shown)


hannesm:
Thank you for this PR!  but could we instead of using `Sys` and its very brittle (with lots of exceptions) interface continue to use Bos here (http://erratique.ch/software/bos/doc/Bos.OS.Env.html contains functionality to read environment variables with explicit error behaviour (i.e. you get an `option` or `result` type back)).  It took me some effort to get rid of all those calls to `Sys` and `Unix`.

Thanks!

timbertson:
Fair enough - rewritten to use `Bos` 👍

yomimono:
Thanks for this patch, @timbertson !

PR #823, opened 2017-04-09
expand ocamlfind paths using ocamlfind, instead of assuming opam layout

timbertson:
This PR uses `ocamlfind query` to expand `@name/path` syntax, instead of assuming libraries  are installed in `opam config var prefix`.

I needed this to build a mirage ukvm image with dependencies resolved by opam2nix, however it also relies on https://github.com/ocaml/opam-repository/pull/8914 to resolve `gmp-freestanding` via ocamlfind. There may be other dependencies I don't know of which also need a dummy `META` file to be found.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -190,7 +190,7 @@ let stdlib_random = impl stdlib_random_conf
 
 let query_ocamlfind ?(recursive = false) ?(format="%p") ?predicates libs =
   let open Bos in
-  let flag = if recursive then "-recursive" else ""
+  let flag = if recursive then [ "-recursive" ] else []
   and format = Cmd.of_list [ "-format" ; format ]
   and predicate = match predicates with
     | None -> []
@@ -198,7 +198,7 @@ let query_ocamlfind ?(recursive = false) ?(format="%p") ?predicates libs =
   and q = "query"
   in
   let cmd =
-    Cmd.(v "ocamlfind" % q % flag %% format %% of_list predicate %% of_list libs)
+    Cmd.(v "ocamlfind" % q %% of_list flag %% format %% of_list predicate %% of_list libs)
   in
   OS.Cmd.run_out cmd |> OS.Cmd.out_lines >>| fst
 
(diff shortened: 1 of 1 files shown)


lib/mirage.ml
```diff
@@ -190,15 +190,15 @@ let stdlib_random = impl stdlib_random_conf
...
+  let flag = if recursive then [ "-recursive" ] else []
   and format = Cmd.of_list [ "-format" ; format ]
   and predicate = match predicates with
     | None -> []
     | Some x -> [ "-predicates" ; x ]
   and q = "query"
   in
   let cmd =
-    Cmd.(v "ocamlfind" % q % flag %% format %% of_list predicate %% of_list libs)
+    Cmd.(v "ocamlfind" % q %% of_list flag %% format %% of_list predicate %% of_list libs)
```
timbertson: IIRC I needed this because with `recursive=false` it was passing an empty string argument to `ocamlfind` instead of omitting it, but I'm not very confident about all the operators here. Let me know if it looks suss...

avsm:
nice! I don't use Nix, but have merged https://github.com/ocaml/opam-repository/pull/8914 to opam-repo-stable which should help.  Is there a docker image of nix available so we can test it as part of the regular OPAM/Mirage CI?

timbertson:
That would be great. There's https://hub.docker.com/r/nixos/nix/, which has the nix base. `opam2nix` is not in mainline nix, so you would need to grab it yourself as part of the build (the only file you need is https://github.com/timbertson/opam2nix-packages/blob/master/nix/release.nix).

If you have an example build / setup I could help figure out how a nix version would work to test the same sorts of things. Currently the opam2nix CI does the automated pulling of `opam-repository` and regenerating the .nix expressions once a week, so if you wanted to integrate the latest master (or a PR) of `opam-repository` instead there would be more fiddling required.

djs55:
I took the liberty of force pushing a rebase against latest master to hopefully fix the CI. I hope this is ok!

timbertson:
Thanks @djs55. I can't figure out the latest build error - it fails consistently for all the 4.04.2, and the log has:

```
[gmp-xen: sh] Command started
-> installed gmp-xen.6.0.0
[zarith-xen: sh] Command started
-> installed zarith-xen.1.4
[nocrypto: ocaml build] Command started
-> installed nocrypto.0.5.4
Done.

( ... )

mirage build

( ... )

ocamlfind: Package `gmp-xen' not found
```

I don't see any place where it's uninstalling `gmp-xen` between those lines, so I'm not sure what the problem is.

lib/mirage.ml
hannesm: this change looks fine to me!

lib/mirage.ml
```diff
@@ -1739,14 +1739,16 @@ let compile libs warn_error target =
   Bos.OS.Cmd.run cmd
 
 (* Implement something similar to the @name/file extended names of findlib. *)
-let rec expand_name ~lib param =
+let rec expand_name param =
+  let pkg_path name = query_ocamlfind ~format:"%d" [name] >>| List.hd in
```
hannesm: hmm, this `List.hd` can fail, or!?  since this is already in a `result` type, shouldn''t it rather be:
```
query_ocamlfind ~format:"%d" [name] >>= function
| x::_ -> R.ok x
| _ -> Error (`Msg "???")
```

but what is the actual semantics?  is `ocamlfind` in this case expected to return exactly one item?

samoht:
I am closing this pull request as the underlying repository has changed quite a bit since it has been opened and we are moving away from ocamlbuild/ocamlfind to use dune exclusively (still WIP but the support should be merged soon). Happy to discuss Nix support once this is done.

timbertson:
> moving away from ocamlbuild/ocamlfind to use dune exclusively

Cool, I expect that to be much less hassle then since most dune stuff works fine under nix 👍

PR #832, opened 2017-06-14
RFC: Use ocamlfind and ocamlobjinfo to find Xen stubs

djs55:
When OCaml is linking a normal binary, it discovers the `stubs.a` files needed by the `library.cmxa` files and includes them. However when linking a Xen/Qubes unikernel we control the final link ourselves, so we need to know the list of `stubs.a` files to include.

The current behaviour is to add an extra key=value to the META file, e.g.:
```
  xen_linkopts = "-lio_page_xen_stubs"
```
and to build 2 versions of the stubs: one for Unix (e.g. `io-page.unix`) and one for Xen (`io-page.xen`) (often these stubs are substantially the same with a few differences guarded by `#ifdef`s)

This patch proposes to discover the `stubs.a` files from the OCaml archives, as the OCaml compiler normally would, so we don't have to edit the META file.

We use the ocamlfind query suggested by @diml to list the archives from a given ocamlfind package:
```
  ocamlfind query -r -format "%+a" -predicates native <package>
```
and then parse the output of ocamlobjinfo, for example:
```
$ ocamlobjinfo /home/opam/.opam/4.04.1/lib/io-page/io_page_xen.cmxa
File /home/opam/.opam/4.04.1/lib/io-page/io_page_xen.cmxa
Extra C object files: -lio_page_xen_stubs
Extra C options:
```

To distinguish between Unix stubs and Xen stubs, we would still need to have a convention like that used by io-page, i.e. insist that if a library `foo` has Mirage-compatible stubs, it must have a subpackage called `foo.xen` which will be queried and used for linking. This part is unimplemented.

WDYT?

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -202,6 +202,35 @@ let query_ocamlfind ?(recursive = false) ?(format="%p") ?predicates libs =
   in
   OS.Cmd.run_out cmd |> OS.Cmd.out_lines >>| fst
 
+type ocamlobj = {
+  extra_c_object_files: string;
+  extra_c_options: string;
+}
+let query_ocamlobjinfo cmxs =
+  let open Bos in
+  let cmd = Cmd.(v "ocamlobjinfo" % cmxs) in
+  OS.Cmd.run_out cmd |> OS.Cmd.out_lines >>| fst
+  >>= fun lines ->
+  let none = {
+    extra_c_object_files = "";
+    extra_c_options = "";
+  } in
+  let extra_c_object_files_prefix = "Extra C object files: " in
+  let extra_c_options_prefix = "Extra C options: " in
+  let r = List.fold_left (fun acc line ->
+    let acc = match String.cut ~sep:extra_c_object_files_prefix line with
+      | None -> acc
+      | Some ("", extra_c_object_files) -> { acc with extra_c_object_files }
+      | _ -> assert false in
+    let acc = match String.cut ~sep:extra_c_options_prefix line with
+      | None -> acc
+      | Some ("", extra_c_options) -> { acc with extra_c_options }
+      | _ -> assert false in
+    acc
+  ) none lines
+  in
+  R.ok r
+
 (* This is to check that entropy is a dependency if "tls" is in
    the package array. *)
 let enable_entropy, is_entropy_enabled =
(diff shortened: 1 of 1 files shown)


talex5:
I believe the reason for not getting the stubs automatically is that if you get it wrong (and link a library that isn't compiled for Xen) then the failure mode is very bad: the code links and appears to work for a while, but will randomly crash if an interrupt occurs at the wrong time (because non-Xen code assumes it has a red zone to play with).

Only linking libraries mentioned in `xen_linkopts` makes it much harder to link a Unix library by accident.

djs55:
@talex5 thanks for the explanation! I guess we have the opposite problem: the `cstruct` package doesn't know it should build Xen compatible stubs, and if it wasn't for the fact that we have a hack for that already :) we would fail to link those Unikernels too.

I guess another way to think about this proposal is: should the convention be a `xen_linkopts` inside the `META` file, or should it be a find lib package `foo.xen` (or should it be an archive with a particular name (e.g. `io_page_xen.a`) I don't know if this makes a big enough difference to merit the change.

I wonder if there's any way to do better than [postconf.ml](https://github.com/mirage/io-page/blob/master/postconf.ml) -- do we have to go around every package which has stubs which should work on Xen and add this?

I wonder if we should install the original `foo.c` files in the `ocamlfind` directory (like we sometimes do for javascript) and have the `mirage build` step compile them with the correct options? At least then the configuration logic would be in one place.

Is it possible to use something more like cross-compilation to build the same stubs twice without having to change the build rules in the package with the stubs?  Perhaps we should wait until we can put everything inside one `jbuilder` workspace and then build globally with Xen-specific options?

dbuenzli:
> I wonder if there's any way to do better than postconf.ml -- do we have to go around every package which has stubs which should work on Xen and add this?

Obviously that won't scale. I know I expressed this more than one time but like for library variants showing up in the ocamlfind package namespace (`.jsoo`, `.os`, etc.), I always found these `-{unix,xen}` opam packages suspicious. 

I really think that the `xen` vs `unix` should be treated as an architecture/os distinction rather than try to make them live in the same package universe. That is they should live in different opam switches with appropriately set `os` variable.

djs55:
I think I agree with @dbuenzli -- ideally the packages should be built separately, once for "unix" and once for "xen" (and again for whatever other architectures are defined) and we wouldn't have to expose the architecture in the package name.

I think what bothered me before about using switches before was my inability to keep the switches in sync: they'd end up with different packages and pins, and then I would do something bad like run `make install` where the `oasis` `setup.data` was stale, and then pollute the wrong switch. Perhaps I could have learned better habits.

Perhaps in future I'll be able to check out all the code into one large "workspace" (is that what `jbuilder` calls it?) and define multiple build environments: the "xen" one can have the magic `CFLAGS` set globally. Perhaps I'll be able to build the same code twice, into 2 different `_build` directories?

Anyway, I'll close this PR -- at least it helped me learn how the code currently works. Thanks for the feedback.

Issue #775, opened 2017-01-16
expression of io-page is inconsistent


yomimono:
There is no longer a module type for `io-page` exposed in Mirage_types or Mirage_types_lwt, but the `mirage` tool's API still has types representing it and a `default_io_page`.

hannesm:
not anymore since #940

Issue #699, opened 2016-11-22
io-page dependency


hannesm:
as remarked on https://github.com/mirage/mirage/pull/585#issuecomment-261794358, the dependency onto `io-page` should be there due to `method deps = [ abstract default_io_page ]` in mirage.ml, but it looks like it does not get added.  it might be an issue in functoria.

worth digging into (maybe re-try after #691)

hannesm:
I tested by removing the dependency, and both `io-page` and `io-page.unix` were still present when using crunch (using mirage-skeleton's kv_ro_crunch).

hannesm:
surprisingly this works for unix/macosx and xen/qubes, but fails for virtio/ukvm (see travis run at https://travis-ci.org/hannesm/mirage-dev/jobs/179095516

@Drup any idea where this could come from?

Drup:
What's the output of `describe` in the failing cases ?

hannesm:
unclear. cannot reproduce this behaviour locally... weird..

samoht:
`io-page` and `io-page.unix` unfortunately relies on a very bad linking trick to work: you need to specify the dependencies in the right order (and hope that ocamlfind will not re-order them for you).

tbh, I wish we could remove that horrible hack.

hannesm:
since #940 was merged, this is no longer an issue

PR #833, opened 2017-06-14
Automatically require io-page-{xen,unix} on Xen and Unix

djs55:
Previously the correct stubs were found because `io-page` was in the
dependency list and everything was in the same `META` file with global
`xen_linkopts`. [mirage/io-page#46] splits `io-page` into 3 packages
and hence 3 distinct `META` files. Therefore we must depend explicitly
on `io-page-xen` to discover the `xen_linkopts`. This patch also adds
a dependency on `io-page-unix` to the `Unix` target for symmetry, although
there is no `xen_linkopts`: the OCaml compiler figures that out for itself.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1933,12 +1933,14 @@ module Project = struct
           package ~build:true "ocamlbuild" ;
         ] in
         Key.match_ Key.(value target) @@ function
-        | `Xen | `Qubes -> [ package ~min:"3.0.0" "mirage-xen" ] @ common
+        | `Xen | `Qubes -> [ package ~min:"3.0.0" "mirage-xen";
+                             package "io-page-xen" ] @ common
         | `Virtio -> [ package ~min:"0.2.1" ~ocamlfind:[] "solo5-kernel-virtio" ;
                        package ~min:"0.2.0" "mirage-solo5" ] @ common
         | `Ukvm -> [ package ~min:"0.2.1" ~ocamlfind:[] "solo5-kernel-ukvm" ;
                      package ~min:"0.2.0" "mirage-solo5" ] @ common
-        | `Unix | `MacOSX -> [ package ~min:"3.0.0" "mirage-unix" ] @ common
+        | `Unix | `MacOSX -> [ package ~min:"3.0.0" "mirage-unix" ;
+                               package "io-page-unix"; ] @ common
 
       method build = build
       method configure = configure


djs55:
This will need https://github.com/mirage/io-page/pull/46 to build

hannesm:
but this dependency is misplaced, isn't it?  there is already an io_page_conf device in the very same file, which -- depending on platform -- depends on io-page or io-page.unix.  could we please not require _every existing_ mirageos unikernel to have an enforced dependency onto io-page, but instead only those using libraries which need io-page!?

djs55:
As I think has been noted elsewhere, there is some asymmetry between current handling of stubs in Unix / Xen / Solo5:
- all Unix unikernels link against `mirage-unix` which  has a dependency on the findlib package `io-page.unix` and the opam package `io-page`.
- all Xen unikernels link against `mirage-xen` which has a dependency on the findlib package `io-page` which uses the special `META` file `xen_linkopts` interpreted by the `mirage` tool to bring in the `io-page.xen` subpackage
- all Solo5 unikernels link against `mirage-solo5` which contains custom `io-page` C stubs

I'm not proposing to solve completely this (hard) problem right now, but I am keen to make forward progress :) I think the ultimate solution will be to put the stubs next to the OCaml code (in `io-page`, `cstruct`) and cross-compile it properly for each build target. Perhaps we'll be able to type "make" and build the same unikernel for all Mirage targets simultaneously.

I think removing the `depopt` which created the `io-page-xen` top-level package is a genuine improvement. The move from `io-page.unix` to `io-page-unix` is marginal: it's only for symmetry with the `-xen` case.

I might not have understood properly, but wouldn't using an `io_page_conf` device either mean that users have to add `default_io_page` to many of their projects (i.e. its a breaking change) or we would have to mirror the `opam` dependencies somehow so `class block_conf` would depend on both `mirage-block-unix` and `io-page-unix`?

I do agree we should avoid magic extra dependencies in the `mirage` tool, and to push the dependencies down the chain to nearer where they're needed. I propose that a simple improvement would be to

- remove this magic from the `mirage` tool.
- replace the `io-page` dependency in [mirage-xen.opam](https://github.com/mirage/mirage-platform/blob/84431dd1adba84338598d71edf6bcb7cecdb040c/mirage-xen.opam#L18) with `io-page-xen` (and the _oasis file)
- replace the `io-page` dependency in [mirage-unix.opam](https://github.com/mirage/mirage-platform/blob/84431dd1adba84338598d71edf6bcb7cecdb040c/mirage-unix.opam#L17) with `io-page-unix` (and the _oasis file)

avsm:
I am in favour of removing the depopts to start with and making the appropriate changes in mirage-xen and mirage-unix as proposed. I can't actually remember why it's not a hard dependency like this to begin with, in the relevant mirage-<os> packages...

hannesm:
On 15/06/2017 21:33, David Scott wrote:
> As I think has been noted elsewhere, there is some asymmetry between current handling of stubs in Unix / Xen / Solo5:
> - all Unix unikernels link against `mirage-unix` which  has a dependency on the findlib package `io-page.unix` and the opam package `io-page`.
> - all Xen unikernels link against `mirage-xen` which has a dependency on the findlib package `io-page` which uses the special `META` file `xen_linkopts` interpreted by the `mirage` tool to bring in the `io-page.xen` subpackage
> - all Solo5 unikernels link against `mirage-solo5` which contains custom `io-page` C stubs

Please see https://github.com/mirage/mirage-platform/pull/188 for
mirage-unix not depending on io-page.

> I'm not proposing to solve completely this (hard) problem right now, but I am keen to make forward progress :) I think the ultimate solution will be to put the stubs next to the OCaml code (in `io-page`, `cstruct`) and cross-compile it properly for each build target. Perhaps we'll be able to type "make" and build the same unikernel for all Mirage targets simultaneously.

That's sad.  Shouldn't we fix this once and for all?  Please see
https://github.com/mirage/mirage-platform/issues/124 for a related
discussion from years ago.. AFAIR the consensus was to move the C stubs
into the libraries which use them (as you propose).  It seems like
nobody managed to invest enough time to do so.

> I think removing the `depopt` which created the `io-page-xen` top-level package is a genuine improvement. The move from `io-page.unix` to `io-page-unix` is marginal: it's only for symmetry with the `-xen` case.
> 
> I might not have understood properly, but wouldn't using an `io_page_conf` device either mean that users have to add `default_io_page` to many of their projects (i.e. its a breaking change) or we would have to mirror the `opam` dependencies somehow so `class block_conf` would depend on both `mirage-block-unix` and `io-page-unix`?

The libraries which need io-page should depend on it.  An example is
`crunch`, which generates code which uses io-page, and thus has a
dependency on io-page-conf in mirage (but there seem to be some issues,
https://github.com/mirage/mirage/issues/699 and
https://github.com/mirage/mirage/issues/775).

> I do agree we should avoid magic extra dependencies in the `mirage` tool, and to push the dependencies down the chain to nearer where they're needed. I propose that a simple improvement would be to
> 
> - remove this magic from the `mirage` tool.

what does "this" mean?

> - replace the `io-page` dependency in [mirage-xen.opam](https://github.com/mirage/mirage-platform/blob/84431dd1adba84338598d71edf6bcb7cecdb040c/mirage-xen.opam#L18) with `io-page-xen` (and the _oasis file)
> - replace the `io-page` dependency in [mirage-unix.opam](https://github.com/mirage/mirage-platform/blob/84431dd1adba84338598d71edf6bcb7cecdb040c/mirage-unix.opam#L17) with `io-page-unix` (and the _oasis file)


I've no issue with io-page being split into 3 opam packages (although I
doubt this is a good idea in the long run), but I dislike superfluous
dependencies introduced for no obvious reason (initially there was no
io-page dependency for each `Project`).

hannesm:
On 15/06/2017 21:58, Anil Madhavapeddy wrote:
> I am in favour of removing the depopts to start with and making the appropriate changes in mirage-xen and mirage-unix as proposed. I can't actually remember why it's not a hard dependency like this to begin with, in the relevant mirage-<os> packages...

AFAICS io-page should only ever be needed in fringe device drivers:
mirage-net-xen, mirage-block-xen, ...  I'm not sure whether tap devices
require page-aligned memory.

There shouldn't be any need for crunch, mirage-block-unix, etc. to
require io-page at all.  And the backend-specific packages need to be
registered in mirage, since this is the only place knowing about the
backend in use.  But please not at the very top level, this will
introduce unnecessary dependencies for the smallest possible unikernels.

djs55:
Thanks for the link to  mirage/mirage-platform#188 -- I knew I had seen that somewhere but couldn't find it when I went looking.

(By "remove this magic" -- sorry I was unclear -- I mean the special case adding of `io-page` in the `mirage` tool)

Regarding removing the `io-page.unix` dependency from `mirage-unix` (and the same from `mirage-xen`?): I agree that this a more accurate description of the true requirements of the `mirage-unix` package. I'll try to do that. I'll probably uncover missing dependencies in other packages, which is no bad thing, but it will require some more minor updates. A similar thing happened when I removed spurious linkage against `cstruct.ppx` -- various other libraries broke because they had failed to specify that they also depended on it.

In summary my revised proposal is:
- remove the special-case adding of `io-page` in the `mirage` tool
- remove `io-page` from `mirage-platform` (as far as possible)
- test other packages (particularly the fringe device drivers) and add in `io-page-unix` or `io-page-xen` dependencies where they are missing (and previously only worked because of dependencies on `mirage-unix` and `mirage-xen`)

Once I've got through this I would like to look at handling the stubs properly.

hannesm:
thx for looking into this!  as noted in 188, mirage-xen actually uses
io_page (not sure whether the use can be removed entirely)

avsm:
> There shouldn't be any need for crunch, mirage-block-unix, etc. to
require io-page at all.  

Agreed (except for mirage-block-unix, which could use `O_DIRECT` for a performance boost, which in turn requires block aligned writes)

hannesm:
mirage-xen uses `Io_page.t` only for `xenstore_start_page` and `console_start_page` (as an output argument), which is then immediately converted to a `Cstruct.t` and passed to `Xenstore_ring.Ring.init` -- I suspect this code can easily be rewritten to not use `Io_page.t` at all (and there is no information lost which is useful here AFAICS -- the fact that these pages are page-aligned doesn't seem to be relevant)

PR #835, opened 2017-06-14
Link io-page-unix on Ukvm and Virtio too

djs55:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+### 3.0.3 (2017-06-14)
+* add a --block configure flag for picking ramdisk or file-backed disk
+* add lower bounds on packages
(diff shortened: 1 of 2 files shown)


hannesm:
uhm, I'm troubled by this PR.  If `io-page-unix` is needed for solo5, it is not `unix` anymore.  Does it contain any C stubs (which won't work).  Could you please elaborate on why adding this new dependency is needed, and why the suffix `unix` was chosen?  I'd appreciate to wait a bit more than 12 hours for review and feedback.

djs55:
Good catch -- I agree this shouldn't be necessary. I think I must have misinterpreted a ukvm link failure while trying to fix the linking following the `io-page` package split and the `get_addr` binding update. It certainly builds fine without this patch in my clean build environment.

I'll revert this-- sorry for the noise.

PR #836, opened 2017-06-15
Remove io-page-unix from Ukvm and Virtio

djs55:
Neither of these backends should need or can use the Unix stub.

I built these using
```
opam init git://github.com/djs55/opam-repository#io-page
opam pin mirage .
git clone git://github.com/mirage/mirage-skeleton
cd mirage-skeleton
make clean && make MODE=xen
make clean && make MODE=unix
make clean && make MODE=virtio
make clean && make MODE=ukvm
```
(I notice the CI only does `xen` and `unix`)
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 * add a --block configure flag for picking ramdisk or file-backed disk
 * add lower bounds on packages
 * fallback to system `$PKG_CONFIG_PATH`
-* update for mirage-qubes-ipv4, io-page-unix, io-page-xen
+* update for mirage-qubes-ipv4
 
 ### 3.0.2 (2017-03-15)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-### 3.0.3 (2017-06-14)
+### 3.0.4 (2017-06-15)
 * add a --block configure flag for picking ramdisk or file-backed disk
 * add lower bounds on packages
 * fallback to system `$PKG_CONFIG_PATH`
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by avsm** [comment]
FreeBSD 11 virtio manual test works for me, with

```
mirage.3.0.3                                          git   git://github.com/djs55/mirage#revert-dodgy-patch     
mirage-unikernel-https-ukvm.~unknown   (uninstalled)  path  /root/mirage-skeleton/applications/static_website_tls
```

and https://github.com/ocaml/opam-repository/pull/9516 in my opam-repository

djs55:
- Linux `device-usage/block` is fine
- Xen `device-usage/block` is fine
- Xen `device-usage/clock` is fine

The `ukvm` monitor on `alpine-3.6.1` fails with an `mprotect` error: https://github.com/mirage/mirage-solo5/issues/22 -- this looks like a loader failure? I'm investigating to see whether it's an interaction with an alpine kernel hardening feature.

hannesm:
please consider [my other comment](https://github.com/mirage/mirage/pull/833#issuecomment-308810467) to move the io-page dependencies out of project, but instead into the io_page_conf device

djs55:
OK as discussed on the other issue, I'll remove the `io-page` dependency from here and `mirage-platform` and add it in other places where it's necessary (suspect: `mirage-block-xen` and `mirage-net-xen`)

djs55:
After further investigation it seems that released `solo5` doesn't work on Alpine 3.6.1, probably because of kernel hardening features. I've reported https://github.com/Solo5/solo5/issues/202

It does work on a more vanilla Ubuntu 16.04 LTS kernel-- I'll use this for local testing.

PR #838, opened 2017-06-16
mirage: adjust name (argv_qubes is gone, this is argv_xen)

hannesm:


djs55:
Probably worth a rebase against master to see if the CI goes green.

djs55:
Actually I can probably force push a rebase myself easily enough...

djs55:
OK, I've force pushed a rebase. Let's see if the CI is working (in general)

lib/mirage.ml
```diff
@@ -1246,7 +1246,7 @@ let no_argv = impl @@ object
 let argv_xen = impl @@ object
     inherit base_configurable
     method ty = Functoria_app.argv
-    method name = "argv_qubes"
+    method name = "argv_xen"
```
samoht: does it need a change in the package constraints too?

hannesm: it's only the name, not used for any package dependencies..

PR #840, opened 2017-06-18
Travis: Run apt update before calling apt install

g2p:


.travis-ci.sh
```diff
@@ -1,4 +1,5 @@
 eval `opam config env`
+sudo apt update
```
samoht: normally the `depext -u` below should run `apt update`:

```
$ opam depext --help

    -u, --update
           Update the OS package sets before installation

```
Do you have some logs where this is broken?

g2p: Oh, good point. Here's my log: https://travis-ci.org/mirage/mirage/jobs/244219628

g2p: Looks like it should be added to mirage-skeleton depext calls then.

g2p: Oh wait, it has to be added to a `make depend` target that's generated by mirage itself.

g2p: See #841. It's a more intrusive change but probably the right thing to do.

g2p:
Running apt update is currently broken due to the Stretch release (testing is now an alias for buster and not stretch, and that breaks things).
Probably something that needs to be fixed at the Docker image level.

avsm:
I'm taking a look at the Docker base images.

samoht:
Closing as the other PR is now in.

PR #842, opened 2017-06-19
doc updates: remove odig pin, trim package list, pin cstruct

avsm:
pinning cstruct ensures that we can cull packages that are
lagging behind the dependency cone. This is just tlstunnel
for the moment.


avsm:
merging this before CI to trigger a docs update

PR #843, opened 2017-06-20
docs: remove notty as it needs to support a new uuseg release

avsm:


dbuenzli:
Note it doesn't need a new `uuseg` release, it needs to support the new `Uchar.t` see https://github.com/pqwy/notty/pull/11

PR #847, opened 2017-06-27
Adding --gdb argument for ukvm

ricarkol:
ukvm has a `gdb` module that starts a GDB server so debuggers can attach to it. Linking the `gdb` module is defined in `Makefile.ukvm` which is generated by mirage for every `ukvm` unikernel. This change adds a mirage `--gdb` flag that defines whether to add the `gdb` module to `Makefile.ukvm`. Additionally, this change also moves the generation of `Makefile.ukvm` to the mirage configure phase.

Here is the proposed flag, for the ukvm target:
```
$ mirage configure --gdb -t ukvm
$
```
and this for all others (just a warning):
```
$ mirage configure --gdb -t unix
mirage: [WARNING] --gdb not supported on this target.
$
```
You might have to pin the lastest solo5.

@mato

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1681,6 +1681,9 @@ let configure i =
   let target = Key.(get ctx target) in
   Log.info (fun m -> m "Configuring for target: %a" Key.pp_target target);
   let opam_name = unikernel_name target name in
+  let gdb = Key.(get ctx gdb) in
+  if gdb && target <> `Ukvm then
+    Log.warn (fun m -> m "--gdb not supported on this target.");
   configure_myocamlbuild () >>= fun () ->
   configure_opam ~name:opam_name i >>= fun () ->
   configure_makefile ~opam_name >>= fun () ->
@@ -1797,7 +1800,7 @@ let ldflags pkg = pkg_config pkg ["--variable=ldflags"]
 
 let ldpostflags pkg = pkg_config pkg ["--variable=ldpostflags"]
 
-let link info name target =
+let link info name target gdb =
   let libs = Info.libraries info in
   match target with
   | `Unix | `MacOSX ->
@@ -1861,7 +1864,7 @@ let link info name target =
         | "mirage-net-solo5" -> "net" :: acc
         | "mirage-block-solo5" -> "blk" :: acc
         | _ -> acc)
-        [] libs
+        [] libs @ (if gdb then ["gdb"] else [])
     in
     pkg_config "solo5-kernel-ukvm" ["--variable=libdir"] >>= function
     | [ libdir ] ->
@@ -1878,9 +1881,10 @@ let build i =
   let warn_error = Key.(get ctx warn_error) in
   let target = Key.(get ctx target) in
   let libs = Info.libraries i in
+  let gdb = Key.(get ctx gdb) in
   check_entropy libs >>= fun () ->
   compile libs warn_error target >>= fun () ->
-  link i name target >>| fun out ->
+  link i name target gdb >>| fun out ->
   Log.info (fun m -> m "Build succeeded: %s" out)
 
 let clean i =
@@ -1927,6 +1931,7 @@ module Project = struct
       method keys = [
         Key.(abstract target);
         Key.(abstract warn_error);
+        Key.(abstract gdb);
       ]
       method packages =
         let common = [
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1681,9 +1681,9 @@ let configure i =
   let target = Key.(get ctx target) in
   Log.info (fun m -> m "Configuring for target: %a" Key.pp_target target);
   let opam_name = unikernel_name target name in
-  let gdb = Key.(get ctx gdb) in
-  if gdb && target <> `Ukvm then
-    Log.warn (fun m -> m "--gdb not supported on this target.");
+  let target_debug = Key.(get ctx target_debug) in
+  if target_debug && target <> `Ukvm then
+    Log.warn (fun m -> m "-g not supported on this target.");
   configure_myocamlbuild () >>= fun () ->
   configure_opam ~name:opam_name i >>= fun () ->
   configure_makefile ~opam_name >>= fun () ->
@@ -1800,7 +1800,7 @@ let ldflags pkg = pkg_config pkg ["--variable=ldflags"]
 
 let ldpostflags pkg = pkg_config pkg ["--variable=ldpostflags"]
 
-let link info name target gdb =
+let link info name target target_debug =
   let libs = Info.libraries info in
   match target with
   | `Unix | `MacOSX ->
@@ -1864,7 +1864,7 @@ let link info name target gdb =
         | "mirage-net-solo5" -> "net" :: acc
         | "mirage-block-solo5" -> "blk" :: acc
         | _ -> acc)
-        [] libs @ (if gdb then ["gdb"] else [])
+        [] libs @ (if target_debug then ["gdb"] else [])
     in
     pkg_config "solo5-kernel-ukvm" ["--variable=libdir"] >>= function
     | [ libdir ] ->
@@ -1881,10 +1881,10 @@ let build i =
   let warn_error = Key.(get ctx warn_error) in
   let target = Key.(get ctx target) in
   let libs = Info.libraries i in
-  let gdb = Key.(get ctx gdb) in
+  let target_debug = Key.(get ctx target_debug) in
   check_entropy libs >>= fun () ->
   compile libs warn_error target >>= fun () ->
-  link i name target gdb >>| fun out ->
+  link i name target target_debug >>| fun out ->
   Log.info (fun m -> m "Build succeeded: %s" out)
 
 let clean i =
@@ -1931,7 +1931,7 @@ module Project = struct
       method keys = [
         Key.(abstract target);
         Key.(abstract warn_error);
-        Key.(abstract gdb);
+        Key.(abstract target_debug);
       ]
       method packages =
         let common = [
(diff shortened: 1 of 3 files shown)


lib/mirage.ml
```diff
@@ -1680,7 +1715,14 @@ let configure i =
...
   let target = Key.(get ctx target) in
   Log.info (fun m -> m "Configuring for target: %a" Key.pp_target target);
+  let gdb = Key.(get ctx gdb) in
   let opam_name = unikernel_name target name in
+  let libs = Info.libraries i in
+  let _ = match target with
+    | `Ukvm -> R.ok ()
+    | _ -> (if gdb then Log.warn (fun m -> m "--gdb not supported on this target."));
+      R.ok ()
+  in
```
hannesm: this is slightly overcomplicated, we can use the following instead:
```OCaml
if gdb && target <> `Ukvm then
  Log.warn ... ;
```
(i.e. no need for the `match`, also not to `R.ok ()` in both branches).

hannesm:
LGTM, from what I've seen over in the solo5 repository, the gdb stuff only works on ukvm if using kvm+linux as backend, is that correct?

lib/mirage.ml
ricarkol: Changed. Thanks.

ricarkol:
Thanks for the review @hannesm . Changed the code according to your suggestion.

"LGTM, from what I've seen over in the solo5 repository, the gdb stuff only works on ukvm if using kvm+linux as backend, is that correct?"
--> Yes, at the moment just kvm+linux is supported. Will take a look at freebsd next.

hannesm:
I looked through the CI failures, 3 of which happened in `MODE=ukvm`.  The root of the problem is that with the changes in this PR, `pkg-config solo5-kernel-ukvm --libdir` is executed at `mirage configure` time, which is before `make depext` was executed to install the missing dependencies.  Is there any reason to do this at configuration time instead of (as it was previously) at `mirage build` time?

ricarkol:
Hi @hannesm . There were two reasons (2. was the real reason):
1. pkg-config and ukvm-configure "felt" like they belonged in the configure phase: they are not building anything (plus they have "config" in their names).
2. It seems that CLI arguments passed to `mirage configure` should be passed to `mirage build` through the `.mirage.config` file, but for some reason it didn't seem to work for flag type args or bools. For example, `warn_error` is not respected in `mirage build`; it always takes the default value. I wasn't sure if it was a bug or I missed something, so I just left it like that for now.

hannesm:
thanks for your elaboration @ricarkol,

1 - there seems to be a lack of documentation (sorry, I'll see where this should be added), but since Mirage3 the `configure` phase (of the `mirage` tool) does not require any backend packages.  This is appreciated for several reasons, among them is cleaner CI builds, and having dependencies being installed via an opam file.  In my opinion the separation between `mirage configure` (requiring only the `mirage` package itself) and `mirage build` (requiring all dependencies for the configured backend etc.) is sensible, and should be kept.  In mirage, it is really about the required dependencies, not so much about the (in Unix) popular `configure`, `make`, and `make install` steps.

2 - this is definitively a bug that boolean options are not preserved from the `configure` to the `build` phase.  I'll look into that.

hannesm:
Indeed, argument passing between `mirage configure` and `mirage build` is completely broken - it seems to work only for the `target` argument.  Additionally, some `` `Configure `` phase parameters are emitted into the `main.ml`.  `mirage describe` (with or without `--eval`) doesn't report anything sensible either.  So far I haven't been able to find the issue, maybe others (@Drup @samoht) have insights whether this ever worked or how this is supposed to work?

hannesm:
I reported the issue with keys to the functoria issue tracker https://github.com/mirage/functoria/issues/116

ricarkol:
Thanks @hannesm . Will fix this change so it doesn't rely on the configure phase.

mato:
Some meta-comments on the usability / future implications of this change:

1. Should the argument be called something more concrete than just `--gdb`? To strawman some syntax, perhaps `--ukvm:gdb`?
2. We may want to think about a more generic mechanism for "target specific" configuration options.

ricarkol:
Thanks @mato , changed the arg to `--ukvm:gdb`.

About point 2. Definitely agree; that's why I was thinking of a generic `--gdb` arg (that unfortunately happens to be implemented by a single target). Although xen or qemu (virtio) could also start a gdb server.

ricarkol:
Question please. I can't build the latest mirage (with or without this change) with the latest functoria (pinning the latest master). I'm getting:
```
# File "lib/mirage.ml", line 342, characters 24-33:
# Error: Unbound value Info.root
# Command exited with code 2.
```
But if something was wrong with functoria (+latest mirage), the CI would definitely catch it. So, I'm doing something wrong. But can't figure it out. I created a new ocaml+opam container, updated, upgraded, pinned functoria and mirage (and getting this compilation error). Am I missing something?

samoht:
Not yet, I was planning to work on this today. I'll keep you updated!

ricarkol:
Thanks @hannesm and @samoht

samoht:
an update: mirage's master is now working fine with functoria's master, where the issue with persistence of build keys should now be resolved.

samoht:
Regarding the feature itself, ideally we should make the acceptation of the `--gdb` key be dependant on the `--target` value. I don't think it is possible to do that today (/cc @Drup who knows more) as that would be a bit difficult to show that dependency properly in the man page.

About the `--ukvm:gdb`: I am not very fond of that convention, but why not (and this can be changed later on if needed anyway). Would be great to double-check what we do something similar with other target dependant keys, but I don't think we had such needs before.

An other option would be to pick up a more generic debug key (for instance `-g`) and let each target backend do whatever it wants with it. I think I would tend to like that option more.

mato:
> An other option would be to pick up a more generic debug key (for instance `-g`) and let each target backend do whatever it wants with it. I think I would tend to like that option more.

I like this option.

hannesm:
@ricarkol if you rebase onto master, the CI will (hopefully) pass.  And then use `--gdb` / `-g` instead of `ukvm_gdb` (as suggested by @mato @samoht), and it should be good to merge! :)

ricarkol:
@hannesm rebased (waiting for the tests)

@mato @samoht changed the arg to:
```
       -g, --gdb
           Start a gdb server (only available for ukvm).
```
The issue is that can't find an easy way of making the help message dependent on the `--target` (as you mention above). So, that's ^^ the message for all targets.

ricarkol:
Thanks @hannesm. Made the `s/ukvm_gdb/gdb/` change.

hannesm:
any objections against merging this?

lib/mirage_key.ml
```diff
@@ -156,6 +156,12 @@ let warn_error =
   let key = Arg.(opt ~stage:`Configure bool false doc) in
   Key.create "warn_error" key
 
+let gdb =
+  let doc = "Start a gdb server (only available for ukvm)." in
```
mato: This ("Start a gdb server") is not actually what the option does. Per the comments on the PR, we decided that `-g` or `--gdb` should be a generic option to "enable (unspecified) debugger support for this target".  I'd suggest rewording this to something like:

"Enables target-specific support for debugging. Supported targets: ukvm (compiles ukvm-bin with GDB server support)."

Also, given the change to a more generic "enable debugging" description for this option, should we call it only `-g` and drop all the references to GDB? This would include renaming the key internally to something more generic (target_debug?)

ricarkol: Hi @mato , actually, what about having the `gdb` module in ukvm start the server by default? We could still have an option to not start it.

mato:
> Hi @mato , actually, what about having the `gdb` module in ukvm start the server by default? We could still have an option to not start it.

I don't have any huge opinion on that, other than some people may want to
compile up a ukvm with the gdb module and only start the gdb server
optionally, so for me things are consistent the way they are now.

However, it doesn't change my comments on the description and intent of
the `mirage` option -- `mirage` is a build tool, what happens at run time
is up to the user and out of scope of that tool.

ricarkol:
@mato : replaced the docs with the suggested text, and replaced `gdb` with `target_debug`.

mato:
@ricarkol The updated change looks OK to me. Thanks!

However: Attempting to `opam pin add mirage .` with either your `ukvm-gdb` branch as-is or rebased onto current mirage `master` gives:

````
opam pin add mirage .
mirage is now path-pinned to /home/mato/solo5/mirage

[mirage] /home/mato/solo5/mirage/ synchronized
[mirage] Installing new package description from /home/mato/solo5/mirage

mirage needs to be reinstalled.
Sorry, no solution found: there seems to be a problem with your request.

No solution found, exiting
[NOTE] Pinning command successful, but your installed packages may be out of
       sync.
````

Can anyone who knows what the current state of mirage `master` is comment? @samoht?

hannesm:
@mato mirage master depends on functoria master atm AFAIU

mato:
@hannesm Pinning functoria to master still gives me the same OPAM error when I try to pin mirage to master, with no indication of how to fix it.

hannesm:
@mato mirage on master depends on a (non-existing) functoria 2.2.0.  You need to `opam pin add functoria.dev --dev-repo` (or `functoria.2.2.0`) to convince opam all is good.

mato:
Ok, tested and works for me, but a minor UX nit: Given that we still have a default target (`unix`), a plain `mirage configure -g` gives `mirage: [WARNING] -g not supported on this target.` which is somewhat non-intuitive. @ricarkol Can you change the warning to "-g not supported for target: <target>", i.e. print the target name?

ricarkol:
Thanks @mato. Made the change.

mato:
Thanks @ricarkol, this LGTM.

hannesm:
squash&merge -- the travis CI failure are related to other changes, see the thread on mirageos-devel https://lists.xenproject.org/archives/html/mirageos-devel/2017-07/msg00009.html

PR #848, opened 2017-07-13
fix reference to tar library for `--kv_ro archive`

mor1:


lib/mirage.ml
```diff
@@ -443,7 +443,9 @@ let archive_conf = impl @@ object
     method name = "archive"
     method module_name = "Tar_mirage.Make_KV_RO"
     method packages =
-      Key.pure [ package ~min:"0.7.0" ~ocamlfind:["tar.mirage"] "tar-format" ]
+      Key.pure [
+        package ~min:"0.7.0" ~ocamlfind:["tar-format.mirage"] "tar-format"
+      ]
```
hannesm: FWIW to me it looks like the `tar` package/ocamlfind got renamed yet another time... is that correct @djs55?  I suspect the current package needed for `tar` and MirageOS is `tar-mirage` (both ocamlfind and opam name), which itself depends on `tar`.

djs55: That's correct -- from version 0.8.0 `tar-format` -> `tar` and `tar-format.mirage` -> `tar-mirage`. The old packages should still work but you might as well update.

mor1: Ok-- I'll update the PR to version constraint `min:"0.8.0"` and to use `tar`/`tar-mirage` appropriately then. Thanks!

djs55:
Thanks!

PR #849, opened 2017-07-18
Adapt to latest functoria

samoht:
/cc @ricarkol and @hannesm

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -339,7 +339,7 @@ let direct_kv_ro_conf dirname = impl @@ object
     method module_name = "Kvro_fs_unix"
     method packages = Key.pure [ package ~min:"1.3.0" "mirage-fs-unix" ]
     method connect i _modname _names =
-      let path = Fpath.(Info.root i / dirname) in
+      let path = Fpath.(Info.build_dir i / dirname) in
       Fmt.strf "Kvro_fs_unix.connect \"%a\"" Fpath.pp path
   end
 
@@ -400,7 +400,7 @@ class block_conf file =
 
     method connect i s _ =
       Fmt.strf "%s.connect %S" s
-        (self#connect_name (get_target i) @@ Info.root i)
+        (self#connect_name (get_target i) @@ Info.build_dir i)
   end
 
 let block_of_file file = impl (new block_conf file)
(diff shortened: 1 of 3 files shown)


lib/mirage.ml
```diff
@@ -1912,6 +1912,8 @@ module Project = struct
     "open Lwt.Infix\n\
      let return = Lwt.return\n\
      let run = OS.Main.run"
+  let packages = [package "mirage"]
```
hannesm: I'm curious about this one... what kind of dependency are you expressing here?  that `Project` (as in, every MirageOS unikernel) depends on `mirage` at run time?  how comes that this is necessary?

samoht: That's the dependency to compile `config.ml` (e.g. to build the object to dynlink to get the configuration)

hannesm: should this then be a `build:true` dependency?  and I'm not sure what the benefit of adding it here is (apart from being exact, which I appreciate :)) -- if you run `mirage configure` or `mirage build`, you will have to have `mirage` installed!?

samoht: Previously, functoria were hard-coding that `Project.name` would be the name of the library to link with when building config.ml. This is now a bit more abstract:

https://github.com/mirage/functoria/blob/master/app/functoria_app.ml#L648-L650

This was done primary to make testing of functoria easier (so you can be a bit more explicit when building the tests: https://github.com/mirage/functoria/blob/master/tests/lib/test_app.ml#L60) but I think that's a bit saner to do this instead of relying on `Project.name` being the right ocamlfind name.

samoht: Ha I think I understand the confusion now. This is just used for ocamlfind names, but I thought re-using the already existing `package` type would be useful (for instance to allow configuration plugins later on). Do you think I should just have a list of strings there?

hannesm:
On 18/07/2017 12:28, Thomas Gazagnaire wrote:
> Ha I think I understand the confusion now. This is just used for
> ocamlfind names, but I thought re-using the already existing
> `package` type would be useful (for instance to allow configuration
> plugins later on). Do you think I should just have a list of strings
> there?

maybe it is worth to put a comment there?  I guess reusing the package
type is fine, if documented..

hannes

samoht:
All green, the only failure issue should be fixed by https://github.com/mirleft/ocaml-nocrypto/pull/124

samoht:
I've added some comments to explain what the new fields are for, and fixed all the CI scripts (and opened https://github.com/mirleft/ocaml-nocrypto/pull/124 to fix a issue when compiling the pnrg example in mirage-skeleton for the unix backend). All is green now, so could be merged.

PR #839, opened 2017-06-17
[wip] port to jbuilder

avsm:
not for merge yet until more documented, but with this PR multidirectory builds of mirage-types works with jbuilder. You can do:

```
for i in block channel clock console device flow fs kv net protocols random stack time; do
  opam source mirage-$i --dev
done
git clone -b jbuilder git://github.com/avsm/mirage
jbuilder build @mirage/types/lwt/doc
jbuilder build mirage/types/lwt/mirage_types_lwt.cmx
```

opam-free local development!  and then multiple environments also work...

```
$ cat jbuilder-workspace
(context ((switch 4.03.0)))
(context ((switch 4.04.1)))
```

make sure both the switches have dependencies that we havent locally git cloned available:

```
OPAMSWITCH=4.03.0 opam install -y -j4 lwt cstruct ipaddr uri io-page
OPAMSWITCH=4.04.1 opam install -y -j4 lwt cstruct ipaddr uri io-page
```

then a build will build all the local files in both switches, in separate build dirs

```
$ jbuilder build mirage/types/lwt/mirage_types_lwt.cmx
<snip>
    ocamlopt mirage-channel/lwt/mirage_channel_lwt.{cmx,o} [4.04.1]
    ocamlopt mirage/types/mirage_types.{cmx,o} [4.04.1]
    ocamlopt mirage-block/lwt/mirage_block_iter.{cmx,o} [4.04.1]
    ocamlopt mirage-block/lwt/mirage_block_lwt.{cmx,o} [4.03.0]
    ocamlopt mirage-block/lwt/mirage_block_lwt.{cmx,o} [4.04.1]
    ocamlopt mirage/types/lwt/mirage_types_lwt.{cmx,o} [4.03.0]
    ocamlopt mirage/types/lwt/mirage_types_lwt.{cmx,o} [4.04.1]
```

PR #850, opened 2017-07-18
Switch to jbuilder

samoht:
superseeds #839 /cc @avsm


hannesm:
I tried to raise the question elsewhere, but didn't get any answer, so trying here again:
You rename `mirage_types.mli` to `mirage_types.ml`.  How does this interact with documentation generation?  Are the documentation comments of the `.ml` used when generating documentation?
Is there a difference (`diff -wur`!?) between the old way of generating documentation and the new way (and if so, is the diff ok or is there something missing)?

samoht:
> How does this interact with documentation generation?

That doesn't change anything, `odoc` works as well as before -- the generated `index.html` page is identical.

hannesm:
On 20/07/2017 11:26, Thomas Gazagnaire wrote:
> That doesn't change anything, `odoc` works as well as before -- the generated `index.html` page is identical.

Ok, so a `diff -urw` between old and new docs is empty?

samoht:
> Ok, so a `diff -urw` between old and new docs is empty?

Yes, at least for `mirage_types.ml{i}`. I can test on something a bit more complex if you like.

samoht:
Same for `mirage_types_lwt.ml{i}`

hannesm:
@samoht thanks for investigating!

types/jbuild
```diff
@@ -0,0 +1,21 @@
...
+   (public_name mirage-types)
+   (modules     (mirage_types))
+   (libraries   (mirage-device mirage-time mirage-random mirage-flow
+                 mirage-console mirage-block mirage-clock mirage-net
+                 mirage-kv mirage-fs mirage-channel
+                 mirage-protocols mirage-stack))))
+
+(library
+  ((name        mirage_types_lwt)
+   (public_name mirage-types-lwt)
```
rgrinberg: Looks a bit fishy, shouldn't this have `(wrapped false)`? (Same for `mirage_types`).

samoht: In practice that doesn't matter as there is only one module with that name, but it's probably saner to add `(wrapped false)`. Thanks for noticing :-)

PR #851, opened 2017-07-24
ARP is compatible with MirageOS3 since 0.2.0 (the latest release)

hannesm:


samoht:
LGTM (if the tests pass)

hannesm:
On 24/07/2017 13:31, Thomas Gazagnaire wrote:
> LGTM (if the tests pass)

it'd surprise me if there is any test which is affected by this change ;)

PR #853, opened 2017-07-30
syslog: allow runtime configuration of log host

hannesm:
previously, no keys were defined for configuring syslog, but all had to be done
in config.ml.  now, the config argument to syslog_udp/tcp/tls is optional, and
three keys are available:
--syslog=IPv4
--syslog-port=int
--syslog-hostname=string

the specific protocol to use (UDP/TCP/TLS) has to be provided at configuration
time (which is sensible because the specific library has to be linked into the
unikernel)

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -593,7 +593,7 @@ let ethernet_conf = object
   method ty = network @-> ethernet
   method name = "ethif"
   method module_name = "Ethif.Make"
-  method! packages = Key.pure [ package ~min:"3.1.0" ~sublibs:["ethif"] "tcpip" ]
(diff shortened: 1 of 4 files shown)


hannesm:
while this is a breaking change (introducing a labeled `?config`), I suspect the user base using the syslog code is not too big.  these changes ease the setup of some unikernels with different IP parameters much more straightforward (esp. if the same unikernel is be run in different IP contexts with different syslog servers, such as authoritative name servers).

hannesm:
this PR is ready for review!  Travis is happy with it.

the last commit should be reverted once there's a migration plan to tcpip.3.2.0, conduit.1.0.0, mirage-conduit.3.0.0 -- I included it here to give travis a chance to compile all mirage-skeleton examples.

hannesm:
thanks @samoht -- I manually merged the first commits into master!

PR #855, opened 2017-08-01
Add upper bounds to all packages specified in mirage config keys

avsm:
This just adds an upper bound to the next major revision of the
relevant library.  This will make us slightly more future proof
against opam package changes, but will require more frequent
releases of this repository to relax upper bounds.  The only other
major source of incompatibilities coming up will be package
renaming (e.g. mirage-conduit to conduit-mirage), but that is
easier as it is a new package and will not require constraints.

This PR also fixes the conduit-mirage package and mirage/mirage-skeleton#232

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
  global:
    - PRE_INSTALL_HOOK="cd /home/opam/opam-repository && git pull origin master && opam update -u -y"
    - POST_INSTALL_HOOK="sh ./.travis-ci.sh"
-   - PINS="functoria.dev:--dev functoria-runtime.dev:--dev mirage.dev:. mirage-types.dev:. mirage-types-lwt.dev:. mirage-runtime.dev:."
+   - PINS="functoria.dev:--dev functoria-runtime.dev:--dev mirage.3.1.0:. mirage-types.3.1.0:. mirage-types-lwt.3.1.0:. mirage-runtime.3.1.0:."
    - TESTS=false #testing via travis-ci.sh
  matrix:
    - PACKAGE=mirage DISTRO=debian-testing OCAML_VERSION=4.04.2 EXTRA_ENV="MODE=xen"
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -55,7 +55,7 @@ let qrexec_qubes = impl @@ object
   val name = Name.ocamlify @@ "qrexec_"
   method name = name
   method module_name = "Qubes.RExec"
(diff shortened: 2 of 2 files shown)


avsm:
A couple of other necessary things:

- does io-page.unix need to become io-page-unix and also io-page-xen similarly? cc @djs55 
- need to add `tcpip.unix` and `tcpip.xen` for the right checksum stubs

.travis.yml
```diff
@@ -9,7 +9,7 @@ env:
  global:
    - PRE_INSTALL_HOOK="cd /home/opam/opam-repository && git pull origin master && opam update -u -y"
    - POST_INSTALL_HOOK="sh ./.travis-ci.sh"
-   - PINS="functoria.dev:--dev functoria-runtime.dev:--dev mirage.dev:. mirage-types.dev:. mirage-types-lwt.dev:. mirage-runtime.dev:."
+   - PINS="functoria.dev:--dev functoria-runtime.dev:--dev mirage.3.1.0:. mirage-types.3.1.0:. mirage-types-lwt.3.1.0:. mirage-runtime.3.1.0:."
```
hannesm: I suspect it should then be `3.dev` here, otherwise we've to change it every other release!?

djs55:
@avsm I think the minimum change would be to replace `io-page.unix` with `io-page-unix` (minimum version `2.0.0`). (There is a backwards compatibility package `io-page.unix` but it would be good not to rely on it). I don't think `io-page.xen` or `io-page-xen` are known to the `mirage` tool -- `io-page-xen` is brought in as a dependency of `mirage-xen`.

However we might be able to remove the `io_page_conf` from the `mirage` tool altogether -- it's not used as a functor argument or considered a "device" anywhere these days -- it's assumed to be a regular library dependency. There's a reference to it in `crunch` but I don't know if that's still required -- `crunch` code that runs on Unix which needs aligned pages should have an explicit dependency itself.

.travis.yml
avsm: We just need to change it after every release. There's no point putting an upper bound on if it's dev, as that is effectively not an upper bound.  The idea is that the upper bound in the `opam` file is 4.0, so our pinned version here needs to be less than that. We could make travis pin to mirage.3.99...

avsm:
Thanks, I'm in favour of removing io_page -- it's an odd device to expose. Will do shortly.

avsm:
I'll do the io-page fix in a separate PR, as it's a different improvement really.

hannesm:
thanks, superseeded by #940 and #944 (the upper bounds will need some adjustment, but will be part of 3.3.0).

PR #856, opened 2017-08-03
build fixes for --dhcp=true, networking with -t unix, and conduit

yomimono:
This won't work without a bunch of pins until the latest charrua packages are released (see https://github.com/mirage/charrua-core/pull/70 for the preparation PR).  If you want to try it out, pin `charrua-core`, `charrua-client`, `charrua-client-lwt`, and `charrua-client-mirage` to that branch.

Lots of stuff is still broken but this is a start.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -700,7 +700,7 @@ let dhcp_conf = impl @@ object
     method ty = time @-> network @-> dhcp
     method name = "dhcp_client"
     method module_name = "Dhcp_client_mirage.Make"
-    method! packages = Key.pure [ package ~sublibs:["mirage"] "charrua-client" ]
+    method! packages = Key.pure [ package "charrua-client-mirage" ]
     method! connect _ modname = function
       | [ _time; network ] -> Fmt.strf "%s.connect %s " modname network
       | _ -> failwith (connect_err "dhcp" 2)
@@ -711,7 +711,7 @@ let ipv4_dhcp_conf = impl @@ object
     method ty = dhcp @-> ethernet @-> arpv4 @-> ipv4
     method name = Name.create "dhcp_ipv4" ~prefix:"dhcp_ipv4"
     method module_name = "Dhcp_ipv4.Make"
-    method! packages = Key.pure [ package ~sublibs:["mirage"] "charrua-client" ]
+    method! packages = Key.pure [ package "charrua-client-mirage" ]
     method! connect _ modname = function
       | [ dhcp ; ethernet ; arp ] ->
         Fmt.strf "%s.connect@[@ %s@ %s@ %s@]" modname dhcp ethernet arp
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1001,8 +1001,7 @@ let tcp_conduit_connector = impl @@ object
     method module_name = "Conduit_mirage.With_tcp"
     method! packages =
       Key.pure [
-        package ~min:"2.3.0" ~ocamlfind:[] "mirage-conduit";
-        package ~min:"0.15.0" ~sublibs:["mirage"] "conduit"
+        package ~min:"3.0.1" "mirage-conduit";
       ]
     method! connect _ modname = function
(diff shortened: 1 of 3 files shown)


lib/mirage.ml
```diff
@@ -669,12 +669,18 @@ let opt_map f = function Some x -> Some (f x) | None -> None
 let (@?) x l = match x with Some s -> s :: l | None -> l
 let (@??) x y = opt_map Key.abstract x @? y
 
+(* convenience function for linking tcpip.unix or .xen for checksums *)
+let right_tcpip_library ?min ?max ?ocamlfind ~sublibs pkg =
+  Key.(if_ is_unix)
+    [ package ?min ?max ?ocamlfind ~sublibs:("unix"::sublibs) pkg ]
+    [ package ?min ?max ?ocamlfind ~sublibs:("xen"::sublibs) pkg ]
```
hannesm: this is unlikely to be correct on `virtio` and `ukvm` targets.

yomimono: Right you are after rereading @djs55's explanation of how this linking works; xen doesn't mean "everything but unix" everywhere.

hannesm: yes, it's slightly awkward how different `backends` do this (atm -- actually since several years).  the simple approach to contain (most!) bindings in a single package (as done by solo5) seems to be more convenient atm, but a real cross-compilation story is needed...

hannesm:
would be great to have one unikernel in skeleton which we pass `dhcp=true` to avoid regression...

lib/mirage.ml
yomimono: There's more discussion of this over in https://github.com/mirage/mirage-tcpip/issues/324 FWIW.

avsm:
@yomimono once this passes CI i can merge with #855

yomimono:
Will @ you when it looks ready, hopefully yet today

yomimono:
The DHCP server test still doesn't pass on Unix, because there we manually note a dependency on `tcpip` in the `packages` argument to `register` in `config.ml`, so `mirage` doesn't get a chance to express its opinion on what additional sublibraries are required.  The actual user of the higher-level `tcpip` stuff is `charrua-core`, which doesn't know which backend we need.

djs55:
Looking at the previous version of `tcpip` we used to reference the Unix checksum stubs from the main `tcpip` ocamlfind package which worked on Xen, ukvm because the `mirage` tool performs a custom link in those cases and ignores the Unix stubs. I think the simplest fix would be to replicate this -- by adding a dependency from `tcpip` to `tcpip.unix`? I'll give this a go and see if it works.

djs55:
FWIW with the latest `tcpip` `master` branch this seems to work for me. @yomimono does it work for you?

yomimono:
I'm AFK for the next several hours but the passing Travis tests are heartening.

djs55:
Charrua has been released; CI is green :) and I think this looks good to me!

yomimono:
Thanks @djs55 !  :)

hannesm:
thanks for fixing this!

But I'm confused: in https://github.com/mirage/mirage-tcpip/pull/335 you introduced a dependency of `tcpip.unix` to `tcpip`, and now in here another set of dependencies to `tcpip.unix` are generated... are both needed?  what is the behaviour on non-unix platforms (is `tcpip.unix` linked and ignored since it is overwritten by symbols from `tcpip.xen` / `solo5`)?  it feels wrong to me that `tcpip` depends on `tcpip.unix`.

also, since `charrua` is merged to opam, various `pin` in `travis.yml` should've been removed here...

djs55:
A PR for removing PINS: https://github.com/mirage/mirage/pull/858

It's probably not necessary to add both sets of dependencies, but it should be safe. In Unix the OCaml toolchain does the linking and will pull in the Unix stubs . The `mirage` tool manually links the non-Unix binaries itself and ignores these dependencies and has its own separate mechanism (`xen_link_opts` and `mirage-solo5`). It's very confusing.

ciarancourtney:
Not sure if this is related, but I'm having reproducible issues with networking on latest release https://github.com/mirage/mirage-skeleton/issues/234

PR #858, opened 2017-08-08
travis: remove PINS on charrua and functoria

djs55:
charrua.0.9 and functoria.2.2.0 have been released.


djs55:
@yomimono thanks for the patch -- sorry for the clashing PRs!

djs55:
Travis is green with the PINS removed, this is good!

Issue #796, opened 2017-03-02
hint for pinning lwt to get profiling no longer works


yomimono:
We used to get a nice hint when building unikernels that wanted tracing in environments where `lwt` wasn't pinned to the tracing branch.  It seems to have gone away:

```4.04.0🐫  ((detached from origin/master)) mirageos:~/mirage-skeleton/device-usage/tracing$ mirage configure -t xen
4.04.0🐫  ((detached from origin/master)) mirageos:~/mirage-skeleton/device-usage/tracing$ make depend
opam pin add --no-action --yes mirage-unikernel-example-xen .
Package mirage-unikernel-example-xen does not exist, create as a NEW package ? [Y/n] y
mirage-unikernel-example-xen is now path-pinned to /home/user/mirage-skeleton/device-usage/tracing

[mirage-unikernel-example-xen] /home/user/mirage-skeleton/device-usage/tracing/ synchronized
[mirage-unikernel-example-xen] Installing new package description from /home/user/mirage-skeleton/device-usage/tracing

opam depext --yes mirage-unikernel-example-xen
# Detecting depexts using flags: x86_64 linux debian
# The following system packages are needed:
#  - debianutils
#  - libxen-dev
#  - m4
#  - ncurses-dev
#  - perl
#  - pkg-config
#  - time
#  - uuid-dev
# All required OS packages found.
opam install --yes --deps-only mirage-unikernel-example-xen

=-=- Synchronising pinned packages =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[mirage-unikernel-example-xen] /home/user/mirage-skeleton/device-usage/tracing/ already up-to-date
opam pin remove --no-action mirage-unikernel-example-xen
mirage-unikernel-example-xen is now unpinned from path /home/user/mirage-skeleton/device-usage/tracing
4.04.0🐫  ((detached from origin/master)) mirageos:~/mirage-skeleton/device-usage/tracing$ make
mirage build
Uncaught exception: Invalid_argument("Fl_package_base.query")
[...]

PR #868, opened 2017-11-17
fix lwt tracing hint (& an `ocamlfind query` invocation bug)

yomimono:
Fixes https://github.com/mirage/mirage/issues/796 .


hannesm:
well spotted, LGTM! (and sorry for breaking it in the first place)

PR #870, opened 2017-11-23
cohttp-mirage is released as 1.0.0

hannesm:
and not as 3.0.0.  the most recent mirage (3.0.6) won't work :(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,7 @@ env:
  global:
    - PRE_INSTALL_HOOK="cd /home/opam/opam-repository && git pull origin master && opam update -u -y"
    - POST_INSTALL_HOOK="sh ./.travis-ci.sh"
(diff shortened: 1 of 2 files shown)


hannesm:
found via https://travis-ci.org/mirage/mirage-skeleton/jobs/306405459

hannesm:
the alternative would be to release cohttp-mirage as 3.0.0, but I fear that will be rather brittle if the rest of cohttp version numbers is 1.0.0 (and thus releases of cohttp would need some custom magic).

this certainly needs another mirage (3.0.7) release really soon now, since 3.0.6 is broken (in case you try to use the http stuff).

hannesm:
Travis CI is passing, it would be great to get this + a fresh mirage released to unbreak the current universe with mirage + http.  @djs55 @yomimono

djs55:
LGTM!

djs55:
A trivial `CHANGES.md` update is in #871

PR #872, opened 2017-11-24
opam: use https://github.com/mirage/mirage as the homepage

djs55:
This fixes the output of `make opam-pkg`, since `topkg` assumes the
download URL is based on the `homepage`.

See for example [ocaml/opam-repository#10843]

Note that mirage-types-lwt already used github.com/mirage/mirage
(and so was inconsistent with the rest). Now all are consistent.


hannesm:
lgtm

PR #874, opened 2017-12-04
allow access to xen block devices other than the first

yomimono:
The logic for the `Xen` and `Qubes` cases is redundant with logic in `mirage-block-xen`'s connect module: see https://github.com/mirage/mirage-block-xen/blob/master/lib/blkfront.ml#L488 .

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -391,12 +391,9 @@ class block_conf file =
       | `Unix | `MacOSX | `Virtio | `Ukvm ->
         Fpath.(to_string (root / b.filename)) (* open the file directly *)
       | `Xen | `Qubes ->
-        (* We need the xenstore id *)
-        (* Taken from https://github.com/mirage/mirage-block-xen/blob/
-           a64d152586c7ebc1d23c5adaa4ddd440b45a3a83/lib/device_number.ml#L64 *)
-        (if b. number < 16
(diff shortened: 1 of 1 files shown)


cfcs:
Thank you so much!

djs55:
+1 on removing the knowledge about Xenstore encoding of Linux `/dev/xvd` device names -- no-one should have to know about that scheme any more :)

Just checking I understand (I'm a bit rusty): the `xl create foo.cfg` write a backend key `params=<filename>` for each disk and then we give `<filename>` to the frontend? So we use `<filename>` as a unique id?

yomimono:
@djs55 Yeah, I'm a little rusty on this too :( but now is a great time to get familiar with it again, since there's so much exciting storage stuff going on.

For the `-t xen` case, `mirage configure` makes the `foo.xl` for unikernel `foo`, and there's some additional transformation done on the given strings before making the `.xl` file - see https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L1559 for the exact logic.

yomimono:
@djs55 can you merge, comment, or close (or tag someone else in), please?  I'd like to cut a point release with this bugfix.

djs55:
Sure, I think this is fine.

yomimono:
Thanks, @djs55 ! :)

PR #879, opened 2018-01-04
allow directly specifying xenstore ids for block devices

### LINKED ISSUE #878: version 3.0.8 broke block device initializiation in Xen
The mirage-skeleton block example reproduces this.

```
2017-12-28 18:57:39 -00:00: ERR [blkfront] Blkfront.connect disk.img: unable to match 'disk.img' to any available devices [ 51728 ]
```

The filename set by the user is given to `Blkfront.connect` instead of the [Xen disk IDs generated in the `.xl` file](https://github.com/mirage/mirage/blob/master/lib/mirage.ml#L1556).  Before #874, this filename would be automatically translated to the first, second, etc Xen block devices.

yomimono:
This patch reverts #874 to fix #878 .  It also tries to solve the original problem #874 addressed by giving the user handles to directly specify a XenStore ID for block devices.  This can be called directly from `config.ml` with `block_of_xenstore_id` or invoked via the configuration language with `generic_block` in `config.ml` and `--block=xenstore`.  Attempting to use `--block=xenstore` with a Unix-y or Solo5-y target will result in a configure-time error.

/cc @cfcs .

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -372,6 +372,38 @@ let make_block_t =
       end in
     b
 
+let xen_block_packages =
+  [ package ~min:"1.5.0" ~sublibs:["front"] "mirage-block-xen" ]
+
+(* this class takes a string rather than an int as `id` to allow the user to
+   pass stuff like "/dev/xvdi1", which mirage-block-xen also understands *)
+class xenstore_conf id =
+  let name = Name.create id ~prefix:"block" in
+  object
+    inherit base_configurable
+    method ty = block
+    method name = name
+    method module_name = "Block"
+    method! packages = Key.pure xen_block_packages
+    method! configure i =
+      match get_target i with
+      | `Qubes | `Xen -> R.ok ()
+      | _ -> R.error_msg "XenStore IDs are only valid ways of specifying block \
+                          devices when the target is Xen or Qubes."
+    method! connect _ s _ =
+      Fmt.strf "%s.connect %S" s id
+  end
+
+let block_of_xenstore_id id = impl (new xenstore_conf id)
+
+(* calculate the XenStore ID for the nth available block device.
+   Taken from https://github.com/mirage/mirage-block-xen/blob/
+   a64d152586c7ebc1d23c5adaa4ddd440b45a3a83/lib/device_number.ml#L64 . *)
+let xenstore_id_of_index number =
+        (if number < 16
+         then (202 lsl 8) lor (number lsl 4)
+         else (1 lsl 28)  lor (number lsl 8))
+
 class block_conf file =
   let b = make_block_t file in
   let name = Name.create file ~prefix:"block" in
@@ -382,7 +414,7 @@ class block_conf file =
     method module_name = "Block"
     method! packages =
       Key.match_ Key.(value target) @@ function
-      | `Xen | `Qubes -> [ package ~min:"1.5.0" ~sublibs:["front"] "mirage-block-xen" ]
+      | `Xen | `Qubes -> xen_block_packages
       | `Virtio | `Ukvm -> [ package ~min:"0.2.1" "mirage-block-solo5" ]
       | `Unix | `MacOSX -> [ package ~min:"2.5.0" "mirage-block-unix" ]
 
@@ -391,12 +423,7 @@ class block_conf file =
       | `Unix | `MacOSX | `Virtio | `Ukvm ->
         Fpath.(to_string (root / b.filename)) (* open the file directly *)
       | `Xen | `Qubes ->
-        (* We need the xenstore id *)
-        (* Taken from https://github.com/mirage/mirage-block-xen/blob/
-           a64d152586c7ebc1d23c5adaa4ddd440b45a3a83/lib/device_number.ml#L64 *)
-        (if b. number < 16
-         then (202 lsl 8) lor (b.number lsl 4)
-         else (1 lsl 28)  lor (b.number lsl 8)) |> string_of_int
+        xenstore_id_of_index b.number |> string_of_int
 
     method! connect i s _ =
       Fmt.strf "%s.connect %S" s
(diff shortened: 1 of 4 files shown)


hannesm:
so @cfcs tested it and came up with a fix for him, please have a look into https://github.com/cfcs/mirage/commit/938bfab19f1233f94dc4f26c586296036102df89 (//cc @djs55 @talex5 @yomimono who may know more about mirage, xen, and block devices)

talex5:
Does @cfcs just need to use the new `block_of_xenstore_id` added by this PR?

I'm not really qualified to approve this PR, but it looks reasonable. Having a separate API for passing Xen IDs vs file paths is certainly the right thing to do.

( I suspect the underlying problem is with `Blkfront.connect` (https://github.com/mirage/mirage-block-xen/blob/a6f482239e4a91c8ff378e60500f02002886ecff/lib/blkfront.mli#L17), which just takes a `string`. Ideally we would deprecate this and add multiple replacements for the different types (or one which takes a variant). Then this code in the mirage tool would probably become more understandable. )

cfcs:
The new patch uses the `number` as the `id`, but the `number` is just an sequentially incremented counter used elsewhere to keep track of *how many* block devices are used, not the *id* of the device in Xen terms.
@yomimono first patch made it possible to use `xvdd` (the Linux naming convention where `xvd` is the driver and `d` is id `3`, starting at `a`/`0`), that was nice.
My current patch means I have to specify `block_of_file "3"` in my `config.ml` which is arguably not very pretty either.

I would really like if the `xvdd`-parsing was restored. I considered adding that to the patch, but it felt like that was already implemented elsewhere and that it would probably be prettier to just use that code somehow. What do you think?

talex5:
@cfcs I don't think I understand what you're trying to do. Why can't you use `block_of_xenstore_id "xvdd"`? Or are you trying to make an actual local file available?

cfcs:
@talex5 because I would like it to be portable :) `block_of_file "xvdd"` ignores the `xvdd` and the number `1` is used instead, which means I end up with `xvdb` no matter what string I put in there.

talex5:
How can "xvdd" be portable to other platforms? It's a Xen disk identifier, isn't it?

Either you have a file that you want to share (in which case it shouldn't matter what Xen device number you end up using), or you want to access a particular Xen device (in which case your code will require an `if_` to say what to do on other platforms).

cfcs:
@talex5 In the context of `block_of_file` it's a filename, presumably pointing to something in the user's working directory (that of either the compiling user or the invoking user), so having `xvdd` is portable whereever those characters are allowed in a file name -  I'm not so religious about what the final name will be; all I want is to be able to use some persistent storage cross-platform from Mirage. ;-)

I think `xvdd` is the Linux driver `xvd`'s 4th (0-indexed) device, so there are also `xvda`(`0`), `xvdb`(`1`) and so on - I don't think there's anything inherently Xen about it; we could also have chosen to use the FreeBSD naming convention for the devices exposed by Xen; it's just a convenience function.

Xen seems to use some kind of insane number system internally, for unknown reasons, which is what the conversion functions in the code here converts to.

So, some people might say that using `block_of_file` with a string that happens to coincide with the name of a Xen block device would be a bit hacky, especially when the FreeBSD/Solo5 people want to join the fun and suddenly find themselves dealing with the Linux names for the Xen block devices.
One could argue that if we don't want people to do this, but we intend for them to use `if_`, we should probably not try to automagically convert their filenames from `block_of_file` to Xen device ids, but rather force them to call `block_of_xenstore_id` or similar directly (ideally using the type system to make sure that `block_of_file` can only be called on platforms where a `file` is a thing, and likewise it doesn't make sense to use `block_of_xenstore_id` on platforms without Xen).

Speaking of `if_`: I have no idea how to use this, and I haven't stumbled upon any examples. Does anyone have a best practices link to a `config.ml` that uses persistent storage?
In general I'm having a hard time finding examples of best practice for parametric compilation, so I've been experimenting with a few different approaches, but none of this feels particularly clean:
- https://github.com/cfcs/PongOS/blob/master/config.ml#L26-L61
- https://github.com/cfcs/mirage-framebuffer-audio/blob/master/config.ml#L6-L45

talex5:
From the interface, it looks like `generic_block` is another option. That might be better if you just want some block device but don't care too much whether it's a local file or a Xen disk.

My understanding is that we can use (dom0) files as virtual disks in Xen (I haven't used this though, so I might be wrong). So `block_of_file "xvdd"` should create a `config.xl` that maps some auto-generated Xen device ID to this file, and passes the device ID to the unikernel. We need to make sure we don't confuse this operation with telling the unikernel to use Xen device "xvdd" (set manually by the user).

Possibly `block_of_file` doesn't make sense under Qubes, or maybe the automatic numbering needs to change.

yomimono:
> My understanding is that we can use (dom0) files as virtual disks in Xen (I haven't used this though, so I might be wrong). So block_of_file "xvdd" should create a config.xl that maps some auto-generated Xen device ID to this file, and passes the device ID to the unikernel. We need to make sure we don't confuse this operation with telling the unikernel to use Xen device "xvdd" (set manually by the user).

This matches my understanding.  I *have* used dom0 files as virtual disks - both files on a local filesystem in dom0, and actual block devices which dom0 is aware of but does not mount and passes directly to the guest.  This at least used to work in Septemberish 2015.

`block_of_file` doesn't make sense (except by accident - you expected two block devices, and the hypervisor happened to give you two for reasons unrelated to your expectation) in any situation where the user doesn't fully control the boot environment - this includes Qubes and public cloud environments.  I talked a bit about this in [the issue](https://github.com/mirage/mirage/issues/878).

I didn't have a non-Qubes Xen platform for testing these things while away and I've only just managed to get one installed; I need to do additional setup work before I can test with it.

@cfcs The portable solution is intended to be `generic_block` (there is a sort-of-convention that `generic_whatever` will try to make an educated guess based on what target the user chose, and allow for configure-time overrides of that).  The need to pass something like "xvdd" as an argument in the source code totally breaks the putative genericity of that approach, though.  I thought there was an issue floating around about this choice in the block device configuration (that the names really need to be on the key/value level, possibly lifted into a variant), but I can't seem to find it at the moment.  I agree that it's unsatisfactory.

yomimono:
I've made an issue referencing the configuration issues discussed here.  Unless there are other suggestions for how to solve this problem in the short term, I will likely merge this PR today and make a release soon.

PR #887, opened 2018-03-03
Add new Solo5-based platform `Muen

Kensan:
This change adds the Muen platform to the mirage tool so it can be chosen as a target for MirageOS unikernels. It has been tested and is actually already in use for several weeks to build and run the Muen project website (see also description [here](https://muen.sk/articles.html)).

This is the remaining puzzle piece to fully upstream Muen support for MirageOS, see also Solo5/solo5#194.

Note: the necessary adaptations on the Muen side have been merged and released as part of [v0.9](https://groups.google.com/forum/#!topic/muen-dev/FPL9sc4yaBE).

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -42,9 +42,9 @@ let connect_err name number =
 
 (* Mirage implementation backing the target. *)
 let backend_predicate = function
-  | `Xen | `Qubes   -> "mirage_xen"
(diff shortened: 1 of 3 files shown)


mato:
This looks good to me, thanks! Merging into the 'dev-solo5' feature branch.

Issue #884, opened 2018-02-02
mirage's autogenerated opam files give error messages (but works) with opam2


yomimono:
Everything works, but opam2's convert-then-lint step gives big red `error` messages noting that `maintainer`, `authors`, and `homepage` are missing, and warns that `bug-reports` is also not present (but then pins the package and installs the dependencies normally, as expected).  We could avoid this by generating these fields, so users don't get irrelevant error output.

PR #888, opened 2018-03-09
Supress lint warnings from generated opam file

mato:
Add maintainer, authors, homepage, bug-reports fields with a content of "dummy" to supress opam lint warnings during "make depend".


yomimono:
This should fix #884 , thanks!

PR #891, opened 2018-03-12
use a path pin when making depends

yomimono:
Otherwise `make depend` when using opam version 2 has some surprising behavior.

/cc @mato

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1697,7 +1697,7 @@ let configure_makefile ~opam_name =
                   all:: build\n\
                   \n\
                   depend depends::\n\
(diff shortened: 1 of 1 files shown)


hannesm:
LGTM

yomimono:
CI fails just like everyone else's builds against the latest cohttp with the latest jbuilder with:

```
File "_none_", line 1:
Error: No implementations provided for the following modules:
         Cohttp__ referenced from unikernel.cmx
```

mato:
Pinning jbuilder to 1.0+beta17 fixes this for me, so the problem has been introduced in beta18,

hannesm:
thanks for this!

PR #897, opened 2018-03-22
mirage: add ocamlbuild and ocamlfind dependencies

hannesm:
mirage calls out to ocamlbuild during mirage build

### DIFF
--- a/mirage.opam
+++ b/mirage.opam
@@ -15,6 +15,8 @@ build: [
 ]
 
 depends: [
+  "ocamlbuild"
+  "ocamlfind"
   "jbuilder"   {build & >= "1.0+beta10"}
   "ipaddr"             {>= "2.6.0"}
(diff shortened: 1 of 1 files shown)


hannesm:
@Drup huh?  if `mirage` calls out to `ocamlfind` and `ocamlbuild` during its runtime (in this specific case during `mirage build`), this should be reflected in its opam, or?  the generated `opam` file is still not self-sufficient -- i.e. only used (optionally) for depext -- the build step is not described in the opam file.

Drup:
`mirage build` will also calls to crunch during its runtime, if the unikernel needs that. Should `mirage` unconditionally depends on `crunch` as well ? The opam file describes all the dependencies necessary for the unikernel to be built, which should include ocamlbuild.

It might be a bit of a pedantic point, new users are going to need to do `opam install --deps .` anyway, so I don't think it matters that much.

PR #898, opened 2018-03-28
respect ld from pkg-config

hannesm:
this is needed again for OpenBSD support in solo5 (since OpenBSD's `ld` is an ancient `GNU ld`, which doesn't work with the solo5 linker script -- but their `ld.lld` (from clang/llvm) does..
//cc @adamsteen @mato

this basically reverts e0aba4dc2dac7f180517b7b129e8e890617a8e15 -- but since `ld` is no longer called via Makefile, but rather from OCaml code, it needs some adjustments.

the failure semantics (if `pkg-config` returns an error) is atm to log (on the warning channel) and call out to `ld`.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1937,8 +1937,6 @@ let link info name target target_debug =
       | [ libdir ] ->
         Bos.OS.Cmd.run Bos.Cmd.(v "ukvm-configure" % (libdir ^ "/src/ukvm") %% of_list ukvm_mods) >>= fun () ->
         Bos.OS.Cmd.run Bos.Cmd.(v "make" % "-f" % "Makefile.ukvm" % "ukvm-bin") >>= fun () ->
-        Log.info (fun m -> m "linking with %a" Bos.Cmd.pp linker);
-        Bos.OS.Cmd.run linker >>= fun () ->
         Ok out
       | _ -> R.error_msg ("pkg-config " ^ pkg ^ " --variable=libdir failed")
     else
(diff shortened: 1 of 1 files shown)


lib/mirage.ml
```diff
@@ -1890,60 +1910,39 @@ let link info name target target_debug =
...
+            | "mirage-block-solo5" -> "blk" :: acc
+            | _ -> acc)
+          [] libs @ (if target_debug then ["gdb"] else [])
+      in
+      pkg_config pkg ["--variable=libdir"] >>= function
+      | [ libdir ] ->
+        Bos.OS.Cmd.run Bos.Cmd.(v "ukvm-configure" % (libdir ^ "/src/ukvm") %% of_list ukvm_mods) >>= fun () ->
+        Bos.OS.Cmd.run Bos.Cmd.(v "make" % "-f" % "Makefile.ukvm" % "ukvm-bin") >>= fun () ->
+        Log.info (fun m -> m "linking with %a" Bos.Cmd.pp linker);
+        Bos.OS.Cmd.run linker >>= fun () ->
```
mato: @hannesm Maybe I'm misreading the diff, but won't this run the linker twice for `ukvm`? The first call is on line 1927 just before the `if target = Ukvm`.

hannesm: well spotted, I pushed 093ba78d which removes the second call to linker

mato:
Tested this manually, LGTM, thanks!

Issue #890, opened 2018-03-12
make configuring block storage nicer


yomimono:
As noted in https://github.com/mirage/mirage/pull/879 , the current functions for configuring block devices are annoyingly non-portable in their string arguments.  In practice the user needs to customize the value passed to `generic_block` in `config.ml`; the interface could also be more typeful.

yomimono:
The issue I mentioned in #879 about block values was actually one about the argument to `generic_kv_ro`, and is at https://github.com/mirage/mirage/issues/791 .

cfcs:
- I would like to have an interface with two steps:
First choosing the storage backend (at compile-time),
and choosing the argument to the storage backend (ie the path to the storage device) at runtime - optionally with a compile-time default.

- Furthermore I'd like the compile-time selection to get a target context so the unikernel can define the default based on the target.
- Also, I think the `Mirage_unix` stub should be able to (at runtime) emulate a block device if given a regular file.
- For small files it would be useful to be able to compile in a default (`crunch`-style) and have it optionally replaced by a runtime argument.

This will accomodate constructions such as:
- *I am a unikernel that reads a configuration file from a block device. The user should tell me which block device to use.*
- *I am a unikernel that reads a configuration file.*
  - *on Unix I default to `$HOME/.config/myunikernel.config`, but the user can override this path with `--config-file`*
  - *on QubesOS I default to the fourth Xen block device (mapped from `modules.img` in dom0)*

I envison the code being something like:
```ocaml
Mirage.config_block
  (function
  | `Unix | `MacOSX -> `Default "~/.config/myunikernel.config"
  | `Solo5 -> `Must_supply_at_runtime
  | `Xen -> `Default "51760"
  )
```

- I don't know if this is too complicated, but optional constraints on size and permissions (read/write) would be nice from a user perspective, so I can complain if a user gives me a 4K byte block device and I know that I need at least 512K of storage.

yomimono:
(also /cc @drup , who can comment more cogently on whether Functoria can support this)

Drup:
As far as functoria/the `Mirage` module/tool goes:

> First choosing the storage backend (at compile-time),

Yes, `generic_kv_ro` works like that.

> and choosing the argument to the storage backend (ie the path to the storage device) at runtime - optionally with a compile-time default.

Yes, `generic_kv_ro` doesn't work like that at the moment, but all the networking things do. Functoria supports that very well.

[...]

PR #903, opened 2018-05-17
Add libvirt.xml for virtio configuration target

bramford:
As per this [discussion](https://discuss.ocaml.org/t/mirageos-orchestration/1989/4), an example libvirt.xml file should be emitted when configuring virtio unikernels (as they are with xen).

I've tested the example output including clock, network and console devices. The only `virtio` feature that isn't demonstrated is `block` devices which don't appear to be supported by mirage for `virtio`.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1507,6 +1507,50 @@ let configure_main_libvirt_xml ~root ~name =
        R.ok ())
     "libvirt.xml"
 
+let configure_virtio_libvirt_xml ~root ~name =
+  let open Codegen in
+  let file = Fpath.(v (name ^  "_libvirt") + "xml") in
+  with_output file
+    (fun oc () ->
+      let fmt = Format.formatter_of_out_channel oc in
+      append fmt "<!-- %s -->" (generated_header ());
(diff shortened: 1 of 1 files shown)


hannesm:
thanks!  I think `block` devices are supported by `solo5/virtio` (@ricarkol didn't you use them for some benchmarks?)?  at least in the `solo5` project there is a `kernel/virtio/virtio_blk.c` file around.  why did you conclude that they are not supported?

bramford:
@hannesm I already knew that `solo5/virtio` supports `block` but found `block_of_file` from `mirage.mli` confusing for the `virtio` case (with libvirt). After further testing, I found that the path to the block file (in the VM) is irrelevant as `solo5/virtio` only supports one device and therefore just grabs the first block device it finds regardless of name.

I've updated the PR to include the block example. Is it ok to have two commits or should I just squash them?

I'll also have a look at what would make the mirage block interface easier for this case. Maybe just some better documentation would be enough.

bramford:
Actually, following on from my previous comment, I've just seen #890 and the comments within. Looks like the interface will change in the near future anyway.

PR #908, opened 2018-06-18
changes for 3.1.0

hannesm:
leftover question when i scrolled through the commits:
what happened with #890?  is the current state in master good for both all use cases, or should we work a bit more on getting the block device story in a better shape? -- esp. asking @yomimono @cfcs

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,17 @@
+### 3.1.0
+
+* solo5 v0.3.0 support (#906, by @mato @kensan @hannesm):
+  The major new user-visible features for the Solo5 backends are:
+    ukvm: Now runs natively on FreeBSD vmm and OpenBSD vmm.
+    ukvm: ARM64 support.
+    muen: New target, for the Muen Separation Kernel.
+    ukvm: Improved and documented support for debugging Solo5-based unikernels.
+* generate libvirt.xml for virtio target (#903, by @bramford)
+* don't make xen config documents for target qubes (#895, by @yomimono)
+* use a path pin when making depends (#891, by @yomimmono)
+* move block registration to `configure` section (#892, by @yomimmono)
+* allow to directly specifying xenstore ids (#879, by @yomimmono)
+
 ### 3.0.8 (2017-12-19)
 
 * when passing block devices to `xen`, pass the raw filename rather than trying to infer the xenstore ID (#874, by @yomimono)


yomimono:
Good question.  I didn't make PRs to implement the stuff in #890.  I think current master is a reversion which works with mirage-www but is not nice for @cfcs's case, but I need to check and make sure this is true.

yomimono:
The relevant fix was #879 and was tested to be OK with the Xen case back when I first proposed it.  `mirage-www` should be buildable and bootable in Xen with the current mirage/mirage master branch's HEAD.

yomimono:
As far as the release goes, I don't think we should delay to implement #890 and rather should release the required changes for the new solo5 off the current HEAD.

cfcs:
I still don't know how to use block devices in Mirage.

yomimono:
@hannes pushed to your branch, hope that's OK.  The PR for the Muen compatibility stuff in mirage-skeleton is at https://github.com/mirage/mirage-skeleton/pull/257 .

@cfcs I think using block devices in Mirage is still kind of broken until we do the stuff in #890.

samoht:
@g2p and I plan to spend to time to add first-class support for wodan in the mirage tool. We'll use this as a good excuse to revisit the block device interface :-) And yes, I don't think this should block the solo5 release.

cfcs:
TL;DR: Don't hold back this PR for my sake. :-)

Yes, I think `mirage-www` proves that it's possible, but so far I haven't found a way to select which block device to use, which makes it kind of useless if you have more than one.
I just gave up and reverted to compiling things into the unikernel instead with `ocaml-crunch`, which is also kind of sad since it seems like I need to [chunk the files](https://github.com/cfcs/mirage-framebuffer-audio/tree/master/tracks) to ensure they are less than `~16 MB` to avoid running into the undocumented *endless-busyloop* feature that seems to hide somewhere in there, but maybe I'm also using that wrong.
TBH that got too depressing to work with, so I tried to write some non-Mirage applications using `Unix` to access storage, but that also turns out to be slow and broken unless you're dealing with very small files, so now I just try to focus on pure code, where the only thing to ruin days are well-understood concepts like the unchecked+undocumented exceptions, absence of immutable arrays/vectors, and Undefined Behavior from the underlying C implementation lying in wait to fuck up any integer arithmetic operation you perform.

PR #912, opened 2018-07-18
Use the exposed signature in functoria for Key modules.

Drup:
This changes nothing. In particular, this doesn't fix the failure in 4.07 since the current definition in functoria is exactly the same.
However, it delegates the definition to functoria which means we can fix everything there.

### DIFF
--- a/lib/mirage_key.ml
+++ b/lib/mirage_key.ml
@@ -346,6 +346,4 @@ let logs =
   let arg = Key.Arg.(opt logs []) info in
   Key.create "logs" arg
 
-(* FIXME: this is a crazy *)
-include (Key: module type of struct include Functoria_key end
-         with module Arg := Arg and module Alias := Alias)
+include (Key: Functoria.KEY with module Arg := Arg and module Alias := Alias)


hannesm:
thanks, this looks good to me.  IIUC to get mirage working on OCaml 4.07.0, we need this change and https://github.com/mirage/functoria/pull/153 (i.e. all earlier releases of mirage should be bound to OCaml < 4.07.0)!?

Drup:
Yes. both previous version of mirage and functoria need to be bound. The new version should compile on all versions.

Drup:
@hannesm If you have successfully tested/read the patch, I invite you to press the big green button. :)

hannesm:
thanks, indeed.  just to be sure, IIUC there's no need to introduce a version dependency between mirage and functoria with this change, or is there?  (meaning old versions of functoria play well with mirage with this PR merged, and the other way around, old mirage versions will play along nicely with new functoria versions)

Drup:
I'm pretty sure this signature has been defined from the beginning, so yes, no need for constraints.

PR #914, opened 2018-07-28
prepare for release

hannesm:
we need #912 for 4.07.0 support, thus I've updated CHANGES and the opam files to reflect what we have in opam-repository atm.. please let me know if you'd like to postpone a mirage release (in case there should be something included which is not yet atm)!?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 ### master
 
+* bump minimum mirage-solo5* and solo5-kernel* to 0.3.0 (#914, by @hannesm, as
+  suggested by @mato)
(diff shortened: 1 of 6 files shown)


mato:
This was mentioned before, but forgotten about: If we're doing a point release, it might be worth adding minimum version constraints (>= 0.3.0) in lib/mirage.ml for the mirage-*solo5 packages generated into the opam file. This won't hurt anything, and will prevent old cruft from being installed accidentally.

hannesm:
@mato done in e5c0ae2

mato:
@hannesm This looks good to me, thanks!

PR #916, opened 2018-08-10
Squash unnecessary warning from "mirage build"

mato:
"using ld as ld" is just informational, users can still get it with
--verbose.


hannesm:
LGTM, remaining CI failures:
```
[ERROR] No solution for mirage-unikernel-noop-xen: The following dependencies couldn't be met:
          - mirage-unikernel-noop-xen -> mirage-xen >= 3.0.4 -> mirage-xen-ocaml >= 2.6.0 -> ocaml < 4.06.0
              base of this switch (use `--unlock-base' to force)
```
for 4.05.0 and 4.04.2 in travis.yml using
```
DISTRO=ubuntu-16.04 EXTRA_ENV="MODE=xen"
```

seem to provide wrong OCaml versions!?

samoht:
Interestingly, the docker scripts are not switching to the right compiler version. This should be fixed by https://github.com/ocaml/ocaml-ci-scripts/pull/239

samoht:
All green with the CI fixes, merging

PR #920, opened 2018-09-01
3.2.0 release: solo5 0.4.0 renamings, revised mirage-stack initialisation, changes

hannesm:
EDIT: this is the mirage 3.2.0 release PR ;)
- discussion in https://github.com/mirage/mirage/issues/923
- solo5 related changes discussed in https://github.com/mirage/mirage/pull/924
- mirage-stack, -protocols related discussion in https://github.com/mirage/mirage-stack/pull/11


mato:
This seems to have the changes from #924 in it, is that deliberate? (Did not see the EDITs, as no notifications show up for those)

hannesm:
@mato yes, this is deliberate -- I wanted the CI to test the full changeset for 3.2.0 -- which it did successfully.  imho good to merge and release.

mato:
@hannesm Fine with me, though the things missing in https://github.com/mirage/mirage/pull/924#issuecomment-421137022 would be nice to have, but unfortunately I don't have time for that in the next days and don't want to hold up the release for it.

hannesm:
@mato thanks. are there any other opinions on the 3.2.0 release? /cc @mirage/core 

in your mentioned #924, there are three items:
- "better error than 'invalid cached configuration' after mirage upgrade in a dirty tree"
- `mirage configure -t ukvm` doesn't hint to use `-t hvt`
- `mirage clean` doesn't remove all `ukvm` artifacts (what's left: `<foo>.ukvm` and `mirage-unikernel-<foo>-ukvm.opam`)

the first and second item: we could re-add `-t ukvm` to lead to the `` `Hvt `` target - not sure if it's worth it, since at some point we may want to discard it.
third item: I pushed 6fafc307, would this be a good solution (this now removes all the generated opam files)?

hannesm:
> ukvm thingies

I tried to achieve this, and failed.

> I'd prefer not introducing a hard dependency on OPAM 2 just yet.

I'm not sure I can follow the argumentation.  Since September 18th, according to https://opam.ocaml.org/blog/opam-2-0-0/, the opam-repository switched to 2.0 in its master branch.  The 1.2 branch "will only be updated for critical fixes". If I understand this correctly, releases of new opam packages will only be visible to opam 2.0 users, which means that opam 1.2 users won't be able to see mirage 3.2.0.

mato:
> The 1.2 branch "will only be updated for critical fixes". If I understand this correctly, releases of new opam packages will only be visible to opam 2.0 users, which means that opam 1.2 users won't be able to see mirage 3.2.0.

Ah, I missed that, have read the whole announcement now. If that's the case, then the OPAM 2-related changes are fine with me.

hannesm:
since there's not much feedback, and some discussion about opam2, i decided this is not urgent enough to wait for the 3.2.0 release.  we should discuss in marrakesh about the future of mirage in respect to opam2 (and other ocaml ecosystem changes).

Issue #923, opened 2018-09-13
[tracking] Mirage 3.2.0 release


mato:
This issue is to track the steps towards a Mirage 3.2.0 release.

## Solo5 changes

As of Solo5 0.4.0 (to be released in the next few days), the Solo5 OPAM packages and `ukvm` target have been renamed to `hvt`. The following needs to be done for a release:

1. [ ] ocaml/opam-repository#12612 needs to land _and_ show up in the public repositories (@mato).
2. [ ] Solo5 0.4.0 changes are completed and an 0.4.0 release is published. (@mato, Solo5/Solo5#274)
3. [ ] Downstream MirageOS dependencies get new 0.4.0 releases (`ocaml-freestanding`, `mirage-solo5`). These changes are straightforward and I have them ready on WIP branches. (@mato)
4. [ ] The required `mirage` frontend changes are reviewed and merged. (WIP: https://github.com/mato/mirage/tree/wip-renaming, @mato + reviewers)
5. [ ] Sundry updates, such as `.travis.yml` and/or other CI scripts for `mirage/mirage-skeleton`.
6. [ ] MirageOS documentation needs to be updated.

@mirage/core I'd appreciate it if someone could volunteer to work on steps 4 through 6 above.

## Other changes

@hannesm Please edit and add, these are mainly the `tcpip` API changes?

mato:
Notes on upgrading an existing Mirage/Solo5 switch to one based on Mirage 3.2.0 (to be added to `CHANGES.md` and/or part of the release announcement:

----

## Upgrading from Mirage 3.1.x or earlier

Due to conflicting packages, `opam` will not upgrade `mirage` to version _3.2.0_ or newer if a version of `mirage-solo5` older than _0.4.0_ is installed in the switch. To perform the upgrade you must run:

    opam upgrade mirage

explicitly. After the upgrade has completed, you may want to verify that the correct `solo5-bindings-*` package was installed. (_Unclear if to refer to this, I just get solo5-kernel-ukvm replaced with solo5-bindings-hvt. `make depend` should do the right thing anyway._)

## Changes required to rebuild and run `ukvm` unikernels

As of Solo5 0.4.0, the `ukvm` target has been renamed to `hvt`. If you are working out of an existing, **dirty**, source tree, you should initially run:

    mirage configure -t hvt
    mirage clean
    mirage configure -t hvt

and then proceed as normal. If you are working with a **clean** source tree, then simply configuring with the new `hvt` target is sufficient:

[...]

PR #924, opened 2018-09-13
Renaming for Solo5 0.4.0 release

mato:
Solo5 0.4.0 will rename the "ukvm" target to "hvt", and the Solo5 OPAM packages to "solo5-bindings-TARGET". Adapt the mirage front-end tool to match the new names.

Note that some references to "ukvm" build artifacts are kept to allow a "mirage clean" to succeed after reconfiguring a dirty "ukvm" unikernel source tree with the "hvt" target.

CI is not expected to pass on this PR until the intermediate steps in #923 have been completed. Posting now for early review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,11 @@ env:
    - TESTS=false #testing via travis-ci.sh
  matrix:
    - PACKAGE=mirage DISTRO=debian-testing OCAML_VERSION=4.04 EXTRA_ENV="MODE=xen"
-   - PACKAGE=mirage DISTRO=debian-testing OCAML_VERSION=4.04 EXTRA_ENV="MODE=ukvm"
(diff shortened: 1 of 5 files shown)


mato:
Some things missing from this so far:

1. I think it'd still be nice to get a better error than `Invalid cached configuration, please re-run mirage configure` when doing a `mirage X` in a dirty tree after upgrading. Not sure if that is possible.
2. `mirage configure -t ukvm` just produces a list of allowed values but no hint as to the change from `ukvm` to `hvt`. If possible we should display a more useful message here.
3. (Minor) `mirage clean` in an upgraded dirty unikernel tree does not remove _all_ `ukvm` artifacts, but otherwise works.

hannesm:
merged as part of #920, thanks!

PR #927, opened 2018-10-04
Convert to dune files

emillon:
Mirage, now with fewer parentheses!


dune-project
```diff
@@ -0,0 +1 @@
+(lang dune 1.1)
```
samoht: missing `(name mirage)`

emillon: thanks, just fixed this.

PR #928, opened 2018-10-05
WIP: extract and test a Mirage_cli module

emillon:
Hello!

While trying to understand how this package works, I noticed that the `Mirage` module does a lot of things. I'd like to make it a bit more modular, in particular to test some internals.

This PR adds some characterization tests - just a way to dump the existing behavior so that we know about how these work and we can touch them without fear of breaking things. When/if this becomes possible, they can be replaced by lower level tests.

I have some questions:
- Is a global `Mirage_cli` module OK, or should it be a bit more hidden (like `Mirage.Private_cli` or under `Mirage.Codegen`?)
- At the moment it only deals with generating files, so maybe `Mirage_codegen` is more appropriate?

Thanks!

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -1575,32 +1575,10 @@ let configure_main_xe ~root ~name =
 let clean_main_xe ~name = Bos.OS.File.delete Fpath.(v name + "xe")
 
 let configure_makefile ~opam_name =
-  let open Codegen in
   let file = Fpath.(v "Makefile") in
   with_output file (fun oc () ->
       let fmt = Format.formatter_of_out_channel oc in
-      append fmt "# %s" (generated_header ());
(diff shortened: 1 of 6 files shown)


samoht:
As all your functions are named `output_*` maybe `Mirage_ouput` is indeed better. And you can drop the function prefix then.

Drup:
I'm not convinced those are a great target for lifting them out but why not. You could make the code far less horrifying by just using `{|....|}`. I do appreciate you added all these tests though!

samoht:
I'm closing this PR as it started to bit-rot. I like the general approach of having a better way to test the generated code, but as #933 has been merged, I think we should move all of these functions as top-level of their modules to ease testing. Will follow-up with new PRs.

PR #929, opened 2018-10-05
WIP: Use ppx for ocaml codegen

emillon:
Hi,

This is more of an experiment before I go further on this.

I noticed that some OCaml code is generated and that is uses string interpolation.
I tried something different: define this code using metaquot, and pretty print using `Pprintast`.

While this does not ensure that the code compiles, it's a bit better since we at least ensure that it is a valid OCaml expression. On the other hand, this pulls `ppx_tools_versioned` (and so OMP and compiler-libs), so this can have library size implications (but if this is just for the CLI it might be OK).

What do you think? Is this a direction I should try to follow?

Note that this branch includes #928, so only the last commits are relevant.

cc @hannesm @samoht 

Thanks in advance for your input!

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -1,4 +1,6 @@
 (library
   (public_name mirage)
   (wrapped     false)
-  (libraries ipaddr logs astring functoria functoria.app mirage-runtime bos))
+  (libraries ipaddr logs astring functoria functoria.app mirage-runtime bos ocaml-migrate-parsetree)
+  (preprocess (pps ppx_tools_versioned.metaquot_404))
+)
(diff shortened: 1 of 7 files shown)


Drup:
You are moving some definitions away from their devices into another module. I'm not particularly fond of that. I think I would prefer to try to have self-contained devices as much as possible.

I'm surprised you didn't started that work by changing functoria. Most of the subtle code emission starts there.

emillon:
Right, I'll first start by making private per-device modules. This will be easier to reexport in `Mirage` and test locally.

samoht:
It's a good idea to simplify code generation, however I'm not particularly fond of introducing a ppx dependency in mirage before the ppx tooling is ready. Happy to revive once this is ready.

PR #930, opened 2018-10-06
WIP: Add Mirage_job to wrap Bos actions

emillon:
This adds a module that wraps some `Bos` actions to add a "dry run" mode. The goal is to expose some of these actions and test them at a high level. (this is not intended to implement a `--dry-run` flag)

The implementation is based on the free monad pattern used in `OpamProcess.Job` and explained in [this article]. A list of files is passed to the "dry run" evaluator, which is used to keep track of a static approximation of the file system. This is used for example to determine which value is returned by `exists`.

The `(>>=)` from `Rresult.R` is replaced by a new one in `Mirage_job`, which means that the conversion is pretty straightforward (that code was monadic already).

As for other similar PRs I have opened, I have some questions as to how to make this not impact the API too much:

- where should the jobs be defined? At the moment I just exposed `configure` behind a `(**/**)`, but this does not feel too great.
- would it make sense to move this to a separate library? Maybe after all similar code has been converted?

Let me know what you think. Thanks!

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -30,10 +30,8 @@ include Functoria
 
 let get_target i = Key.(get (Info.context i) target)
 
-let with_output ?mode f k err =
-  match Bos.OS.File.with_oc ?mode f k () with
-  | Ok b -> b
-  | Error _ -> R.error_msg ("couldn't open output channel for " ^ err)
+let with_output ?mode path k purpose =
+  Mirage_job.with_out ~mode ~path ~purpose k
 
 let connect_err name number =
   Fmt.strf "The %s connect expects exactly %d argument%s"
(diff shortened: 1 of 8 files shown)


emillon:
(cc @samoht)

Drup:
This is nice. I would actually put this directly in Functoria instead of Mirage. Maybe just ask for a job in the configure/build/clean actions ? You can then remove some of the ceremony and extend the API to cover all the required bits. It also has the advantage of having more structure for those actions.

For `Mirage.configure`, I'm not sure we really have a choice. Maybe add some documentation that this function is exposed for testing purposes.

emillon:
I'll try that. Most of the jobs are defined in functoria objects, so it might make sense to have a `build_job` and `build` that by default calls `run self#build_job`. If we go that way this means that we'll have to expose the actual `configurable` rather than the `impl` (in a separate module like `Mirage_crunch` etc maybe), but that seems better to me.

Drup:
I was rather thinking of simply changing the type of `build` (and others) directly, instead of having two methods. It breaks compat, but it's probably worth it.

samoht:
Superseeded by #1077

emillon:
Thanks for taking care of this @samoht !

PR #931, opened 2018-10-09
unix bootvar

hannesm:
the semantics between the unix target and other targets in respect to bootvars is different (see https://github.com/mirage/mirage-solo5/issues/33 and https://github.com/Solo5/solo5/issues/281).  This PR fixes this by using the same code path (esp. the `parse-argv` package) on all targets (by introducing a new mirage-bootvar-unix package).


hannesm:
rebased, and imho good to be merged

mato:
Quoting logic/proposals should be done in Parse-argv, there is a PR with some proposed changes in mirage/parse-argv#6. Not enough bandwidth right now to comment on the semantics.

PR #932, opened 2018-10-09
revise stdlibrandom.initialize signature

hannesm:
which now lives inside of Lwt.t, see https://github.com/mirage/mirage-random/pull/7


hannesm:
see #934

PR #933, opened 2018-10-09
Split mirage.ml into private impl modules

emillon:
This creates several modules named `Mirage_impl_*`. The first part of `mirage.ml`, that deals with implementations, can be expressed as reexport of these modules.

There are two goals to that refactoring:
- `mirage.ml` becomes smaller
- details not relevant to `mirage.mli` can be exported there and tested. This will benefit to #928, #929, #930.

`mirage.mli` is not changed, so the public API is not broken.

Does that look like a good architecture? If that sounds OK I'll convert the other parts (ethernet, ARP, ip, etc).

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -23,21 +23,10 @@ module Key = Mirage_key
 module Name = Functoria_app.Name
 module Codegen = Functoria_app.Codegen
 
-let src = Logs.Src.create "mirage" ~doc:"mirage cli tool"
-module Log = (val Logs.src_log src : Logs.LOG)
+module Log = Mirage_impl_misc.Log
+open Mirage_impl_misc
 
 include Functoria
-
-let get_target i = Key.(get (Info.context i) target)
-
-let with_output ?mode f k err =
-  match Bos.OS.File.with_oc ?mode f k () with
-  | Ok b -> b
-  | Error _ -> R.error_msg ("couldn't open output channel for " ^ err)
-
-let connect_err name number =
-  Fmt.strf "The %s connect expects exactly %d argument%s"
-    name number (if number = 1 then "" else "s")
 
 (** {2 OCamlfind predicates} *)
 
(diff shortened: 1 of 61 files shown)


emillon:
I converted the different parts and reexported them in `Mirage`. A nice followup to this would be to cleanup the new modules, e.g. renaming each type to `t`, etc. Since we've already lost the git history for these new files, this is also probably a good moment to set up `ocamlformat` for these files as well.

Thoughts @samoht @Drup ?

lib/mirage.ml
```diff
@@ -23,21 +23,12 @@ module Key = Mirage_key
...
-let with_output ?mode f k err =
-  match Bos.OS.File.with_oc ?mode f k () with
-  | Ok b -> b
-  | Error _ -> R.error_msg ("couldn't open output channel for " ^ err)
-
-let connect_err name number =
-  Fmt.strf "The %s connect expects exactly %d argument%s"
-    name number (if number = 1 then "" else "s")
+let with_output = Mirage_impl_misc.with_output
+let query_ocamlfind = Mirage_impl_misc.query_ocamlfind
```
hannesm: I'd `open Mirage_impl_misc` in `Mirage`

hannesm:
LGTM, I also agree with renaming to `type t` and the ocamlformat suggestion!

samoht:
LGTM but do you really needthe `mirage_impl` prefix? This could be handled by dune namespaces, right?

emillon:
A wrapped library would be nice indeed, but at the moment it would mean that `mirage` (public library) would depend on `mirage-impl` (private library), which is not supported at the moment. That will be a nice refactoring once it will be available.

emillon:
Done. If that's fine with you I'll rename/reformat individual modules in further more focused PRs.

samoht:
Thanks!

Drup:
I'm okay with the general idea, but I think "one module per device" is quite a bit overkill. I would have made big categories, so that people have a easier time working on those. I guess we will see in practice how it holds.

PR #935, opened 2018-10-15
Test fs module

emillon:
We have to start from somewhere!

This sets up tests for the code generation in the FS module, and normalizes the code a bit:
- `ocamlformat` - I reused an existing config so that at least we have terminator sequences & vertical type declarations (this is not enforced in CI yet)
- rename `fs` to `t` & `typ`

As follow ups, I'll setup ppx codegen in a module that generates some nontrivial ocaml, and next I'll convert the rest at the same time if there's not much going on.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,7 @@
+break-cases=all
+break-infix=fit-or-vertical
+field-space=loose
+margin=79
+parens-tuple=always
+sequence-style=terminator
+type-decl=sparse
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
    - POST_INSTALL_HOOK="sh ./.travis-ci.sh"
    - PINS="mirage.dev:. mirage-types.dev:. mirage-types-lwt.dev:. mirage-runtime.dev:."
    - EXTRA_REMOTES="https://github.com/mirage/mirage-dev.git"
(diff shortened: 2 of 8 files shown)


Drup:
While you are in there, could you change the shell script generation code to
```
fprintf chan {|
.....
|}
a b c d
```

emillon:
I tried this but I don't think it improves the implementation. This requires aligning the string to the first column, and this moves all interpolated arguments at the end, further from their use.

I agree that there are a lot of strings there and it's not very pretty, but maybe that's something we can solve later by having `Mirage_job` emit a shell script (though the current situation is probably good enough).

Drup:
> This requires aligning the string to the first column, and this moves all interpolated arguments at the end, further from their use.

Well, as far as I'm concerned, those are good things. At least we will be able to read it like a shell script with holes, instead of the current unreadable conglomeration of strings.

emillon:
OK, let's try this!

Drup:
@emillon Does test run on CI ? It looks like it doesn't.

emillon:
Seems to be all clear now: https://travis-ci.org/mirage/mirage/jobs/442638593#L2380

emillon:
Can we move this forward? ping @samoht

samoht:
LGTM

Drup:
Sorry! Looks good! :+1:

PR #936, opened 2018-10-23
migrate to opam2

hannesm:


mirage-types.opam
```diff
@@ -1,23 +1,22 @@
...
   ["dune" "build" "-p" name "-j" jobs]
-]
-build-test: [
-  ["dune" "runtest" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 
-depends:   [
+depends: [
+  "opam" {>= "4.04.2"}
```
samoht: should be `ocaml`

samoht:
```suggestion
  "ocaml" {>= "4.04.2"}
```

hannesm:
@samoht thanks for spotting this, I should have used `opam-package-upgrade` instead (but last time I checked it couldn't import synopsis + description from a separate file).

I additionally pushed a commit that emits `synopsis` into the generated opam files -- this should make the CI happy (and/or at least `opam2.0.1 lint`) :)

PR #938, opened 2018-11-03
adapt to charrua-client changes: Dhcp_client requires a random device now

hannesm:


yomimono:
I've merged the changes in charrua-client corresponding to this PR.  We'll need to do a release on that side so we can bound the client use correctly here.

hannesm:
@yomimono thanks, yes.  The changes are already staged in https://github.com/mirage/mirage-dev (pointing to my branches, with increased version numbers and various bounds (see upper bounds #946)). I already added conflicts with `charrua >= 0.11` to older mirage versions (in https://github.com/ocaml/opam-repository/pull/12937). but yes, there still need to be lower bounds for charrua in mirage (interface change).

hannesm:
hmm, since the conflicts in opam-repository for older mirage versions are `charrua >= 0.11`, it is easier to release charrua 0.11.0 now (rather than 0.10.1 as mentioned above). sorry for confusion, haven't had enough coffee yet.

hannesm:
merging since the corresponding charrua PR has been merged :) lower bounds are introduced in #946

Issue #939, opened 2018-11-05
release 3.3.0


hannesm:
@mirage/core I plan to do a Mirage release this week - please let me know if you'd like to include one or the other outstanding PR (feedback and merging/approving PRs is welcome as well) - noteworthy changes:
   - [x] introducing mirage-bootvar-unix #931 (unified semantics for boot arguments across all (not entirely sure, at least virtio, unix, hvt, likely xen) targets!)
   - [x] mirage-random-stdlib: new opam package that provides a properly seeded Random module (used to be part of mirage-random, is no longer) -- not to be used directly (instead, use mirage utility and the Mirage_random.C interface) :)
   - [x] the OS module no longer includes an `Env` module (used to be used on Unix to retrieve argv, now via mirage-bootvar-unix)
   - [x] segregation into `_impl` modules #933
   - [x] nicer backwards-compatible handling of `-t ukvm` #937
   - [x] move to dune #927
   - [x] opam2 output #926 #936
   - [x] charrua-client uses the old Mirage_random.Stdlibrandom which is no longer existing, see https://github.com/mirage/charrua-core/pull/87
   - [x] Dhcp_client now needs the Random explicitly #938
   - [x] remove io_page configuration #940
   - [x] generate upper bounds #946 
   - [x] mirage-tcpip (uses Stdlibrandom in tests) https://github.com/mirage/mirage-tcpip/pull/377
   - [x] Mirage/Solo5: fix handling of Cstruct buffer offset in block/net stubs (mirage/mirage-solo5#37)
   - [x] Native Genode target (#942)

to stage these changes:
   - [x] old mirage-os-shim won't be happy with old mirage-{xen,unix,solo5} (due to "missing" OS.Env module), I'll need to put some upper bounds
   - [x] old mirage versions will use Mirage_unix.OS.Env.argv, need to conflict with newer mirage-unix
   - [x] old mirage versions will use Mirage_random.Stdlibrandom for randomness  ~~I plan to add upper bounds of mirage-random in existing mirage-types{-lwt?}~~, conflict with newer mirage-random
   - [x] opam-repository PR introducing upper bounds at https://github.com/ocaml/opam-repository/pull/12937
   - [x] opam repository PR introducing upper bounds in tcpip for mirage-random https://github.com/ocaml/opam-repository/pull/12941
   - [x] Solo5 upper bounds in OPAM ocaml/opam-repository#12944

releases required:
[...]

PR #941, opened 2018-11-06
remove conflicts from opam - mirage generates code with such lower bounds

hannesm:
no need to keep this information a second time

### DIFF
--- a/mirage.opam
+++ b/mirage.opam
@@ -25,14 +25,6 @@ depends: [
   "logs"
   "mirage-runtime"     {>= "3.2.0"}
 ]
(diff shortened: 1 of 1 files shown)


mato:
If I understand this correctly, what you're saying is that these "conflicts" only apply to unikernels generated with `mirage`, not to the front-end tool itself. Correct?

hannesm:
Yes, exactly. See https://github.com/mirage/mirage/issues/939#issuecomment-436352589 for some explanation why we introduce some upper bounds via conflicts to the `mirage` front-end tool.

PR #944, opened 2018-11-07
Release engineering for 3.3.0

hannesm:
this includes #942 #941 #940 #938 and a revival of #855 which adds lots of upper bounds:
- I bound some packages (tcpip, mirage-qubes, fat, conduit etc.) to variables to avoid repeating the version constraints
- if version < 1.0.0 then min:a.b.c max:a.(b+1).0
- if version > 1.0.0 then min:a.b.c max:(a+1).0.0
- for tcpip it's min:3.5.0 max:3.6.0 (i expect there to be changes before 4.0.0) ;)

PR #940, opened 2018-11-06
remove io-page configuration

hannesm:
proposed in https://github.com/mirage/mirage/pull/855#issuecomment-319688993

successfully tested with mirage-skeleton/device-usage/kv_ro :)

PR #946, opened 2018-11-08
upper bounds

hannesm:
extracted from #944, superseeds #855:
- some packages (tcpip, mirage-qubes, fat, conduit etc.) to variables to avoid repeating the version constraints
- if version `< 1.0.0` then `~min:"a.b.c"` `~max:"a.(b+1).0"`
- if version `> 1.0.0` then `~min:"a.b.c" `~max:"(a+1).0.0"`
- an exception is tcpip with `~min:"3.5.0"` `~max:"3.6.0"`
- `mirage-block-ramdisk` is still unconstrained
- is this ok for e.g. `charrua` (which uses 0.x, constrained to `< 0.11`), `mirage-qubes` (also 0.x, constrained to `< 0.7`), `mirage-profile*` (constraint to `< 1.0.0`)

~~may result in CI failure (until #938 is merged)~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
  global:
    - PRE_INSTALL_HOOK="cd /home/opam/opam-repository && git pull origin master && opam update -u -y"
    - POST_INSTALL_HOOK="sh ./.travis-ci.sh"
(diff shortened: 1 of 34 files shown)


hannesm:
rebased now that #940 got merged. /cc @mirage/core

Issue #852, opened 2017-07-26
`make depend` adds a pin but doesn't remove it after failing


yomimono:
`make depend` adds a pin in order to install dependencies for the unikernel before building.  If the dependency installation succeeds, the pin is removed, but in the case of failure (network problem, solver failure, etc) the pin is not removed.

hannesm:
I agree this is a bug.  Unfortunately my Makefile skills are not sufficient to formulate "in any case, please `opam pin remove` after opam install --deps-only".   Is there a way to get away without `pinning` at all in opam?

As alternative, we could move the `opam pin` and `depext` into the mirage cli...

Drup:
> Is there a way to get away without pinning at all in opam?

Well, yes. Opam 2.0 :)

hannesm:
now, opam pin is only used during depext. this can be removed once depext gets support for local opam files.

PR #948, opened 2018-11-15
avoid opam pin during installation of dependencies in generated makefile

hannesm:
unfortunately, opam depext AFAICT doesn't support a local opam file, it needs
 to get the opam file pinned

addresses #852


hannesm:
the second commit also fixes #401 again by re-adding a `no-depext` flag to mirage. btw, `warn-error` used to be a bool parameter (i.e. you'll have to pass `--warn-error true`), it's now a flag as well `--warn-error`.

hannesm:
ci is green, merging

PR #952, opened 2018-12-06
Ipv4 signature

hannesm:
The signature of Static_ipv4 is changing, see https://github.com/mirage/mirage-tcpip/pull/382


hannesm:
will rebase with #1006

PR #953, opened 2018-12-13
Add mirage dependency to all the generated unikernels

hannesm:
to solve (1b) //cc @linse @rjbou @marmarek


Drup:
Let's just make sure that it stays a build dependency and doesn't get linked in! ;)

hannesm:
@Drup agreed. this one does not make mirage a run-time dependency. anyone against this PR, otherwise it would be great to have this merged and part of the next mirage release!

Drup:
I don't think there is really any argument against. Let's merge!

PR #955, opened 2018-12-17
Httpaf types

anmonteiro:
This is a proposal to add http/af support in Mirage. Happy to take suggestions on naming / config.

This depends on https://github.com/inhabitedtype/httpaf/pull/83


lib/mirage.mli
```diff
@@ -457,12 +457,18 @@ type conduit
...
 
 type http
 val http: http typ
 val http_server: conduit impl -> http impl
 
+(** {2 HTTP (http/af) configuration} *)
+
+type httpaf
+val httpaf: httpaf typ
+val httpaf_server: conduit impl -> httpaf impl
```
hannesm: I'm curious why/whether there's need for `type httpaf` -- if we'd reuse `type http`, `httpaf` could be a drop-in replacement for `cohttp` -- similarly to how `arp` (using https://github.com/mirage/mirage-tcpip/tree/master/src/arpv4) and `farp` (using https://github.com/hannesm/arp) are defined.

anmonteiro: There probably isn't. I mostly didn't know what I was doing until everything compiled and ran.
I'll clean this up, I agree that `httpaf_server` could probably be:

`val httpaf_server: conduit impl -> http impl`

hannesm: yes, I just added links to how arp is done in mirage.mli / mirage_impl_arpv4. thanks!

anmonteiro:
@hannesm I addressed your comments, let me know if there's something that could still look better.

hannesm:
LGTM! I'd appreciate some docstrings for the HTTP section, since now I'm confused why there's `http` and `httpaf`. On a related note, I'd appreciate to rename `val http` to `val cohttp`, and add a deprecated `http` (alias to cohttp for now), please see https://github.com/mirage/canopy-data/blob/master/Posts/Deprecating for more details if you are interested.

anmonteiro:
Done. Would you recommend different docstrings or deprecation message?

hannesm:
thanks again @anmonteiro! The `@@deprecated` annotation is only needed in the `mirage.mli` AFAICT, no need to retain it in `mirage.ml`.

anmonteiro:
Ah right. Makes sense. Updated.

Drup:
Just looked at the latest version, looks very nice (and definitely better than the initial one, I agree with @hannesm's comments)

dinosaure:
Good job :+1: ! Are you able to make a PoC of `httpaf` on `mirage` ?

anmonteiro:
@dinosaure there’s an example here: 
https://github.com/inhabitedtype/httpaf/pull/83

Is that what you were looking for?

hannesm:
imho we can merge this now, and once httpaf-mirage is released to opam-repository add some tests to mirage-skeleton using httpaf. any other opinions?

lib/mirage.mli
```diff
@@ -461,7 +461,15 @@ val conduit_direct: ?tls:bool -> stackv4 impl -> conduit impl
 
 type http
 val http: http typ
+
 val http_server: conduit impl -> http impl
+[@@ocaml.deprecated "`http_server` is deprecated. Please use `cohttp_server` or `httpaf_server` instead."]
```
avsm: Instead of deprecating this, how about just making it a choice between Cohttp and Httpaf via an optional argument, and then swapping the defaults in a future release as httpaf support matures?  Most users will probably not care too much about the specific server implementation chosen, and advanced users have the functions below with the specific choice.

avsm:
@hannesm I'd prefer not to deprecate `http_server` if possible (see my comment inline)

anmonteiro:
Do you mean via the `mirage configure ...` command?

hannesm:
@anmonteiro yes, the user should be able to:
`mirage configure --http httpaf` and `mirage configure --http cohttp` to select configure-time the http implementation -- similar to `--prng stdlib` and `--prng fortuna` where `stdlib` is the default atm.

to implement this, this new configure-time argument needs to be implemented in mirage_key.ml/mli  (the cohttp and httpaf bindings you developed in this PR should be kept) - it should be very similar to the prng key

samoht:
> unikernel code would need a common interface (i.e. a Mirage_types.HTTP

That would be a good idea to add in a later PR!

lib/mirage.mli
hannesm: since there is no common signature of httpaf and cohttp yet (esp. in respect to headers etc.), I don't think application developers can be http-server agnostic right now. in the future, this will hopefully be cleared up.

hannesm:
merging -- this improves the current situation, and allows to easily use httpaf in the mirage context (once httpaf-mirage is released, this will also work out of the box). thanks!

PR #956, opened 2018-12-20
Use pp and of_string from Ipaddr modules instead of s-expressions

hannesm:
The problem I try to solve: a hello world (not stripped, no lto, hvt target) is on my machine 3MB in binary size. I investigated why this is the case. A major consumer of binary size is the sexplib library, which is required by the ocaml-ipaddr library at the moment.

Mirage itself also uses the s-expression conversion for arguments. This PR uses pp and of_string instead.

Once ipaddr will split its sexp usage into a separate library (as uri did recently, introducing uri.sexp), a resulting hello world unikernel will use only 2MB in binary size.

This PR is ready for review, and I'm interested in your feedback. At its current state, it assumes that "forall ip. of_string_exn (pp ip) = ip", is this reasonable? I tried to use `to_string` instead, but `Ipaddr.V6.to_string` has a different signature (by having the optional labeled keyword `v4`) than `Ipaddr.V4.to_string`.

### DIFF
--- a/lib/mirage_key.ml
+++ b/lib/mirage_key.ml
@@ -23,34 +23,34 @@ open Astring
 module Arg = struct
   include Key.Arg
 
+  let from_run s = "Mirage_runtime.Arg." ^ s
+
+  let make d m of_string pp =
+    let parser s = match of_string s with
+      | Some ip -> `Ok ip
+      | None -> `Error ("Can't parse ip address: "^s)
+    in
+    let serialize ppf t =
+      Fmt.pf ppf "(Ipaddr.%s.of_string_exn \"%a\")" m pp t
+    in
+    Functoria_key.Arg.conv
+      ~conv:(parser, pp)
+      ~serialize
+      ~runtime_conv:(from_run d)
+
   module type S = sig
-    include Mirage_runtime.Arg.S
-    val sexp_of_t : t -> Sexplib.Type.t
+    type t
+    val of_string : string -> t option
+    val pp : Format.formatter -> t -> unit
   end
 
-  (* Could be improved with metaocaml and/or some reflection.
-     [description] and [m] should not need to be provided.
-  *)
-  let of_module
-      (type t) runtime_conv m (module M: S with type t=t) =
-    let conv = Mirage_runtime.Arg.of_module (module M) in
-    let serialize ppf x =
-      Fmt.pf ppf "(%s.t_of_sexp (Sexplib.Sexp.of_string %S))"
-        m (Sexplib.Sexp.to_string @@ M.sexp_of_t x)
-    in
-    Functoria_key.Arg.conv ~conv ~serialize ~runtime_conv
+  let of_module (type t) d m (module M:S with type t = t) =
+    make d m M.of_string M.pp
 
-  let from_run s = "Mirage_runtime.Arg." ^ s
-  let builtin d mn m = of_module (from_run d) mn m
-
-  let ipv4_address = builtin "ipv4_address" "Ipaddr.V4" (module Ipaddr.V4)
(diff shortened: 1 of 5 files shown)


hannesm:
@mirage/core has an opinion how to proceed here? is pp and of_string a good solution? I'd really like to chop the 1MB off a hello world

avsm:
I’ve not had a chance to understand if that property holds for ipv6. I’d be happy to split the ipaddr sexp functionality out...

hannesm:
@avsm: having a separate ipaddr.sexp sublibrary would indeed be great :)

> that property holds for ipv6
`Ipaddr.V6.pp ppf ip` calls `Ipaddr.V6.to_string ip` (without `?v4`), which then calls `Ipaddr.V6.to_buffer`:
```OCaml
    let v4 = match comp with
      | (0,0,0,0,0,0xffff,_,_) -> true
      | _ -> v4
    in
...
      | [Some hi;Some lo] when v4 ->
        let addr = V4.of_int16 (hi, lo) in
        V4.to_buffer buf addr
```
The `Ipaddr.V6.of_string_exn` (calls `Ipaddr.V6.parse`):
```OCaml
        else if !i < len && s.[!i] = '.'
        then begin
          i:= pos;
          let v4 = V4.of_string_raw s i in
          let (hi,lo) = V4.to_int16 v4 in
          lo :: hi :: acc
        end
```

so I suspect that `ppf` and `of_string` handle both IPv6 addresses (with compression), as well as IPv4 addresses, the code references https://tools.ietf.org/html/rfc5952 -- maybe we should apply some fuzz-testing to test the invariant that "forall ip. of_string_exn (pp ip) = id" for the Ipaddr module and their submodules.

My intuition is that the IPv6 module should only handle IPv6 data, the IPv4 module only IPv4 data, and the IP module takes care of either. Now since IPv6 may contain IPv4 addresses, is there any need for the (main) Ipaddr module and its Ipaddr.t?

avsm:
See https://github.com/mirage/ocaml-ipaddr/pull/79 for the first cut. I agree with the possibility of treating V6 address parsing as a superset of V4, but I think that'll be a bigger change -- in scope for the major bump after this one.

hannesm:
force-pushed to work with ipaddr 3.0 \o/

lib/mirage_key.ml
```diff
@@ -23,50 +23,54 @@ open Astring
...
+      ~conv:(parser, pp)
+      ~serialize
+      ~runtime_conv:(from_run d)
 
-  let from_run s = "Mirage_runtime.Arg." ^ s
-  let builtin d mn m = of_module (from_run d) mn m
+  module type S = sig
+    type t
+    val of_string : string -> (t, [ `Msg of string ]) result
+    val to_string : t -> string
```
samoht: The Cmdliner way of defining these is to have a `pp` function here instead of a `to_string`

hannesm: this signature is used by `make` above, for (a) Cmdliner style parser and pretty printer, but also (b) serialise and deserialise between configuration and runtime phase -- in this second phase we use `to_sting` directly.

I find it easier to reason that `\forall x . of_string (to_string x) = Ok x` than to include a pretty printer in that equation (which btw was the initial implementation in this PR, as written in the first comment). But I'm open for discussion why/whether we should use `pp` and `of_string` instead.

samoht: pp combinators are just more efficient to combine, but I don't have a strong opinion.

samoht: And you should probably use `Fmt.to_to_string` and `Fmt.of_to_string` to convert between `pp` and `to_string` functions.

PR #957, opened 2018-12-20
Use String.split_on_char instead of Astring in mirage-runtime

hannesm:
fewer dependencies are better, and split_on_char exists since 4.04.0 -- mirage depends on >=4.04.2.

most unikernels won't see an improvement in binary size, since they usually use parse-argv that depends (rightfully!) on Astring -- but we have the `no_argv` device -- if your unikernel is not interested in boot arguments that can leverage ~300kB of binary size.

no, I've no plans to remove the fmt and/or logs dependency from mirage-runtime ;)

### DIFF
--- a/lib_runtime/dune
+++ b/lib_runtime/dune
@@ -2,4 +2,4 @@
   (name        mirage_runtime)
   (public_name mirage-runtime)
   (wrapped     false)
-  (libraries   functoria-runtime ipaddr astring logs fmt))
+  (libraries   functoria-runtime ipaddr logs fmt))
--- a/lib_runtime/mirage_runtime.ml
+++ b/lib_runtime/mirage_runtime.ml
@@ -87,10 +87,10 @@ module Arg = struct
       with Not_found -> "warning"
     in
     let parser str =
-      match Astring.String.cut ~sep:":" str with
-      | None            -> `Ok (`All    , level_of_string str)
-      | Some ("*", str) -> `Ok (`All    , level_of_string str)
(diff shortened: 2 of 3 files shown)


hannesm:
will merge unless someone objects @mirage/core

avsm:
The semantics here seem different with multiple occurrences of the char ... is that a problem?

hannesm:
@avsm not really. it is used internally for the `-l` argument (which uses a `list` of log_threshold), with the possible inputs:
"debug" -> `All, level_of_string "debug"
"*:debug" -> `All, level_of_string "debug"
"my_source:debug" -> `Src "my_source", level_of_string "debug"

earlier, we used `cut ~sep:":" input -> (string * string) option`, now we use `split_on_char ':' input -> string list`: `None -> empty list`, `Some (a, b) -> [ a; b ]`.

it will be an issue whenever a log level contains a `:`. if a source contains `:`, instead of preserving the existing (broken?) behaviour to use the part before the first `:` as sourcename, the remainder as level, we now error (earlier, level_of_string would error)

avsm:
Makes sense, thanks for the explanation!

cfcs:
I don't think `String.split_on_char` can return an empty list since if there's no match it will return the original string, shouldn't this be `[ _ ]` or `[ str ]` instead of `[]` below?
```ocaml
      | [] -> `Ok (`All, level_of_string str)
```

hannesm:
@cfcs yes, see https://github.com/mirage/mirage/pull/961 -- thanks for your comment!

PR #958, opened 2018-12-20
default to socket stack on unix and macosx when using generic_stackv4

hannesm:
currently, when compiling for unix, the default network stack is the direct one -- unless `--net=socket` is provided. various unikernels contain the following boilerplate in the config.ml:
```OCaml
let net =
  if_impl Key.is_unix
    (socket_stackv4 [Ipaddr.V4.any])
    (static_ipv4_stack ~config:address default_network)
```
to use the socket stack by default on unix. This is not a good idea in several aspects, including that now on unix it is no longer possible to configure a direct stack.


hannesm:
I tested this locally, it is for me a great development improvement (mirage configure no longer needs `--net socket` in my dev cycle :) -- could we get this merged, please?

hannesm:
ping @mirage/core

PR #959, opened 2018-12-23
Mirage_key: remove ?group parameter from PRNG, there's only one PRNG

hannesm:
and improve documentation of PRNG key

### DIFF
--- a/lib/mirage_impl_random.ml
+++ b/lib/mirage_impl_random.ml
@@ -60,7 +60,7 @@ end
 let nocrypto_random = impl nocrypto_random_conf
 
 let default_random =
-  match_impl (Mirage_key.value @@ Mirage_key.prng ()) [
+  match_impl (Mirage_key.value Mirage_key.prng) [
     `Stdlib  , stdlib_random;
     `Nocrypto, nocrypto_random;
   ] ~default:stdlib_random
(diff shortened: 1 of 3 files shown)


hannesm:
any opinion on this PR?

avsm:
seems good to me -- what was the original purpose of the `group` parameter in the first place? we seem to expose the same choices with stdlib/fortuna already...

Drup:
`group` is there to easily make multiple prng devices and obtain different configuration keys on the terminal.

@hannesm Do you think having multiple different prngs is useless ? I don't really object, but it seems a bit strong.

hannesm:
at first, I'd like to know how you'd have in a MirageOS unikernel multiple PRNGs? In the [mirage-random interface](https://github.com/mirage/mirage-random/blob/master/src/mirage_random.ml) I don't see how to do that. similar to Time, PRNG functions do not take a `type t` argument. I'm fine with this (and hope to move PCLOCK/MCLOCK to the same shape). is there a real world example where having two PRNGs is useful? how should the entropy gathering work here then? do both get seeded with the same entropy? atm, nocrypto/mirage-random-stdlib call out to mirage-entropy.connect and register themselves as consumer thereof for the main event loop. I don't have an intuitive understanding how this should work with multiple PRNGs.

in the unix world, we only ever have a single /dev/[u]random - imho this is really sufficient, and having multiple is confusing.

If you really need a custom PRNG for some random, you can pass a static seed `?g` to `generate` and get your deterministic random. this works fine - there's an example at https://github.com/roburio/udns/blob/ef6dc0a07a50a6c81071711bf0166f4540c972eb/mirage/dns_mirage_certify.ml#L167-L182

Drup:
@hannesm You are the PRNG expert in the room, so I'll trust you on that one! ;)

PR #965, opened 2019-02-02
adjust to arp and ethernet changes

hannesm:
originally, they were part of tcpip, and there was an alternative arp implementation in a separate repository.

in https://github.com/mirage/mirage-tcpip/issues/386, we discussed to remove the tcpip-builtin arp implementation and use the alternative one by default. to avoid cyclic opam package dependencies, the ethernet (used by arp) package needed to be split out from tcpip as well.

This PR is good to go, but needs to be merged together with:
- [x] https://github.com/mirage/mirage-tcpip/pull/390
- [x] https://github.com/mirage/arp/pull/9
- [x] https://github.com/mirage/charrua-core/pull/91
- [ ] a release of https://github.com/mirage/ethernet
- [x] https://github.com/mirage/mirage-skeleton/pull/263

Feedback is welcome, the set of PRs do not change the API of network layers (though, I plan to do this in subsequent PRs to avoid that e.g. IP needs to know of the ethernet implementation). A successful build of mirage-skeleton with the above changes is at https://travis-ci.org/mirage/mirage-dev/builds/487722396

Merge plan: rather sooner than later ;)

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,5 +1,5 @@
 eval `opam config env`
 opam depext -uiy mirage
 cd ~
-git clone -b master https://github.com/mirage/mirage-skeleton.git
+git clone -b ethernet https://github.com/hannesm/mirage-skeleton.git
 make -C mirage-skeleton && rm -rf mirage-skeleton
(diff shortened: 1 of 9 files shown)


hannesm:
the changes are merged to master now, awaiting releases of the packages above (which @avsm will take care of).

PR #968, opened 2019-02-09
[RFC] TCP/IP stack MTU, layering, and write allocation discipline changes

hannesm:
A brief list of changes:

- Ethernet encapsulation is done in Ethernet - currently done in Arp and IPv4 and IPv6 libraries
- MTU is passed from the network interface upwards - currently passed into the Ethernet layer
- Write buffer allocation is done by the network interface, which knows best about page boundary restrictions, etc. - currently done in IPv4, IPv6, ARP using Io_page.get 1
- write functions signature change throughout the stack up to Ip: the buffer to send is no longer given directly, instead a size argument and a fill function of type (buffer -> int).
- IPv4 applies fragmentation!
- no longer used function are removed

I did performance evaluation as follows (I'm pretty sure this is not exhaustive, and likely others have better ideas on how to actually do such a benchmark):
- Server: the MirageOS unikernel iperf_server (TCP, thanks to Taka), https://github.com/hannesm/mirage_iperf, virtualized with Solo5/hvt, deployed on a PC Engines (AMD Embedded G series GX-412TC, 1 GHz quad Jaguar core)
- Client: iperf on Unix running on my Lenovo X250, arguments -i 1 -t 20
- Connection was a 1 GBit/s copper cable (a switch was in between the devices)

The latest released opam packages (mirage 3.4.1, tcpip 3.7.0):
Throughput = 16.97 [MBs/sec]
gc minor words 214425271.000000 promoted words 4421898.000000 major words 4428911.000000 minor collections 884 major collections 131 heap_words 61440

With the changes above:
Throughput = 16.85 [MBs/sec]
gc minor words 212194711.000000 promoted words 4286488.000000 major words 4293501.000000 minor collections 876 major collections 132 heap_words 61440

I don't think there's any significant difference between old and new.

This PR adjusts only version constraints in mirage, the actual changes are in PRs in other repositories:
Interfaces:
- [x] https://github.com/mirage/mirage-net/pull/16 and https://github.com/mirage/mirage-net/pull/18 https://github.com/ocaml/opam-repository/pull/13511
- [x] https://github.com/mirage/mirage-protocols/pull/15 https://github.com/ocaml/opam-repository/pull/13516

Mirage-net providers:
- [x] https://github.com/mirage/mirage-net-macosx/pull/31 and https://github.com/mirage/mirage-net-macosx/pull/32 https://github.com/ocaml/opam-repository/pull/13519
- [x] https://github.com/mirage/ocaml-tuntap/pull/32 (providing `get_mtu`) https://github.com/ocaml/opam-repository/pull/13513
- [x] https://github.com/mirage/mirage-net-unix/pull/46 https://github.com/ocaml/opam-repository/pull/13521
- [x] https://github.com/mirage/mirage-net-solo5/pull/28 https://github.com/ocaml/opam-repository/pull/13518
- [x] https://github.com/mirage/mirage-net-xen/pull/82 https://github.com/ocaml/opam-repository/pull/13520
- [x] https://github.com/mirage/mirage-vnetif/pull/26 https://github.com/ocaml/opam-repository/pull/13517

Network layers:
- [x] https://github.com/mirage/ethernet/pull/3 https://github.com/ocaml/opam-repository/pull/13522
- [x] https://github.com/mirage/arp/pull/13 https://github.com/ocaml/opam-repository/pull/13522
- [x] https://github.com/mirage/mirage-tcpip/pull/394 https://github.com/ocaml/opam-repository/pull/13527
- [x] https://github.com/mirage/charrua-core/pull/94 https://github.com/ocaml/opam-repository/pull/13529

Applications
- [x] https://github.com/mirage/mirage-skeleton/pull/264
- [ ] https://github.com/mirage/mirage-skeleton/pull/265

### DIFF
--- a/lib/mirage_impl_arpv4.ml
+++ b/lib/mirage_impl_arpv4.ml
@@ -13,7 +13,7 @@ let arp_conf = object
   method name = "arp"
   method module_name = "Arp.Make"
   method! packages =
-    Key.pure [ package ~min:"1.0.0" ~max:"2.0.0" "arp-mirage" ]
+    Key.pure [ package ~min:"2.0.0" ~max:"3.0.0" "arp-mirage" ]
   method! connect _ modname = function
     | [ eth ; _time ] -> Fmt.strf "%s.connect %s" modname eth
     | _ -> failwith (connect_err "arp" 3)
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,5 +1,5 @@
 eval `opam config env`
 opam depext -uiy mirage
 cd ~
-git clone -b master https://github.com/mirage/mirage-skeleton.git
+git clone -b layering https://github.com/hannesm/mirage-skeleton.git
 make -C mirage-skeleton && rm -rf mirage-skeleton
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
    - PRE_INSTALL_HOOK="cd /home/opam/opam-repository && git pull origin master && opam update -u -y"
    - POST_INSTALL_HOOK="sh ./.travis-ci.sh"
    - PINS="mirage:. mirage-types:. mirage-types-lwt:. mirage-runtime:."
-   - EXTRA_REMOTES="https://github.com/mirage/mirage-dev.git"
+   - EXTRA_REMOTES="https://github.com/mirage/mirage-dev.git#layering"
    - PACKAGE=mirage
  matrix:
    - DISTRO=debian-testing OCAML_VERSION=4.04 EXTRA_ENV="MODE=xen"
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -887,7 +887,7 @@ module Project = struct
       ]
       method! packages =
         (* XXX: use %%VERSION_NUM%% here instead of hardcoding a version? *)
-        let min = "3.4.0" and max = "3.5.0" in
+        let min = "3.5.0" and max = "3.6.0" in
         let common = [
           package ~build:true ~min:"4.04.2" "ocaml";
           package "lwt";
--- a/lib/mirage_impl_ethernet.ml
+++ b/lib/mirage_impl_ethernet.ml
@@ -12,7 +12,7 @@ let ethernet_conf = object
   method name = "ethernet"
   method module_name = "Ethernet.Make"
   method! packages =
-    Key.pure [ package ~min:"1.0.0" ~max:"2.0.0" "ethernet" ]
+    Key.pure [ package ~min:"2.0.0" ~max:"3.0.0" "ethernet" ]
   method! connect _ modname = function
     | [ eth ] -> Fmt.strf "%s.connect %s" modname eth
     | _ -> failwith (connect_err "ethernet" 1)
--- a/lib/mirage_impl_ip.ml
+++ b/lib/mirage_impl_ip.ml
@@ -33,7 +33,7 @@ let (@??) x y = opt_map Key.abstract x @? y
 
 (* convenience function for linking tcpip.unix or .xen for checksums *)
 let right_tcpip_library ?ocamlfind ~sublibs pkg =
-  let min = "3.7.0" and max = "3.8.0" in
+  let min = "3.7.1" and max = "3.8.0" in
   Key.match_ Key.(value target) @@ function
   |`MacOSX | `Unix ->
     [ package ~min ~max ?ocamlfind ~sublibs:("unix"::sublibs) pkg ]
@@ -62,7 +62,7 @@ let ipv4_keyed_conf ?network ?gateway () = impl @@ object
   end
 
 let charrua_pkg =
-  Key.pure [ package ~min:"0.11.0" ~max:"0.12.0" "charrua-client-mirage" ]
+  Key.pure [ package ~min:"0.12.0" ~max:"0.13.0" "charrua-client-mirage" ]
 
 let dhcp_conf = impl @@ object
     inherit base_configurable
--- a/lib/mirage_impl_network.ml
+++ b/lib/mirage_impl_network.ml
@@ -17,14 +17,14 @@ let network_conf (intf : string Key.key) =
     method! keys = [ key ]
     method! packages =
       Key.match_ Key.(value target) @@ function
-      | `Unix -> [ package ~min:"2.3.0" ~max:"3.0.0" "mirage-net-unix" ]
-      | `MacOSX -> [ package ~min:"1.4.0" ~max:"2.0.0" "mirage-net-macosx" ]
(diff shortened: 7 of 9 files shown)


hannesm:
I removed the two-step transaction `allocate_frame` followed by `write` -- I don't think we get much from the split into two functions, and we lose control over the allocated buffer(s). now the lifetime of a buffer is very limited by the `write` API -- once write returns, the buffer is gone.

btw, if we push the same API another layer, we can have a zero-copy sending using:
```OCaml
UDP.write_into : UDP.t -> ?src:ipaddr -> ?src_port:int -> dst:ipaddr -> dst_port:int -> ?size:int -> (buffer -> int) -> (unit, error) result io
```
as long as `size` is below MTU (on the path -- i.e. Netif.mtu - sizeof ip_hdr - sizeof udp_hdr), the buffer provided to the fill function will be allocated by the network interface. this may be worth doing for DNS etc. look into the charrua changes how this avoids allocation and copying.

also, the API as it is allows the netif to reuse buffers for sending -- since the fill function is blocking (not in the lwt monad), we can ensure there's only a single one of these executed at the same time :)

what I'm still struggling with is how to handle huge data from the application -- i.e. bigger than a page, or an ethernet frame payload... if allocated by the network interface, I've trouble designing the interface (i.e. the fill function would need a list of buffers, but that's slightly ugly, since then the application developer will need to write code for distributing their data across buffers).

avsm:
Just a quick “very positive” from me for the interface changes. I’ll have a proper read of the diffs next week.

hannesm:
I did some more measurements, esp. in respect to number of cstruct + io-page allocations (using [this io-page branch](https://github.com/hannesm/io-page/tree/stats) and [this cstruct branch](https://github.com/hannesm/ocaml-cstruct/tree/stats)):

released:
```
2019-02-10 17:04:06 -00:00: INF [application] iperf server: Duration = 20018817955 [ns] (start_t = 31289534364662163, end_t = 31289554383480118),  Data received = 337641472 [bytes], Throughput = 16866204.23 [bytes/sec]
2019-02-10 17:04:06 -00:00: INF [application] iperf server: Throughput = 16.87 [MBs/sec]
2019-02-10 17:04:06 -00:00: INF [application] gc minor words 214589154.000000 promoted words 4484821.000000 major words 4491834.000000
minor collections 889 major collections 137 heap_words 61440 heap chunks 1
compactions 4 top heap words 565248 stack size 114
2019-02-10 17:04:06 -00:00: INF [application] cstruct stats count 494519, size 350607508, buckets: 
0..10: 250451
11..50: 6392
51..100: 6386
101..200: 0
201..500: 0
501..700: 0
701..1000: 0
1001..1514: 231290
>1514: 0

2019-02-10 17:04:06 -00:00: INF [application] iopage stats number of calls 6386 number of pages 6386 size 26157056
from top: virtual 27136K, rss 6564K, time 0:19
```

with the above changes:
```
2019-02-10 17:05:07 -00:00: INF [application] iperf server: Duration = 20047845412 [ns] (start_t = 31289595377404141, end_t = 31289615425249553),  Data received = 340656128 [bytes], Throughput = 16992156.56 [bytes/sec]
2019-02-10 17:05:07 -00:00: INF [application] iperf server: Throughput = 16.99 [MBs/sec]
2019-02-10 17:05:07 -00:00: INF [application] gc after minor words 216166888.000000 promoted words 4246636.000000 major words 4253649.000000
minor collections 890 major collections 129 heap_words 61440 heap chunks 1
compactions 4 top heap words 565248 stack size 114
2019-02-10 17:05:07 -00:00: INF [application] cstruct stats count 485859, size 353707502, buckets: 
0..10: 239737
11..50: 6389
51..100: 6387
101..200: 0
201..500: 0
501..700: 0
701..1000: 0
1001..1514: 233346
>1514: 0

2019-02-10 17:05:07 -00:00: INF [application] iopage stats number of calls 0 number of pages 0 size 0
from top: virtual 27136K, rss 6504K, time 0:19
```

by no means exhaustive or real benchmarks -- feel free to run your own (I used https://github.com/hannesm/mirage_iperf/tree/4e3ce248dd081e8d9915025c547602974f908ad3) -- but I'd claim this patch series doesn't make it worse ;)

avsm:
I've extended vmnet to also expose the mtu from vmnet>=1.4.0

hannesm:
@avsm thanks, I added a small PR for mirage-net-macosx which uses the `Vmnet.mtu`.

@mato was not satisfied with the hardcoded magic number 14 (see https://github.com/mirage/mirage-net-solo5/pull/28) for the ethernet_header_size.  I adapted mirage-net implementations to use a binding for that number instead.

hannesm:
further discussion in https://github.com/mirage/mirage-net-solo5/pull/28 and with @mato on IRC lead me to the following thoughts:
- 8021q is hard, we leave it off the table for now!
- with this set of PRs, the mirage-net implementations magically add ethernet header sizes to incoming and outgoing packets (but do not encapsulate / decapsulate them), since they're now responsible for allocation of buffers.
--> this means `14` (as ethernet_header_size) is hardcoded in all mirage-net implementations.

an alternative would be:
- ethernet is responsible for ethernet decapsulation/encapsulation and reporting the ethernet_header_size to the mirage-net implementation
- this is trivially done on the send path with these PRs
- this needs an additional argument to `listen` for the receive path
--> mirage-net would just be an abstraction over different network interfaces, providing mtu, (potentially feature flags), and allocation (disciplines) for network frames. no encapsulation done in there

I'll give this alternative a try later today and maybe push updates.

hannesm:
I did the above changes (and pushed to the existing PRs / made a new one for mirage-net), and found on the way an issue in the (new) IPv4 fragmentation code (the length field was incorrect by not including the IPv4 header :/).
The revised API is cleaner:
- Mirage-net implementations no longer contain the ethernet header size information
- `listen` takes an `header_size:int` argument, the `size:int` passed to `write` is mandatory now
- Ethernet adds the Ethernet_wire.sizeof_ethernet in calls to `write`
- the netif implementations can no longer return `Exceeds_mtu, but Ethernet does the check instead
  --> netif.write is slightly less safe, it gets a buffer to put on the wire. The underlying interface may complain about the length.

lib/mirage_impl_arpv4.ml
```diff
@@ -13,7 +13,7 @@ let arp_conf = object
   method name = "arp"
   method module_name = "Arp.Make"
   method! packages =
-    Key.pure [ package ~min:"1.0.0" ~max:"2.0.0" "arp-mirage" ]
+    Key.pure [ package ~min:"2.0.0" ~max:"3.0.0" "arp-mirage" ]
```
cfcs: these don't follow the pattern of `y` being bumped in `x.y.z` translating to an interface break?

PR #978, opened 2019-04-04
Add HTTP/2 types for h2-mirage

anmonteiro:


lib/mirage_impl_http.ml
```diff
@@ -2,30 +2,79 @@ open Functoria
...
-    method name = "httpaf"
-    method module_name = "Httpaf_mirage.Server_with_conduit"
-    method! packages =
-      Mirage_key.pure [ package "httpaf-mirage" ]
-    method! deps = [ abstract conduit ]
-    method! connect _i modname = function
-      | [ conduit ] -> Fmt.strf "%s.connect %s" modname conduit
-      | _ -> failwith (connect_err "httpaf" 1)
-  end
+let cohttp_server conduit =
```
samoht: Can you try to minimize the diff on that file? We are not using ocamlformat on mirage yet, so for now the stye guide  is to reduce diff size ;-)

anmonteiro: oh wow.. I don't know how this got here :) probably ocamlformat enabled by default. Will do

samoht:
Is there any hope to have a common HTTP/1 et HTTP/2 interface for unikernels? I know @TheLortex tried to have a common API the cohttp and httpaf but I'm not sure that's really the way forward anymore. One solution would be to evolve the cohttp API to look way more like httpaf.

avsm:
@samoht: yes, see https://github.com/anmonteiro/ocaml-h2/pull/37

anmonteiro:
@samoht reverted the formatting. I figured out what happened. I have auto-formatting on save setup in Vim when it detects and ocamlformat project, and there's indeed a config file in this repo :)

samoht:
@avsm sweet! Can we get this somehow as a combinator? For an end-user perspective, it doesn't really make sense to offer 3 different http devices...

TheLortex:
Blocker for the common API are PRs in httpaf that enables a Mirage adapter and response with upgrade:
https://github.com/inhabitedtype/httpaf/pull/83 and https://github.com/inhabitedtype/httpaf/pull/91

Issue #969, opened 2019-02-12
Use Dune instead of OCamlbuild to build unikernels


TheLortex:
I used several unikernel samples to explore what are the problems blocking the Dune transition. I've taken them from `mirage-skeleton`. After identifying problems and workarounds, I managed to build entirely with Dune + ld/pkg-config the following unikernels with unix, xen and hvt target: DNS, http_fetch, echo_server. I'm aware this does not cover everything but it's a good start.

# Used features of OCamlbuild that aren't implemented in Dune:
1. `-dontlink <pkg>` used for `unix`, `threads`, `str` and `num`. This removes these packages from the final linking step, thus allowing to link on non-Unix targets.
2. Predicates is sometimes used for library selection.
3. Use of `-output-obj` in conjunction with `..._linkopts` in META files to link C stubs and custom OCaml runtime.

# Workaround to each problem
1. Explore the dependency graph to figure out which packages have unneeded dependencies and update them accordingly. Use OCaml 4.07 and Stdlib.Bigarray in order not to depend on the legacy `bigarray` which includes undefined stubs on xen/freestanding.
2. Use virtual libraries and implementations. To not break everything, it's possible to keep the library name `package` as the default implementation and make it depend on the virtual library `package-virtual`. This way the update is transparent for users of the package and the ones that want to use the virtual library feature just have to change their `package` dependency into a `package-virtual` dependency.
3. Dune uses `-output-complete-obj` to build object files. This also links C stubs and runtime. Therefore we don't have to care about C stubs anywore as they are automatically linked along with used libraries. No need for `..._linkopts` flags in META. For the custom runtime, `ocamlopt` has a `-runtime-variant` option to choose a custom runtime library to link.

# Packages to update
1. ocplib-endian, mirage-fs-lwt, mirage-console-lwt, xen-evtchn, sexplib, tcpip, asn1-combinators, mirage-entropy
2. mirage-entropy, zarith + dependencies (asn1-combinators, nocrypto)
3. mirage-xen-ocaml, ocaml-freestanding: just need to add a symlink from `lib...asmrun.a` to ocaml runtime directory so that ocaml finds the runtime variants. I've already modified these packages locally to make my experiments so it's not that hard to make these changes.

# Issues and comments
1. Duniverse forks may have already fixed some of these packages.
2. The uses of virtual libraries disable cross-module optimisation such as inlining. We have to be careful with that. 

[...]

PR #979, opened 2019-04-08
Use dune to build unikernels

TheLortex:
I'm creating this PR in order to track the modifications of the mirage CLI I've made for mirage/mirage#969.

`mirage config -t <target>` creates a `dune` file with all the rules needed to build the end user unikernel. This allows `mirage build` to be very simple (equal to `dune build @target`) and take full advantage of dune incremental builds.
 
Using variants, less packages will need (unix-freestanding-xen) specialization but this requires an update of packages that have C stubs.

One good indicator of this working fine is the `mirage/mirage-dev#dune` CI on mirage-skeleton. It doesn't pass yet but it should improve as we dufiny packages.

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -1,4 +1,4 @@
 (library
   (public_name mirage)
   (wrapped     false)
-  (libraries ipaddr logs astring functoria functoria.app mirage-runtime bos))
+  (libraries ipaddr logs astring functoria functoria.app mirage-runtime bos sexplib))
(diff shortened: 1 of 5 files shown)


Drup:
Looks promising! 

I'm still hoping that someone will figure out the build as part of functoria, which a clean separates between the "generic" part and the "mirage" part. It's not necessary at all, but since you are actively working on the topic, the idea can sit in the corner of your mind. ;)

dinosaure:
Works of @TheLortex was integrated in an other way (and according current status of master) in this PR: #1020

PR #980, opened 2019-04-08
work on the log reporter

hannesm:
how it used to be in any case:
current

now, what I wanted to have are colours... on Unix to check whether the output terminal is a usual one, or redirected to a log file, or a dumb terminal.

now, the same unikernel with this patchset on unix:
new-default-unix

when redirected to a file:
new-default-unix-redirect

solo5 target:
new-default-solo5

now, modifying `config.ml` to include in the `register` call: `~reporter:(default_reporter ~timestamp:false ())` leads to (on unix):
new-timestamp-false

and ``~reporter:(default_reporter ~timestamp:false ~style:`Ansi_tty ())`` compiled for solo5:
new-timestamp-false-style-ansi-tty-solo5

advantages:
- the user experience changes: unix unikernels reporting to standard output have colors now
- when the output is redirected to a file, there aren't color escape sequences
- you can explicitly enable colours for freestanding targets
- you can disable the timestamp ;)
- the mirage-logs library and its ring buffer is only used as dependency if the `ring_buffer` argument is passed to `default_reporter`

downsides are:
- tags are not printed atm
- log source name is not printed
- we should pass `style` as boot parameter (maybe `timestamp_prefix` as well?)

I guess to remove redundancy, the reporters should be part of mirage-logs..

### DIFF
--- a/lib/mirage.mli
+++ b/lib/mirage.mli
@@ -98,13 +98,15 @@ val reporter: reporter typ
 (** Implementation of the log {!reporter} type. *)
 
 val default_reporter:
-  ?clock:pclock impl -> ?ring_size:int -> ?level:Logs.level ->
-  unit -> reporter impl
-(** [default_reporter ?clock ?level ()] is the log reporter that
-    prints log messages to the console, timestampted with [clock]. If
-    not provided, the default clock is {!default_posix_clock}. [level] is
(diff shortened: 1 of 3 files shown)


hannesm:
this is not yet ready, and feels slightly wrong.... but it gets us colourful output :) (maybe another approach will find us soon)

PR #981, opened 2019-04-13
remove unused Dockerfile.doc (done by https://github.com/mirage/docs)

hannesm:
refresh travis.yml slightly (CI should be faster now)

@avsm IIUC the Dockerfile.doc is not used anymore. what about the Dockerfile?


avsm:
I've unhooked ci/dockercloud from the build so this will stop failing. Docs.mirage.io builds are now driven from https://github.com/mirage/docs and so Dockerfile.doc is no longer needed.  I'd suggest removing Dockerfile as well for now and I can reinstate it closer to the mirage4 release with the latest image bases.

hannesm:
@avsm thanks for your information! I stumbled upon it while searching for "charrua-core" (which soon will be known as "charrua"). I updated this PR to remove Dockerfile as well as .opam-config-exec.

.travis.yml
```diff
@@ -1,22 +1,20 @@
 language: c
 install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-docker.sh
 script: bash -ex .travis-docker.sh
-sudo: required
 services:
   - docker
 sudo: false
 env:
  global:
-   - PRE_INSTALL_HOOK="cd /home/opam/opam-repository && git pull origin master && opam update -u -y"
```
avsm: I think this is still useful -- why remove it?

hannesm: because it is already done in [.travis-docker.sh](https://github.com/ocaml/ocaml-ci-scripts/blob/4466db9737c1e615d4aab631bb32b3505ea68ce6/.travis-docker.sh#L78)

PR #983, opened 2019-04-29
Qrexec expose

linse:
We don't automatically invoke gui and qrexec for -t qubes anymore.
In QubesOS 3.2 it was necessary to start default handlers for all VMs,
otherwise QubesOS would assume the VM was not working properly and kill it
after 60 seconds.  This is configurable in QubesOS 4, so now that
QubesOS 3.2 is past end-of-life, don't do it by default.
/cc @yomimono

### DIFF
--- a/lib/mirage.mli
+++ b/lib/mirage.mli
@@ -53,6 +53,9 @@ val qubesdb: qubesdb typ
 val default_qubesdb: qubesdb impl
 (** A default qubes database, guessed from the usual valid configurations. *)
 
+type qrexec
+val qrexec: qrexec typ
+val default_qrexec: qrexec impl
 
 (** {2 Time} *)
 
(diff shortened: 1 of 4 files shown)


lib/mirage.mli
```diff
@@ -53,6 +53,9 @@ val qubesdb: qubesdb typ
 val default_qubesdb: qubesdb impl
 (** A default qubes database, guessed from the usual valid configurations. *)
 
+type qrexec
```
talex5: Some doc-comments are needed here. Most people reading this aren't going to know what qrexec (or Qubes) is.

yomimono:
> I wonder if we should go further - do you think it's useful to keep qrexec support in the mirage tool at all? It seems like it can't select something else with the same interface for e.g. -t unix.

Hm, that's a good point.  Less being more and all, I think that's the way to go.

yomimono:
(On further thought I think @linse suggested this too, and I had some half-baked idea about it being useful to have setup done before the unikernel starts, but after having slept on it I don't think that logic is very sound.)

talex5:
Closing in favour of #984 - thanks!

PR #984, opened 2019-04-30
remove autostart for qrexec and qubes gui

yomimono:
We don't automatically invoke gui and qrexec for -t qubes anymore.
In QubesOS 3.2 it was necessary to start default handlers for all VMs,
otherwise QubesOS would assume the VM was not working properly and kill
it after 60 seconds. This is configurable in QubesOS 4, so now that
QubesOS 3.2 is past end-of-life, don't do it by default.

Stubs in mirage_impl_gui and _qrexec weren't easily usable by the unikernel,
so remove them; applications can use the setup code in mirage-qubes
directly.

Co-authored-by: linse <sschirme@gmail.com>


yomimono:
Alternative to #983 . /cc @linse

PR #985, opened 2019-05-04
New tracing API no longer requires Gnt and Xs arguments

talex5:
Needed once https://github.com/mirage/mirage-profile/pull/33 is released.


hannesm:
ah thanks for pointing here. I suspect we'll need a mirage-profile release first which includes the above fix, and adjust the generated lower bound for mirage-profile (in line 20 of mirage_impl_tracing).

talex5:
I've added the lower bound. This can be merged once https://github.com/ocaml/opam-repository/pull/14466 is in.

hannesm:
thanks @talex5

PR #988, opened 2019-05-16
require ocaml 4.05.0 or higher

hannesm:
see https://github.com/mirage/mirage-kv/commit/3df5b30f240f675d8d95c94440825e013b7cfc2f for further information

with ocaml 4.04.2, mirage itself could be compiled, but any generated configured unikernel (opam file) would require 4.05.0 due to the mirage-kv >= 2.0.0 dependency of mirage-types, where mirage-kv requires 4.05.0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ locally or use the Docker image.
 
 You will need the following:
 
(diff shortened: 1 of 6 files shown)


hannesm:
CI is green, merging

Issue #992, opened 2019-07-11
Solo5: Multiple devices support, spt target


mato:
This issue tracks the work being done to update MirageOS to the new Solo5 APIs with support for multiple devices and application manifest (Solo5/solo5#372). Additionally, these changes also enable and add support for the spt (seccomp) target to Mirage.

Current work in progress is available as an OPAM remote, at `git+https://github.com/Solo5/opam-solo5.git`, which includes the following changes:

- ~~solo5-bindings-*: Point to Solo5 `master`.~~ Now in OPAM as 0.6.2.
- mirage#993 ([branch](/mato/mirage/tree/solo5-manifest)): Add support for spt target, generate application manifest at configure time using information from config.ml and link into unikernel at build time.
- ~~mirage/mirage-solo5#46 ([branch](/mato/mirage-solo5/tree/solo5-manifest)): Update C stubs to new multiple-device APIs.~~ Now in OPAM as 0.6.0.
- ~~mirage/mirage-block-solo5#18 ([branch](/mato/mirage-block-solo5/tree/solo5-manifest)): Update to new multiple-device APIs.~~ Now in OPAM as 0.6.0.
- ~~mirage/mirage-net-solo5#32 ([branch](/mato/mirage-net-solo5/tree/solo5-manifest)): Update to new multiple-device APIs.~~ Now in OPAM as 0.6.0.

@mirage/core: Please review these changes and test the remote.

At least the following remains to be addressed:

1. ~~The Lwt main loop (in mirage-solo5:lib/main.ml) has not been updated to take multiple devices into account and wake up only those threads that have work ready. The current implementation works "by accident" even with multiple networks, but results in spurious wakeups. mirage-net-solo5 will also need to be updated to match any changes here.~~ Done, changes pushed to mirage#993 and mirage/mirage-net-solo5#32.
2. The meaning of `block_of_file X` is unclear for Solo5; I have modified the mirage tool to just pass `X` through verbatim as the Solo5 block device name, however this precludes having a single config.ml that would be portable to all targets without changes.
3. Should we also introduce a a `default_block`, similar to `default_network`, for the "single block device" case?
4. The code in `generate_manifest` in the mirage tool should check that device names are unique and alphanumeric (as per the Solo5 manifest [documentation](https://github.com/Solo5/solo5/issues/372).
[...]

PR #993, opened 2019-07-12
Solo5: "spt" target, application manifest generation, multiple devices support

mato:
These are the `mirage` front-end changes for #992.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -869,7 +869,18 @@ let build i =
   link i name target target_debug >>| fun out ->
   Log.info (fun m -> m "Build succeeded: %s" out)
 
+let clean_opam name target =
+  Bos.OS.File.delete (opam_file (unikernel_opam_name name target))
+
+let clean_binary name suffix =
+  Bos.OS.File.delete Fpath.(v name + suffix)
+
 let clean i =
+  let rec rr_iter f l =
+    match l with
+    | [] -> R.ok ()
+    | x :: l -> f x >>= fun () -> rr_iter f l
+  in
   let name = Info.name i in
   clean_main_xl ~name "xl" >>= fun () ->
   clean_main_xl ~name "xl.in" >>= fun () ->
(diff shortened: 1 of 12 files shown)


dinosaure:
LGTM

lib/mirage.ml
```diff
@@ -565,6 +565,42 @@ let unikernel_opam_name name target =
...
+  let open Codegen in
+  let file = Fpath.(v "manifest.json") in
+  with_output file (fun oc () ->
+      let fmt = Format.formatter_of_out_channel oc in
+      append fmt "{ \"version\": 1, \"devices\": [ %s ] }" s;
+      R.ok ())
+    "Solo5 application manifest file"
+
+let generate_manifest_c () =
+  let json = "manifest.json" in
```
cfcs: Why do we not put this in `_build/` as well?

mato:
Seeing as I've done all the heavy lifting here, could someone else from @mirage/core @mirage/devs please figure out what to do about `block_of_file` so that we can get `mirage-skeleton` to compile? From my questions on Slack today:

I need some help with Functoria for mirage/mirage#993 - @samoht or @drup perhaps? already asked Hannes but he doesn't have a solution. (Thread follows) (edited) 

In order to get mirage-skeleton to compile with the new Solo5 changes, I need to pass a different value to block_of_file depending on whether or not we are configuring for Solo5.

This is point (2) mentioned here https://github.com/mirage/mirage/issues/992.

I've tried adding an `is_solo5` to `Mirage_key`, and so far so good:

```ocaml
val is_solo5: bool value
(** Is true iff the {!target} key is a Solo5-based target.
*)
```

```ocaml
let is_solo5 =
 | #mode_solo5 -> true
 | #mode_xen | #mode_unix -> false
```

However, I can't figure out how to sensibly use this function in device-usage/block's config.ml. I've tried both of the following:

```
27 | let img = block_of_file (Key.(if_ is_solo5) "storage" "disk.img")
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Error: This expression has type
         string Mirage.Key.value = string Functoria_key.value
       but an expression was expected of type string
```

and

```
27 | let img = block_of_file (if Key.is_solo5 then "storage" else "disk.img")
                                 ^^^^^^^^^^^^
Error: This expression has type
         bool Mirage.Key.value = bool Functoria_key.value
       but an expression was expected of type bool
       because it is in the condition of an if-statement
```

samoht:
@mato You can write this:

```ocaml
let img = Key.(pure block_of_file $ if_ is_solo5 "storage" "disk.img")
```

But this will make `img` something of type `block impl value` and might need to be propagated to the various use points.

samoht:
After looking at device-usage/block's config.ml, what you want is probably more something like this:

```
let img = Key.(if_impl is_solo5 (block_of_file "storage") (block_of_file "disk.img"))
```

mato:
@samoht I added `is_solo5` in 741ea6d. Feel free to push to mirage/mirage-skeleton and see if you can get the CI to pass here.

hannesm:
I took the initiative (here https://github.com/hannesm/mirage-skeleton/commit/1acfba506d1af4acd0a4670183e27081de27fc8a) and pushed a commit to this PR which uses that mirage-skeleton remote. let's see whether travis succeeds. if that is the case, I'll merge and cut a release!

hannesm:
with the most recent force-push a mirage-skeleton compiles fine within CI (see https://travis-ci.org/mirage/mirage/builds/592619722 for result) - the only issue left is qubes backend and device-usage/prng failing to install solo5-bindings-spt -- which I'll ignore for now (and in a later step fix the prng unikernel). thanks for this work!

PR #995, opened 2019-08-05
Bump mirage-xen constraint to < 5.0.0

reynir:
It should be fine to use mirage-xen 4.0.0.


hannesm:
I restarted CI after fixing mirage-skeleton, this looks good to me, thanks!

PR #996, opened 2019-08-14
revised resolver

hannesm:
this requires https://github.com/mirage/ocaml-conduit/pull/290 (which in turn requires a release of ocaml-dns, which will happen soon).

what this PR does:
- the Resolver_mirage.With_stack functor now requires a random device (to generate the identifier of the DNS packet) -- as usual, the `default_random` device is used (configurable via `--prng` configuration time parameter)
- two new command-line keywords are provided (both stages, configuration and runtime): `resolver` and `resolver-port` for configuring the resolver to use

### DIFF
--- a/lib/mirage_key.ml
+++ b/lib/mirage_key.ml
@@ -343,6 +343,14 @@ module V6 = struct
 
 end
 
+let resolver ?(default = Ipaddr.V4.of_string_exn "91.239.100.100") () =
+  let doc = Fmt.strf "DNS resolver (default to anycast.censurfridns.dk)" in
+  create_simple ~doc ~default Arg.ipv4_address "resolver"
+
+let resolver_port ?(default = 53) () =
+  let doc = Fmt.strf "DNS resolver port" in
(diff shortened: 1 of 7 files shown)


hannesm:
I updated this PR to reflect the recent conduit (renaming and version number changes) in https://github.com/ocaml/opam-repository/pull/14689

hannesm:
this is green now, I'll merge and release within the next 24h unless someone objects //cc @mirage/core

lib/mirage_key.ml
```diff
@@ -343,6 +343,14 @@ module V6 = struct
 
 end
 
+let resolver ?(default = Ipaddr.V4.of_string_exn "91.239.100.100") () =
+  let doc = Fmt.strf "DNS resolver (default to anycast.censurfridns.dk)" in
```
avsm: Could we expose this default as a value in the resolver library in the future, rather than duplicating it here?

hannesm: this sounds preferable, but I do not know how with the current mirage and functoria design -- anyone has an example somewhere how to do this?

PR #999, opened 2019-09-28
ocamlbuild: Respect verbosity, quiet by default

mato:
Make the ocamlbuild invocation respect the mirage tool's verbosity
setting. As the default log level is Warning, this will result in quiet
builds by default; mirage build --verbose can be used to get an
equivalent of the old behaviour.

/cc @mirage/core

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -627,8 +627,13 @@ let compile ignore_dirs libs warn_error target =
     | [] -> Bos.Cmd.empty
     | dirs  -> Bos.Cmd.(v "-Xs" % concat dirs)
   in
+  let want_quiet_build = match Logs.level () with
+  | Some Info | Some Debug -> false
+  | _ -> true
+  in
   let cmd = Bos.Cmd.(v "ocamlbuild" % "-use-ocamlfind" %
-                     "-classic-display" %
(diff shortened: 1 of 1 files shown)


hannesm:
this looks fine to me. the travis failure I'm not too sure about -- this is while installing `device-usage/prng` (which atm depends on mirage-solo5) on the qubes target. opam selected `solo5-bindings-spt 0.6.2`:
```bash
# cp bindings/spt/solo5_spt.o bindings/spt/solo5_spt.lds \
#     /home/opam/.opam/4.07/lib/solo5-bindings-spt
# cp: cannot stat 'bindings/spt/solo5_spt.o': No such file or directory
# GNUmakefile:96: recipe for target 'install-opam-spt' failed
```

unfortunately any concrete build error (why there's no solo5_spt.o) is not part of travis log. I'd be happy to merge this now.

mato:
> unfortunately any concrete build error (why there's no solo5_spt.o) is not part of travis log. I'd be happy to merge this now.

Without looking at the details, this is likely due to Solo5/solo5#418. I
will cut an 0.6.3 fixing this eventually, but in the mean time it shouldn't
be a showstopper.

PR #1002, opened 2019-10-03
travis: use alpine in hope this makes CI green

hannesm:
this reduces the variance of linux distributions to build, but it is not entirely clear to me which distributions / docker images have recent enough toolchains (for solo5). it is also a bit alien to me what are valid strings in the `DISTRO` environment variable...


hannesm:
maybe not too useful now that I managed to slim the "offending" unikernel in respect to their dependencies (see https://github.com/mirage/mirage-skeleton/pull/276/commits/217426f9e5ddb69a15ceac356cb838cb3c55024a)

Issue #1004, opened 2019-10-15
fewer opam packages


hannesm:
I intended to work on [cstruct with capabilities integration](https://github.com/mirage/ocaml-cstruct/pull/237) into MirageOS interfaces. A first attempt was to extend various interfaces with instead of a `type buffer` having a `type read_buffer` and `type write_buffer`. This does not scale. I thought a bit more about how the current ecosystem is hackable (i.e. easy to understand and find definitions). Now, while documentation is great, [this landing](https://mirage.github.io/mirage-protocols/mirage-protocols-lwt/Mirage_protocols_lwt/index.html) is a bit rough when I intend to understand "how is IP configured"?

So, going five steps backwards, maybe "abstract all the things" is not too useful in the end. As facts, I actually do not intend to execute a MirageOS unikernel with multiple concrete IO abstractions. I do not even intend to have a mirage release that is able to deal with multiple IO abstractions (lwt, core, effect system). The very nice work by @dinosaure on `Cstruct_cap` (where `type t` is abstract now) also led me to think that I do not really intend to have various `type buffer` which get disjoint concrete types. And to be honest, I find a single `Ipaddr.V4.t` (and `Macaddr.t`) sufficient -- and dozens of type aliases and equalities that need to be written down rather boilerplate than "neat system design". Furthermore, various signatures were completely abstract, but the `-lwt` package made not only the `type 'a io = 'a Lwt.t` concrete, but as well `type value = string` or `type ipaddr = Ipaddr.V4.t`.

This is the main motivation why I sat down for some time, folding the `-lwt` packages into the base ones, and making the interfaces more concrete (by no means I want to stick with lwt forever, I'm eager to switch to the upcoming effect system rather sooner than later -- but I'd suggest to switch over with a major release of mirage, and remove lwt in the same release). This patchset is not complete (if you want to try it out, add this opam remote https://github.com/mirage/mirage-dev/tree/easy and enjoy some sample unikernels at https://github.com/hannesm/mirage-skeleton/tree/easy) -- its sufficiently mature to successfully run mirage-tcpip tests, or compile a device-usage/network and tutorial/hello unikernel on hvt (the resulting binaries are only marginally smaller (maybe 1%) -- I did not measure the reduction in opam solver time due to the reduced package set). You can take a look (and compare with existing) at the [odig odoc generated documentation](https://berlin.ccc.de/~hannes/mirage-easy/html/).

[...]

PR #1006, opened 2019-10-20
adapt to 1004 changes of the interfaces

hannesm:
~~this is a draft until the interfaces are actually out there~~

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,5 +1,5 @@
 eval `opam config env`
 opam depext -uiy mirage
 cd ~
-git clone https://github.com/mirage/mirage-skeleton.git
+git clone -b easy https://github.com/hannesm/mirage-skeleton.git
 make -C mirage-skeleton && rm -rf mirage-skeleton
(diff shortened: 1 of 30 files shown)


hannesm:
this is ready for review, briefly what it does:
- bump version numbers of various packages (see #1004)
- revise Static_ipv4.connect to comply with tcpip 4.0.0 signature
- use `Nocrypto_entropy_mirage.initialize` on all platforms (also unix & macos) since the mirage-unix main loop changed -- it does not use Lwt_main.run anymore, thus the Lwt_main hooks are not executed (instead, mirage-runtime provides the hooks)
  what we "loose" here is the seeding from /dev/random on unix targets.

hannesm:
this https://github.com/ocaml/opam-repository/pull/15192 is blocking the release -- since #1010 (and the mirage-unix changes) moved away from Lwt_main.run / hooks, the entropy gathering on the unix target needs to be done via nocrypto_entropy_mirage / mirage-entropy --- and nocrypto's mirage support is at the moment only available if mirage-xen / mirage-solo5 is around (see the opam repo PR which fixes this)

PR #1009, opened 2019-10-23
mirage-entropy functorized

dinosaure:
Hi @mirage/core!

This PR is one patch to paving the way for the _dunification_ of MirageOS. But, at least, it rollbacks the MirageOS eco-system to something more homogeneous. Indeed, the choice of MirageOS is to take the advantage of _functors_ to be able to choose an implementations of an abstraction - like choose `irmin` or `crunch` as a read-only KV-store.

`functoria` serves this purpose and orchestrates application of _functors_ according definitions.

### The black duck

However, a part of the MirageOS decided to choose an other way to specialize an implementation from another by a non-conventional way: `mirage-entropy` with `mirage-os-shim`. These packages use the _linking-trick_:
- `mirage-entropy` expects __at the link time__ a CMX which follows CMI provided by `mirage-os-shim`
- `nocrypto.mirage` and `mirage-stdlib-random` are done on top of that
- a _dummy implementation_ exists to ensure that this CMX is not missed when we configure an _unikernel_ - to avoid any struggle at the build-time

According discussion on mirage/mirage-os-shim#8, the goal of this PR is to delete this non-explicit dependency path of the MirageOS eco-system and provide a new way to have a _random generator_ implementation.

### The solution

This new way is, of course, the _functor_ way. Random generator must initializes an entropy engine at the beginning of the process - so, just before `start`. `mirage-entropy` trusts on a CMI to have an access to `at_enter_iter` and fill a state used by the entropy engine.

However, as I said, implementation of this CMI provided by `mirage-os-shim` does not exist - not yet. Then, `mirage-os-shim` provided a specialization of implementations of targets (eg. `mirage-solo5`, `mirage-xen` & `mirage-unix`) under the same CMI.

At the end, `mirage-entropy` will be linked with the right specialization/implementation of `mirage-os-shim` (with an help of `META` file/`ocamlfind` - **hint** _dunification_) and will be have an access to `at_enter_iter`. However, `functoria` is not able to follow this kind of dependency - or precisely not conventionally.

The choice was made to provide a new interface `Mirage_main.S` (see mirage/mirage-os-shim#9) into the `mirage-os-shim` package and only an interface - so specialization of targets under this interface should not exist anymore.

Then, by this way, we are able to reverse the dependency graph between targets and `mirage-os-shim`. targets must depend on `mirage-os-shim` and implement this interface - the choice was made to provide a new sub-package `main`:
- mirage/mirage-solo5#53
- mirage/mirage-unix#13

`mirage-entropy` will be _functorized_ over this interface (mirage/mirage-entropy#44) and, by cascading, `mirage-random-stdlib` will be _functorized_ too (mirage/mirage-random-stdlib#2). `nocrypto.mirage` is not needed anymore when a new package [`mirage-entropy-nocrypto`](https://github.com/dinosaure/mirage-entropy-nocrypto.git) follows the same implementation but, again, in a _functorized_ way.

Random generators expect now an implementation which provides entry-points of the target - see `Mirage_impl_entry_points`. Of course, any random occurrence must be updated to, at least, choose the default random implementation (eg. `mirage-random-stdlib`) with the right _default_ target.

Finally, to ensure the existence of the CMX at the link time when we currently compile an _unikernel_, a _dummy implementation_ was made to track this dependency globally and ensure that CMX of the target to provide `at_enter_iter` exists at the link time.

Because we follow in this PR an usual pattern (with `functoria` and dependencies with `package`), this _dummy implementation_ is not needed anymore - I think. But I don't know real implication of this deletion.

### Status of this PR

At first, this PR wants to show what is going on currently and what we will do to start proper discussion of all of that when, even if it concerns few packages, update seems deep enough to avoid to click with blinded eyes.

Secondly, it breaks compatibility specially about `random`, so we need to take care across several others packages and still have a consistent eco-system. `pins` was not done, `versions` should be cleaned and we should start a release plan.

Third, `entry_points`/`mirage-os-shim`/`*.main` use bad names or non-explicit purposes names and we should find a consensus to be clear about all of that to avoid any mis-conception/mis-comprehension.

### _functor_ design, _linking-trick_ and _virtual-library_

As I said into mirage/mirage-os-shim#8, `virtual_libraries` provided by `dune` is interesting for us when `mirage` should be only an _interface provider_. However, the goal, at this time, is to _dunify_ MirageOS - so I prefer to post-pone this goal.

Design of `mirage-entropy`/`mirage-os-shim` is not bad and is an example of how to use `virtual_libraries` outside the scope of `dune` - and mirage/mirage-os-shim#8 provides a compatible way across packages of this idea independently of the build-system. However, this patch highlight how _linking-trick_ (at the end, or more generally we can talk about the _linking-trick_) is fragile specially with `functoria` (which was not developed with pattern in our mind).

It shows that, even if we can find good aspects of it, it's hard to follow what is really going on when I spend finally 2 weeks to really understand the dependency graph of all with one of my smallest _unikernel_ [`entropieur`](https://github.com/dinosaure/mirage-os-shinimy).

I think, we should stay far from all of that at the beginning and start a full documentation of all and may be a migration to this kind of design then, but I consider it as orthogonal to the _dunification_ of MirageOS first.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -54,11 +54,14 @@ type mclock = Mirage_impl_mclock.mclock
 let mclock = Mirage_impl_mclock.mclock
 let default_monotonic_clock = Mirage_impl_mclock.default_monotonic_clock
 
+type entry_points = Mirage_impl_entry_points.entry_points
+let entry_points = Mirage_impl_entry_points.entry_points
+let default_entry_points = Mirage_impl_entry_points.default_entry_points
+
 type random = Mirage_impl_random.random
 let random = Mirage_impl_random.random
 let stdlib_random = Mirage_impl_random.stdlib_random
 let default_random = Mirage_impl_random.default_random
-let nocrypto = Mirage_impl_random.nocrypto
 let nocrypto_random = Mirage_impl_random.nocrypto_random
 
 type console = Mirage_impl_console.console
--- a/lib/mirage.mli
+++ b/lib/mirage.mli
@@ -110,6 +110,12 @@ val no_reporter: reporter impl
 (** [no_reporter] disable log reporting. *)
 
 
+(** {2 Entry-points} *)
+
+type entry_points
+
+val entry_points : entry_points typ
+val default_entry_points : entry_points impl
 
 (** {2 Random} *)
 
@@ -119,13 +125,13 @@ type random
 val random: random typ
 (** Implementations of the [Mirage_types.RANDOM] signature. *)
 
-val stdlib_random: random impl
+val stdlib_random: entry_points impl -> random impl
 (** Passthrough to the OCaml Random generator. *)
 
-val nocrypto_random: random impl
+val nocrypto_random: entry_points impl -> random impl
 (** Passthrough to the Fortuna PRNG implemented in nocrypto. *)
 
-val default_random: random impl
+val default_random : ?entry_points:entry_points impl -> unit -> random impl
 (** Default PRNG device to be used in unikernels.  It defaults to {!stdlib}, but
     users can override using the {!Key.prng} command line argument.
 *)
@@ -305,7 +311,9 @@ type ipv6_config = {
 (** Types for IP manual configuration. *)
 
 val create_ipv4: ?group:string ->
-  ?config:ipv4_config -> ?random:random impl -> ?clock:mclock impl ->
+  ?config:ipv4_config ->
+  ?entry_points:entry_points impl ->
+  ?random:random impl -> ?clock:mclock impl ->
   ethernet impl -> arpv4 impl -> ipv4 impl
 (** Use an IPv4 address
     Exposes the keys {!Key.V4.network} and {!Key.V4.gateway}.
@@ -313,12 +321,15 @@ val create_ipv4: ?group:string ->
     in the ipv4 configuration record, if that has been provided.
 *)
 
-val ipv4_qubes: ?random:random impl -> ?clock:mclock impl ->
+val ipv4_qubes:
+  ?entry_points:entry_points impl ->
+  ?random:random impl -> ?clock:mclock impl ->
   qubesdb impl -> ethernet impl -> arpv4 impl -> ipv4 impl
 (** Use a given initialized QubesDB to look up and configure the appropriate
  *  IPv4 interface. *)
 
 val create_ipv6:
+  ?entry_points:entry_points impl ->
   ?random:random impl -> ?time:time impl -> ?clock:mclock impl ->
   ?group:string -> ethernet impl -> ipv6_config -> ipv6 impl
 (** Use an IPv6 address.
@@ -338,7 +349,9 @@ val udp: 'a udp typ
 val udpv4: udpv4 typ
 val udpv6: udpv6 typ
 
-val direct_udp: ?random:random impl -> 'a ip impl -> 'a udp impl
+val direct_udp:
+  ?entry_points:entry_points impl ->
+  ?random:random impl -> 'a ip impl -> 'a udp impl
 
 val socket_udpv4: ?group:string -> Ipaddr.V4.t option -> udpv4 impl
 
@@ -357,6 +370,7 @@ val tcpv6: tcpv6 typ
 
 val direct_tcp:
   ?clock:mclock impl ->
+  ?entry_points:entry_points impl ->
   ?random:random impl ->
   ?time:time impl ->
   'a ip impl -> 'a tcp impl
@@ -375,6 +389,7 @@ val stackv4: stackv4 typ
 (** Direct network stack with given ip. *)
 val direct_stackv4:
   ?clock:mclock impl ->
+  ?entry_points:entry_points impl ->
   ?random:random impl ->
   ?time:time impl ->
   ?group:string ->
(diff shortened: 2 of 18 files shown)


dinosaure:
> first, we can expose a Mirage_main function, which exposes only the scheduler hooks that Lwt_main is exposing (run, at_enter and at_exit). -- so using using the linking trick here should work and be legitimate. For names, Mirage_main is kind of similar to Lwt_main so it's ok, but Scheduler could also work (as it is exactly what this module is: hooks to interact with the thread scheduler).

This is the current solution used by `mirage-entropy` and mirage/mirage-os-shim#8 still continue to use it but provides a compatible solution with `dune` (eg. `mirage-unix`) and a plan to _dunify_ MirageOS - and, by this way, avoid orchestration by `ocamlfind`.

As @hannesm said, if we look into the MirageOS eco-system, this solution does not really fit smoothly when a _hack_ with `functoria` is needed (eg. deletion on `mirage_impl_random.ml`) and it's hard to really understand what is going on, like:
- where is `mirage_os.cmx`? (`mirage-os-shim`)
- who provides it? (`mirage-solo5`, `mirage-unix` or `mirage-xen`)
- what is the step to compile `mirage_os.cmx` with `entropy.cmx`? (`-no-keeps-locs` and `ocamlfind`)

So we can document it but the hack with `functoria` (which concerns `ocaml-conduit` at the end) still is needed. The hack is like: add a dependency between `nocrypto` and `nocrypto-random` (in terms of `functoria`) as dependency like application of functor `Nocrypto_random(Nocrypto)` but don't try to emit this application because I will do that by hands.

In other side, as I said, playing with the _linking-trick_ currently - with the current state of the MirageOS eco-system - is not so good when we have nothing (or, at least, `ocamlfind`) to orchestrate implementations under an interface. `digestif` is an example of that when a `_tags` is needed at the end to choose the right implementation - and it's not convenience to see a linking error otherwise.

But, as I said too, `dune` try to solve that and if we do the migration to `dune`, see the usability of `virtual_libraries` and `variants` into the MirageOS eco-system should definitely be a proper and a good solution. But in this goal, we must compile a _unikernel_ with `dune` first.

But at this stage, I would like to point an other incompatibility: **if we use `virtual_library` in `dune`, we don't have a way to trick or emit an other `META` file**. `mirage-os-shim` and `mirage-entropy` still work because they continue to use `topkg`. Into details, a move to `dune` should be fine for `mirage-entropy` but a move of `mirage-os-shim` to `dune` will be more tricky (orchestration, if we look into mirage/mirage-os-shim#8, will be done by the `mirage` tool - but again, as the `_tags` file for `digestif`, it's a hands-written orchestration).

> an other option (I am still not totally sure if it's a good idea) would just be to pass the scheduler around as a first class value. The only used function so far is at_enter_iter so that should be fairly easy to construct and pass around.

I'm more convince by this solution when we can generate a `main.ml` with the right `at_enter_iter` and pass it by values however. This solution will completely remove `mirage-os-shim` and `Mirage_random.S.initialize` will take an `at_enter_iter` as argument.

It still is a special case but as you said - and I agree with this - "we do not try to functorise over the scheduler implementation", The patch should applied only into `mirage-random-stdlib`, `mirage-entropy` and `mirage-entropy-nocrypto` at the end I think.

hannesm:
- TL;DR: let's leave mirage-os-shim alone and start a `mirage-scheduler` interface (with the four respective hooks `at_enter_iter/at_exit_iter/at_enter/at_exit`), release that, adapt mirage-entropy and the mirage utility, and get it out!

first of all: thanks @dinosaure, this is IMHO the way to go forward. some nitpicks:
- I like the name `scheduler` better than `main` as well (and now wondering whether Mirage_time -- providing the single `sleep_ns` function should be part of it? -- furthermore I had at various points the need for a `cron` like scheduling, i.e. `Lwt_engine.on_time` for MirageOS -- which would fit into a `scheduler` as well!) -- if we all agree on scheduler, this should live in its own repository (as our other module types)
- the `run` -- I'm no longer sure whether we should expose this or not, in the end we don't expect a user to care about it (neither call it ever)
- about functors vs first-class modules / linking trick: I like homogenous environments, functors are used everywhere in MirageOS, let's continue with that. first-class modules / linking trick comes out of nowhere and is hard to explain that I have a `Scheduler.on_enter` binding in my unikernel.ml -- this is esp. important if you think about tooling for developing unikernels (e.g. merlin) -- i don't quite understand how the first-class module approach would scale (for other devices -- which ones to have as first class modules, which ones as functor arguments) -- am happy to be convinced otherwise
- it's imho fine to have `run/bind/map` special (using whatever trick) since a user is not confronted with this - not supposed to call it. let's reduce the tricks we need in mirage/functoria and improve the tooling ;)
- while toying around with [noconfig](https://github.com/roburio/mirage-noconfig) (rough prototype, but already working on lots of mirage-skeleton examples), I'm more and more convinced that `config.ml` contains mainly redundant information (that can be derived from unikernel.ml), and we can improve usability (developer experience, merlin integration) quite a bit by even further using functors in unikernel.ml -- i.e. is there a reason why `boot arguments` and `app_info` are not done via functor arguments? -- both are atm rather unfriendly in my emacs since key_gen is generated and merlin won't know much about it -- but that's something to discuss elsewhere

PR #1013, opened 2019-11-01
Clean *.opam file when we configure a new target

dinosaure:
Delete all OPAM file when we `mirage configure` and avoid duplicate. Not sure if it's exactly the right direction.


hannesm:
this looks fine to me! would you mind to make it a proper PR?

hannesm:
thanks!

PR #1017, opened 2019-11-28
Factorize the way to configure/build/link an unikernel

dinosaure:
This PR wants to factorize `mirage.ml` to multiple files:

Configuration step:
- `mirage_configure_solo5`
- `mirage_configure_virtio`
- `mirage_configure_xen`
- `mirage_configure`

Build (and Link) step:
- `mirage_link`
- `mirage_build`

Clean step:
- `mirage_clean`

Some _misc_ functions are added into `mirage_impl_misc.ml` (instead to find them into `mirage.ml` with a non-obvious order) and all new files have an interface. Conventions are respected (eg. `*_path` instead `fn` or `*_file` to know where is files).

Nothing was added into this PR, it's only a factorization.

PS: in the way of the _dunification_, due to my local patch, I prefer to push this PR before to have something more clear about configure/build/link/clean steps.


hannesm:
I'm not a big fan of `_misc`. I opened #1019 which should unbreak CI (once merged, would you mind rebasing pls). apart from that, I believe you that no code got modified ;)

lib/mirage.ml
```diff
@@ -930,7 +205,7 @@ module Project = struct
   let packages = [package "mirage"]
 
   (* The directories to ignore when compiling config.ml *)
-  let ignore_dirs = ignore_dirs
+  let ignore_dirs = Mirage_configure_solo5.ignore_dirs
```
hannesm: this is dead code afaict (the `ignore_dirs` are used elsewhere but not in the Mirage module anymore)

lib/mirage_configure_solo5.ml
```diff
@@ -0,0 +1,69 @@
...
+  let pkg, _post = solo5_pkg target in
+  let c = "_build/manifest.c" in
+  let obj = "_build/manifest.o" in
+  cflags pkg >>= fun cflags ->
+  let cmd = Bos.Cmd.(v "cc" %% of_list cflags % "-c" % c % "-o" % obj)
+  in
+  Log.info (fun m -> m "executing %a" Bos.Cmd.pp cmd);
+  Bos.OS.Cmd.run cmd
+
+let ignore_dirs = ["_build-solo5-hvt"; "_build-ukvm"]
```
hannesm: you could move them to `Mirage_build` -- the only place they're used AFAICT

dinosaure:
> I'm not a big fan of _misc.

Me too, but I don't want to do a huge clean of the `mirage` tool.

lib/mirage.ml
hannesm: CI claims you still have this line in mirage.ml and it leads to compilation error..
```suggestion

```

hannesm:
sorry about the `ignore_dirs` -- looks like it was actually used and is required to satisfy some functoria_app interface

dinosaure:
Ok if nobody has some concerns about this PR, I will merge it this afternoon (/cc @mirage/core).

lib/mirage_configure.ml
```diff
@@ -0,0 +1,119 @@
...
+  let open Codegen in
+  let file = Fpath.(v "Makefile") in
+  with_output file (fun oc () ->
+      let fmt = Format.formatter_of_out_channel oc in
+      append fmt "# %s" (generated_header ());
+      newline fmt;
+      append fmt "-include Makefile.user";
+      newline fmt;
+      let depext = if no_depext then "" else "\n\t$(DEPEXT)" in
+      append fmt "OPAM = opam\n\
```
samoht: might be nice to use `{| ... |}` here

Issue #1018, opened 2019-11-28
Delete pkg-config


dinosaure:
Currently, `mirage` tool calls `pkg-config` to get some information about how to link a Solo5 _unikernel_. _dunification_ of MirageOS highlighted that `pkg-config` is, may be, not the right direction to do that when it chooses to use an other way (provide some files which contains flags - eg. mirage/ocaml-freestanding#66).

The question is not so easy where it's not clear what we really need (when we link an _unikernel_) and it can have a non-obvious impact into the MirageOS eco-system. At least, implications are not (yet!) fully knowable at this stage.

To re-contextualize:
The _dunification_ of MirageOS requires some information of:
- how to compile C stubs and be usable into an _unikernel_ (`cflags.sexp`)
- how to link a Solo5 unikernel (`ldflags`)
- with which libraries we need to link (`libs.sexp`)
- who will link the _unikernel_ (`ld`)

**NOTE**: all of these requirements are not fundamentally true. The _dunification_ is not yet done! But let's start with that.

However, `pkg-config` can provide this information. @hannesm denotes that with mirage/ocaml-freestanding#66, we will have 2 ways to get these information and we should delete one of them.

> In a Mirage3 world (ocamlbuild), pkg-config and .pc files are used to pass around cflags/ldflags/.. . Now, in a Mirage4 world (dune), the same information is embedded in separate files, and pkg-config is still around, called at installation time of solo5 + ocaml-freestanding. Is this correct?

My concerns is that `pkg-config` is more expressive that simple files when the `Requires` field appear into a `*.pc` file - so expansion of these information can be more complex. However, according mirage/ocaml-freestanding#66, all information come from a call of `pkg-config` (with a post-process with `sed`).

In other side, `mirage` already integrates a call to `pkg-config`. So, from what I know about the _dunification_ of MirageOS and what we have, we can move forward without mirage/ocaml-freestanding#66.

@mato, then, described where `pkg-config` is really needed:

> As far as I can tell, the dependencies on pkg-config in the whole Mirage(/Solo5) stack are:
> 
> - `solo5-bindings-*`
>     - `ocaml-freestanding`
>         - `mirage-solo5`
>             - (indirectly) the `mirage` tool and generated _unikernel_ build system
>         - `gmp-freestanding`
>         - `zarith-freestanding`
>         - (unclear, indirect?) `nocrypto` / `mirage-entropy`

Then, a consensus from these people is clear:
[...]

PR #1010, opened 2019-10-26
Add a Hooks submodule to get callbacks when the lwt main loop runs

samoht:

PR #1020, opened 2019-12-04
dunification of MirageOS

dinosaure:
/cc @mirage/core and @TheLortex 

Hi all, this is the first **draft** of a _not-completely_ tested _dunification_ of MirageOS. First, this PR does not work but I think, I reached a step where we need to start to talk technically about the _dunification_ of MirageOS and when we should track minor updates.

### The Pull-Request

The PR is the smallest of what I can do (and #1017 helps me a lot). Each commit has a description about the goal and update only one file per one file the project to avoid a misleading of patches.

This PR **does not** have deletion - not yet. Some warnings (eg. 32) was added. The goal is to propose a smooth view about the _dunification_ of MirageOS.

This PR is **only** about the _dunification_. Behind this word, we can imagine a lot (`duniverse`, `variants` and so on) but the only goal is to use `dune` to compile an _unikernel_.

This PR is not yet fully tested. Currently, only [pasteur](https://github.com/dinosaure/pasteur) was tested - `pasteur` includes a large stack when it use `irmin.2.0.0`, `git` and `digestif` and `conduit` with `httpaf` (including `tls` and `nocrypto` even if we don't support HTTPS, shame on me). Only UNIX and Solo5 target are tested - I don't have a deployment process for Xen. Despite all this, `pasteur` still is a good PoC to test/improve _dunification_ of MirageOS.

This PR trusts on `dune.2.0.0`. Several features are used but not so really required. Small patches on some libraries needed by `pasteur` are needed because:
- they don't compile on `dune.2.0.0`
- they have the UNIX module

This PR should not expect anything else from `dune.2.0.0`

### Issues on the eco-system

MirageOS has a large eco-system where we can find some unusual packaging politics. When we are the ownership on these packages, it's fine to update them according the _dunification_ of MirageOS. However, it's not totally the case where some MirageOS _unikernels_ depend on some libraries outside the scope of the MirageOS organization.

According all of that, and even if we have this PR, we still have some issues on our eco-system. The goal of this PR and the _dunification_ in general is not to fix all of these issues. At least, we highlighted them. But, please, keep in your mind that the only goal of this PR is to tap `dune build` in his keyboard.

One of the biggest issue on the eco-system was fixed with #1010. I don't want to re-explain all the story about `mirage-os-shim`, however, as I said, this issue highlighted a weak about a documentation of: how to be MirageOS compatible? And we just discovered that we have many ways.

But an other issue still exists on this PR and was explained/discussed into #1018 about `pkg-config`. `dune` can provides some others ways to do some tricks at the link time (eg. variants) but __it's not the purpose of this PR__. However, we should start then to talk about all of that.

The last issue is about GMP and we will go a bit into details.

#### Zarith and GMP

The `zarith.cmxa` has an information used by `ocamlopt` then: `-lgmp`. If you want to compile an executable or an object with `zarith` and without `-noautolink`, `ocamlopt` will emit to `gcc` (at the link time) `-lgmp` (and then, `gcc` will take `libgmp.a` available into your `LD_LIBRARY_PATH`).

Of course, considering targets like Solo5 or Xen, `libgmp.a` available on your system does not fit into constraints expected by Xen or Solo5. GMP must be compiled with specific C flags.

This is the purpose of the package `gmp-freestanding` and `gmp-xen`. My only concern about this way is to rename `libgmp_freestanding.a` and `libgmp_xen.a` to `libgmp.a` and let the compiler to choose these implementations according the target. Considering flags, Solo5 and Xen can provide in some ways `-L$(opam config var lib)/gmp-{freestanding,xen}` as a dependency.

**NOTE**: that means for any compilation of any _unikernels_, `-L` shoud always be emitted by `mirage-os-xen` or `ocaml-freestanding` even if the _unikernel_ does not depend on `gmp`. However, `-lgmp` will be emitted only if we want to link with `zarith.cmxa`.

#### Versatility of the linking

The linking step is done by `gcc` currently in our OCaml world, the linking step is super versatile where static libraries are chosen depending of which path is available on the command-line (and the environment). The predictability of which library will be statically linked to our _unikernel_ is hard to follow/track.

A new tool was added into `mirage`: `cc-to-opt` or `cc-to-ocamlopt` which will takes C flags and do a translation of `-L` to `-I` and `-l` to `-cclib -l`. It wants to resolve static libraries and say that they exists according all given path by `-L`. Then, it re-order _options_ due to the precedence of `-L` over `-l` (`-LA -la -LB` and `-LA -LB -la` don't have the same behavior if `liba.a` is available into `A` and `B` - this behavior is pretty-close to the GMP case where `libgmp.a` is available into several directories).

At the end, the choice was made to let `ocamlopt` to __link__ the object file (for Xen/Solo5/Unix). Then, and `ld` command will be done according the target:
- `unix`: nothing to do, at least, an `ln -sfn`
- `solo5`: `ld` with Solo5 link script
- `xen`: `ld` with some options

#### Correctness of C stubs linked

The previous point want to serve an other issue about the expected flags used to compile C sources (like GMP or `digestif`). An _unikernel_ for Solo5 which links with `digestif.c` (compiled to be used by an UNIX executable) __can work__. That means that silently, we can link with the wrong `libdigestif.a` and the _unikernel_ still works.

The only way to verify that we link with the right C stubs is to check (from what I know) assembly code between your _unikernel_ and expected static library.

At this stage, no solution exists so. But, again, this PR highlighted this issue.

#### UNIX module and `-dontlink`

On the currant status of MirageOS, `ocamlbuild` is used with `-dontlink` to avoid any link with `unix.cmxa`. However, we lost this option if we want to use `dune`.

About that, the deletion of this option/feature highlighted several packages which are linked with `unix.cmxa` mostly because they want to be linked with `bigarray.cmxa` which contains `map_file` (an UNIX _syscall_). Again, some packages have this dependency again (like `sexplib`).

This is the purpose of [`bigarray-compat`](https://github.com/mirage/bigarray-compat.git) and we should use it at any of our projects, at least.

### Issue on `dune`

`dune` unlocks several features to improve the way to compile an _unikernel_. Some of them are not used now but they exist. The notably feature is, may be, the variant where `dune` can orchestrate the choice of which OCaml library we should use according the target (`freestanding` or `xen`).

At this stage, it's possible to use it when the generated `dune` file describe which variant it will use (with the `variant` field).

A proposition on Zarith was made by @TheLortex in this way. We can talk about that on an other issue/PR but the idea is to have a Zarith library which provides `freestanding` and `xen` which the right link with GMP.

Another feature is about the `forbidden_libraries` which is not yet used - but it can be. The purpose of it is to follow who want to link with a specific library and, in our case, `unix.cmxa`. This `dune`'s field can help us a lot to track the compatibility with MirageOS of our libraries.

The compilation of an _unikernel_ with `dune` unlocks an other way to develop an our MirageOS where we can start to use `duniverse` and have a self-contained workspace of all of what we need. In this way, `mirage` still continues to emit an OPAM file used by `duniverse` then to download all libraries needed by our _unikernel_.

And from that, `dune` can provide a well-defined workspace with right C flags according the target to let the user to use C stubs without the _hard_ plumbing needed when we want to package that to OPAM.

#### (dirs ...)

The only problem today about `dune` is: we can not use `(dirs ...)` to load some specific underlying directories into the `dune.build` file. I don't know why. The current trick is that `mirage configure` rewrite not-only `dune.build` but `dune` too to load correctly files needed to compile _unikernel_.

#### flags needed by the target

Currently, a proper way to get flags needed by the target (C flags for C stubs and linking flags for `ld`) is not well-defined. The current way chosen by Solo5 is to use `pkg-config`. The other way chosen by Xen is to provide into the distribution of `mirage-os-xen` some files needed by `dune` then.

An explanation is available here: mirage/ocaml-freestanding#66

This question is pretty-close to what we want to do with `pkg-config` (if we want to use it or remove it). Currently, we don't have an homogeneous way to get these information and we should start to discuss/document that.

### Conclusion

Let start to talk about the _dunification_ of MirageOS. Please, I really would like to stay focus on the real _dunification_ of MirageOS. I know that this PR can unlock several possibilities but keep in your mind that as long as this PR will not be merged, all other discussions/issues will be vain - because the predicate is to use `dune` first.

Next week, I will try to provide a CI with this PR to show that this PR works! But we can start to talk about details of the way to _dunify_ a MirageOS.

PS: thanks for reading all!

### DIFF
--- a/lib/dune
+++ b/lib/dune
@@ -1,4 +1,4 @@
 (library
   (public_name mirage)
   (wrapped     false)
-  (libraries stdlib-shims ipaddr logs astring functoria functoria.app mirage-runtime bos))
+  (libraries stdlib-shims ipaddr logs astring functoria functoria.app mirage-runtime bos sexplib))
--- a/lib/mirage_build.ml
+++ b/lib/mirage_build.ml
@@ -17,6 +17,7 @@ let check_entropy libs =
   else
     R.ok ()
 
+(*
 let compile ignore_dirs libs warn_error target =
   let tags =
     [ Fmt.strf "predicate(%s)" (backend_predicate target);
@@ -65,22 +66,27 @@ let compile ignore_dirs libs warn_error target =
   in
   Log.info (fun m -> m "executing %a" Bos.Cmd.pp cmd);
   Bos.OS.Cmd.run cmd
+*)
 
 let ignore_dirs = ["_build-solo5-hvt"; "_build-ukvm"]
+
+let cross_compile = fun _ -> None
+
+let compile target =
+  let target_name = Fmt.to_to_string Key.pp_target target in
+  let cmd = match cross_compile target with
+    | None -> Bos.Cmd.(v "dune" % "build" % ("@" ^ target_name))
+    | Some b -> Bos.Cmd.(v "dune" % "build" % ("@" ^ target_name) % "-x" % b) in
+  Log.info (fun m -> m "Executing %a" Bos.Cmd.pp cmd) ;
+  Bos.OS.Cmd.run cmd
 
 let build i =
   let name = Info.name i in
   let ctx = Info.context i in
-  let warn_error = Key.(get ctx warn_error) in
   let target = Key.(get ctx target) in
   let libs = Info.libraries i in
   let target_debug = Key.(get ctx target_debug) in
   check_entropy libs >>= fun () ->
-  compile ignore_dirs libs warn_error target >>= fun () ->
-  (match target with
-    | #Mirage_key.mode_solo5 ->
-        Mirage_configure_solo5.generate_manifest_c () >>= fun () ->
-        Mirage_configure_solo5.compile_manifest target
-    | _ -> R.ok ()) >>= fun () ->
-  link i name target target_debug >>| fun out ->
-  Log.info (fun m -> m "Build succeeded: %s" out)
+  compile target >>= fun () ->
+  link i name target target_debug >>| fun () ->
+  Log.info (fun m -> m "Build succeeded.")
--- a/lib/mirage_build.mli
+++ b/lib/mirage_build.mli
@@ -1,5 +1,4 @@
 open Rresult
 
 val ignore_dirs : string list
-val compile : string list -> string list -> bool -> Mirage_key.mode -> (unit, [> R.msg ]) result
 val build : Functoria.Info.t -> (unit, [> R.msg ]) result
--- a/lib/mirage_cc_to_opt.ml
+++ b/lib/mirage_cc_to_opt.ml
@@ -1,0 +1,167 @@
+open Rresult
+
+module Log = Mirage_impl_misc.Log
+
+let search directories library =
+  let rec go = function
+    | [] -> R.error_msgf "Library <%a> does not exist." Fpath.pp library
+    | x :: r ->
+      let path = Fpath.(x // library) in
+      Bos.OS.File.exists path >>= function
+      | true -> Ok path
+      | false -> go r in
+  go directories
+
+let resolve directories libraries =
+  Log.debug (fun m -> m "Start to resolve static libraries into: %a" Fmt.(Dump.list Fpath.pp) directories) ;
+  let rec go resolved = function
+    | [] -> R.ok (List.rev resolved)
+    | `Name library :: r ->
+      Log.debug (fun m -> m "Search -l%s into library directories" library) ;
+      let path = Fpath.(v ("lib" ^ library) + "a") in
+      search directories path >>= fun x -> go (x :: resolved) r
+    | `Filename path :: r ->
+      if Fpath.is_rel path
+      then ( Log.debug (fun m -> m "Search -l:%a into library directories" Fpath.pp path)
+           ; search directories path >>= fun x -> go (x :: resolved) r )
+      else
+        ( Log.debug (fun m -> m "Search -l%a into library directories" Fpath.pp path)
+        ; Bos.OS.File.exists path >>= function
+          | true -> go (path :: resolved) r
+          | false -> R.error_msgf "Library <%a> does not exist" Fpath.pp path ) in
+  go [] libraries >>= fun _ ->
+  Ok (directories, libraries)
+
+let is_opt s = String.length s > 1 && s.[0] = '-'
+
+let parse_opt_arg s =
+  let l = String.length s in
+  if s.[1] <> '-'
+  then
+    if l = 2 then s, None
+    else String.sub s 0 2, Some (String.sub s 2 (l - 2))
+  else
+    try
+      let i = String.index s '=' in
+      String.sub s 0 i, Some (String.sub s (i + 1) (l - i - 1))
+    with Not_found -> s, None
+
+let parse_lL_args args =
+  let rec go lL_args other_args = function
+    | [] -> R.ok (List.rev lL_args, List.rev other_args)
+    | "--" :: args -> R.ok (List.rev lL_args, List.rev_append other_args args)
+    | x :: args ->
+      if not (is_opt x)
+      then go lL_args (x :: other_args) args
+      else
+        let name, value = parse_opt_arg x in
+        match name with
+        | "-L" | "--library-directory" | "--library-path" | "-l" | "--library" ->
+          ( match value with
+            | Some value -> go ((name, value) :: lL_args) other_args args
+            | None -> match args with
+              | [] -> R.error_msgf "%s must have a value." name
+              | value :: args ->
+                if is_opt value
+                then R.error_msgf "%s must have a value." name
+                else go ((name, value) :: lL_args) other_args args )
+        | _ -> go lL_args (x :: other_args) args in
+  go [] [] args
+
+let to_cmdliner ~binary lL_args =
+  let res = Array.make (1 + List.length lL_args) "" in
+  res.(0) <- binary ;
+  List.iteri (fun i (k, v) -> res.(i + 1) <- Fmt.strf "%s%s" k v) lL_args ;
+  res
+
+let resolve directories libraries =
+  resolve directories libraries |> function
+  | Ok v -> `Ok (Ok v)
+  | Error (`Msg err) -> `Ok (Error (`Msg err))
+
+open Cmdliner
+
+let existing_directory =
+  let parser x = match Fpath.of_string x with
+    | Ok v when Fpath.is_dir_path v && Sys.is_directory x -> Ok v
+    | Ok v when Sys.is_directory x -> Ok (Fpath.to_dir_path v)
+    | Ok v -> R.error_msgf "Directory <%a> does not exist" Fpath.pp v
+    | Error err -> Error err in
+  let pp = Fpath.pp in
+  Arg.conv ~docv:"<directory>" (parser, pp)
+
+let library =
+  let parser x = match Astring.String.cut ~sep:":" x with
+    | Some ("", path) ->
+      ( match Fpath.of_string path with
+        | Ok v when Fpath.is_abs v && Sys.file_exists path -> Ok (`Filename v)
+        | Ok v when Fpath.is_rel v -> Ok (`Filename v)
+        | Ok v -> R.error_msgf "Library <%a> does not exist" Fpath.pp v
+        | Error err -> Error err )
+    | Some (_, _) -> R.error_msgf "Invalid <namespec> %S" x
+    | None ->
+      ( match Fpath.of_string x with
+        | Ok v when Fpath.is_file_path v && Fpath.filename v = x -> Ok (`Name x)
+        | Ok v -> R.error_msgf "Invalid library name <%a>" Fpath.pp v
+        | Error err -> Error err ) in
+  let pp ppf = function
+    | `Filename path -> Fmt.pf ppf ":%a" Fpath.pp path
+    | `Name x -> Fmt.string ppf x in
+  Arg.conv ~docv:"<library>" (parser, pp)
+
+let path_of_libraries =
+  let doc = "Add directory $(i,dir) to the list of directories to be searched for $(b,-l)." in
+  let docv = "dir" in
+  Arg.(value & opt_all existing_directory [] & info [ "L"; "library-directory"; "library-path" ] ~doc ~docv)
+
+let libraries =
+  let doc = "Add the archive of object file specified by $(i,namespec) to the list of files to link. \
+             This option may be used any number of times. \
+             If $(i,namespec) is of the form $(i,:filename), $(b,ld) will search the library path for a file called $(i,filename), \
+             otherwise it will search the library path of a file called $(i,libnamespec.a)." in
+  let docv = "namespec" in
+  Arg.(value & opt_all library [] & info [ "l"; "library" ] ~doc ~docv)
+
+let cmd =
+  let doc = "cctoopt" in
+  let exits = Term.default_exits in
+  let man = 
+    [ `S Manpage.s_description
+    ; `P "Transform a $(i,ld)/$(i,cc) command-line to an $(i,ocamlopt) command-line." ] in
+  Term.(ret (const resolve $ path_of_libraries $ libraries)),
+  Term.info "cctoopt" ~doc ~exits ~man
+
+let null_buffer = Buffer.create 0x100
+let null = Format.formatter_of_buffer null_buffer
+
+let process_argv argv =
+  parse_lL_args (List.tl (Array.to_list argv)) >>= fun (lL_args, other_args) ->
+  let argv = to_cmdliner ~binary:argv.(0) lL_args in
+  ( let res = Cmdliner.Term.eval ~help:null ~err:null ~catch:false ~argv cmd in
+    Buffer.clear null_buffer ; match res with
+    | `Error (`Exn | `Term | `Parse) ->
+      R.error_msgf "Error to process command-line: %a." Fmt.(Dump.array string) argv
+    | `Ok (Ok v) -> Ok (`Do (v, other_args))
+    | `Ok (Error (`Msg err)) -> R.error_msg err
+    | `Version | `Help -> assert false (* XXX(dinosaure): TODO! *) )
+
+let with_I path =
+  [ "-I"; Fpath.to_string path ]
+
+let with_cclib_l = function
+  | `Filename path -> [ "-cclib"; "-l:" ^ Fpath.to_string path ]
+  | `Name name -> [ "-cclib"; "-l" ^ name ]
+
+let run_with_binary argv =
+  if Array.length argv = 0 then Fmt.invalid_arg "run_with_binary: must have, at least, one argument" ;
+  Log.debug (fun m -> m "Start to ocamlify: %a" Fmt.(Dump.array string) argv) ;
+  process_argv argv |> function
+  | Ok (`Do ((directories, libraries), other_args)) ->
+    let a = List.(concat (map with_I directories)) in
+    let b = List.(concat (map with_cclib_l libraries)) in
+    let c = other_args in
+    Ok (a @ b @ c)
+  | Error err -> Error err
+
+let run ?(binary= "a.out") argv =
+  run_with_binary (Array.concat [ [| binary |]; argv ])
--- a/lib/mirage_cc_to_opt.mli
+++ b/lib/mirage_cc_to_opt.mli
@@ -1,0 +1,24 @@
+(** [cc to opt] modules is an {i ocamlifier} of a {i cc}/{i ld} command-line.
+
+    The goal of this module is to do a predictible translation from a {i cc}/{i ld}
+    command-line to an {i ocamlopt} command-line. It converts:
+    {ul
+    {- [-l] option to [-cclib -l] option}
+    {- [-L] option to [-I] option}}
+
+    Due to the pervasive and versatile context of the linker, it could be hard to
+    follow which static libraries will be linked with our {i unikernel}. This module,
+    while it formats options, resolves absolute path of static libraries - and checks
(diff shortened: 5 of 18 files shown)


samoht:
Nice work!

> The only problem today about dune is: we can not use (dirs ...) to load some specific underlying directories into the dune.build file. I don't know why. The current trick is that mirage configure rewrite not-only dune.build but dune too to load correctly files needed to compile unikernel.

Do you have an example where you need this? This is supposed to be working fine I guess.

I'll have a proper review tomorrow.

avsm:
I like the overall approach.  As a quick response, I think that with:
> And from that, dune can provide a well-defined workspace with right C flags according the target to let the user to use C stubs without the hard plumbing needed when we want to package that to OPAM.

That this should never happen:
> An unikernel for Solo5 which links with digestif.c (compiled to be used by an UNIX executable) can work.

If we just compile every target in a well defined workspace, then dune caching in dune 2.0 can make it as fast, without any dangerous sharing involved.

dinosaure:
> Do you have an example where you need this? This is supposed to be working fine I guess.

By facts, we need to put a `(dirs ...)` into `dune` instead `dune.build` when we want to have a Solo5 _unikernel_. `(dirs ...)` wants to integrate some generated files by `mirage configure -t hvt` available in `.mirage.solo5` directory - so you can view `(dirs .mirage.solo5)`. These flags are `cflags.sexp`, `lflags`, etc.

Into an higher-level, the question is already opened on mirage/ocaml-freestanding#66 where:
- in this PR, `mirage` will do the call to `pkg-config` to get flags
- in the `ocaml-freestanding` PR, the `ocaml-freestanding` will provides these files

As @hannesm said, the `ocaml-freestanding` PR finally propose 2 ways (`pkg-config` and files) to get flags. May be it's wrong. In other side, `mirage-os-xen` provides files and the `mirage` tool use them - but it can, in the same way than a Solo5 target, generate by itself these files.

avsm:
I do prefer having explicit files with the flags rather than invocations to `pkg-config`. The reason is that this makes tracking build systems rules far easier, instead of the environment variables we have to propagate for pkg-config. I'm happy to port the mirage-xen pieces to install files with flags (some already are installed, but possibly not completely).

dinosaure:
> I do prefer having explicit files with the flags rather than invocations to pkg-config. The reason is that this makes tracking build systems rules far easier, instead of the environment variables we have to propagate for pkg-config. I'm happy to port the mirage-xen pieces to install files with flags (some already are installed, but possibly not completely).

Ok good, from what I know, `mirage-os-xen` does not need an update about this point. If we want to follow this way about `ocaml-freestanding`, mirage/ocaml-freestanding#66 need to be merged. By this way, `(dirs ...)` will not be a problem!

It seems that this last way can remove all occurrence of `pkg-config` into the mirage tool - however, we still to figure out about libraries which has C stubs and want to be packaged into OPAM.

avsm:
> we still need to figure out about libraries which has C stubs and want to be packaged into OPAM

Yes, mostly this the xen support libraries, and in those case we adapt them all via files containing CFLAGS or some `linkopts` META file as you suggest.

dinosaure:
I did a new branch `clean-dunification-without-pkg-config` available on my repository and diff is available here: #1021. Please I would like to continue the discussion here, I just would like to see the status of Travis and give you an easy way to see the patch.

From what I did, if we integrate mirage/ocaml-freestanding#66, we mostly can delete any use of `pkg-config` into the mirage tool which can be good but if we do the choice, we will be stick on a less expressive mechanism to get right flags and static libraries.

At this stage and to avoid an overlap with #1018, do we want to use files or `pkg-config` to shape the _dunification_ of MirageOS. We can imagine that the design with files works (but limitation can be reach easily, I think, in a near future). From what I know, nobody has a strong opinion on that, but we should keep in our mind that this solution can come with:
- a double maintenance cost (with `pkg-config` for libraries, and files for the `mirage` tool)
- a problem with we want to tune compilation of an _unikernel_

dinosaure:
@samoht highlighted some problems about the linking:
- [ ] `ocplib-endian` comes with `unix` (we should use https://gihtub.com/dune-universe/ocplib-endian.git#duniverse-1.0` (some others packages have this problem like `sexplib`)
- [ ] `cstruct.5.0.0` must be used (old version comes with `unix`)
- [ ] `sexplib` should not be a dependency where it was used only to pretty-print _s-expression_
- [ ] `pkg-config` should be removed when underlying packages like _targets_ should not be available first - in other words, `opam install mirage && mirage configure -t hvt` should work (eg. #1021 for implementation)

About dependencies on `unix`, I will add `(forbidden_libraries unix)` into `dune.build` generated for Solo5 and Xen targets to give you a proper error message (instead a link error).

dinosaure:
`forbidden_libraries` was added into the generated `dune.build` file for Xen and Solo5 target here: af028f9 and it will be cherry-pick into this PR. The PR without `pkg-config` is better when the suppression of `pkg-config` is quasi-total, only Xen was not fixed about that.

The change is available in this patch: d18c02b.

Into details, flags provided by `ocaml-freestanding` is not very consistent where `libmirage-solo5_bindings.a` must be added into `libs.sexp` even if this static library __is not__ a part of `ocaml-freestanding` (cc @mato). So I updated them to add `libmirage-solo5_bindings.a`, then `dune.build` with a specific rule copies the static library according the environment (OPAM environment or `dune` environment - the second is priority).

A `mirage configure && mirage build` or a `mirage configure -t hvt && mirage build` works with `mirage-skeleton` with OCaml 4.07.1.

dinosaure:
Ok, good news, from `clean-dunification-without-pkg-config`, only 2 packages need an update and `MODE=hvt make` into `mirage-skeleton` works :tada: !

- [ ] `nocrypto` depends on `bigarray`: I did a new branch into dinosaure/ocaml-nocrypto#1 with the smallest patch. According current status of `nocrypto`, I advise to _add an other patch_ to `opam-repository`.
- [ ] `sexplib` depends on `bigarray`: I did janestreet/sexplib#40
- [ ] `ocplib-endian` depends on `bigarray`.

mato:
@dinosaure: Thanks for working on this. It's a complex change with many moving pieces, and I'm not sure I understand all the changes you are making. I'm going to try and reply to the various points you've raised, but will also ask questions that may have already been answered elsewhere, or seem "obvious" to others, so please bear with me.

> This is the purpose of the package gmp-freestanding and gmp-xen. My only concern about this way is to rename libgmp_freestanding.a and libgmp_xen.a to libgmp.a and let the compiler to choose these implementations according the target.

If I understand this correctly, you are proposing that:

For a _package_ which requires specially ("cross") compiled C libraries and a _target_, where target is, for example, _xen_ or _freestanding_ (a.k.a. Solo5):

_Instead of_ _package_ providing a library `lib<package>_<target>.a` and mucking about with META files to force the linker to use the non-standard library name, _package_ will provide `lib<package>.a` in some _target_-specific location and _something_ (possibly ocaml-freestanding, but see below) will provide information to _mirage_ needed to set the linker's library search path to use _only_(!) libraries from the _target_-specific location(s).

Correct?

> Considering flags, Solo5 and Xen can provide in some ways -L$(opam config var lib)/gmp-{freestanding,xen} as a dependency. NOTE: that means for any compilation of any unikernels, -L shoud always be emitted by mirage-os-xen or ocaml-freestanding even if the unikernel does not depend on gmp

Without `pkg-config` we lose the ability to compose `-L` search paths -- what you write above seems to require ocaml-freestanding to know about every possible _package_ in order to be able to emit the `-L$(opam config var lib)/<package>`. That does not scale, and is the biggest limitation of dropping pkg-config altogether.

As an alternative, could we instead have_package_-_target_ install its libraries into a _shared_ directory, such as `$(opam config var lib)/mirage-<target>/lib`, and this would be added to `-L`/provided to _mirage_ by some suitable package? I write "some suitable" deliberately as this is probably not ocaml-freestanding's job, but should instead be done by `mirage-<target>`, i.e. mirage-xen or `mirage-solo5.

Related to this:

> Into details, flags provided by ocaml-freestanding is not very consistent where libmirage-solo5_bindings.a must be added into libs.sexp even if this static library is not a part of ocaml-freestanding (cc @mato). So I updated them to add libmirage-solo5_bindings.a, then dune.build with a specific rule copies the static library according the environment (OPAM environment or dune environment - the second is priority).

That's because `libmirage-solo5_bindings.a` comes from a pkg-config `.pc` file: https://github.com/mirage/mirage-solo5/blob/master/lib/bindings/mirage-solo5.pc which is then used by `mirage` as the actual package queried from pkg-config: https://github.com/mirage/mirage/blob/master/lib/mirage_link.ml#L8. `mirage` itself currently does not ask for ocaml-freestanding explicitly, only as a pkg-config requirement of mirage-solo5.

Could the solution I write about further above, using a single, _shared_ target-specific directory for C libraries be made to work for `mirage-solo5` also? How does `dune` determine the name of the C library it will link in? For a (OCaml? OPAM?) package _foo_, is it always `lib<foo>.a`?

> A new tool was added into mirage: cc-to-opt or cc-to-ocamlopt which will takes C flags and do a translation of -L to -I and -l to -cclib -l. It wants to resolve static libraries and say that they exists according all given path by -L. Then, it re-order options due to the precedence of -L over -l (-LA -la -LB and -LA -LB -la don't have the same behavior if liba.a is available into A and B - this behavior is pretty-close to the GMP case where libgmp.a is available into several directories).

Three questions:

1. Do we need this tool at all? Couldn't ocaml-freestanding or mirage-solo5 generate the information required in a format already suitable for passing to `ocamlopt`?
2. Relying on directory precedence only in the presence of multiple libraries with the same name is **incredibly fragile** and will come back to bite us in the future. Can't we have one and _one only_ specific search path per xen/freestanding/...?
3. If you are using `ocamlopt` to link instead of `ld`, does that mean you are no longer passing `-nostdlib`? That would be bad, see above. The unikernel must **not** accidentally link in host libraries.

> The previous point want to serve an other issue about the expected flags used to compile C sources (like GMP or digestif). An unikernel for Solo5 which links with digestif.c (compiled to be used by an UNIX executable) can work. That means that silently, we can link with the wrong libdigestif.a and the unikernel still works.

Again, this seems like almost but not quite the same issue as above? Sorry if I'm repeating myself, but it's important. We need a way to ensure that the toolchain is using `-nostdlib` and an explicit target-specific list of search path(s) _only_. Is this not possible?

dinosaure:
### Link step

As @samoht asked, I delayed the resolution of static libraries compiled with specific flags according the target to `mirage build` (instead `mirage configure`). A command-line like:

```sh
$ opam install mirage && mirage configure -t hvt
```

Should works. However, `make depends` still is needed before a `mirage build`. Computation still uses `cc-to-opt` small pass to correctly be more predictable about which static libraries will be linked. However, this pass still be on top of what `META` files give to us.

A new file appears under the hood `extra-libraries.sexp` which is empty at the beginning and then was filled by `pkg-config --static --libs %{target}` and `ocamlfind query -predicate native -format '%d %(%{target}_linkopts)` to still be happy with the current status of the MirageOS eco-system.

In others words, `pkg-config` and `ocamlfind` still are needed.

A new flag was added (`-noautolink`) into the link command only for Solo5 and Xen target when `ocamlopt` __must not__ follows `-l` flags given by `*.cm{,x}a` and takes them from `META` files. So from the _old_ world (`nocrypto`, `digestif.0.74` and `checkseum.0.0.9`), all works. I still need to figure out the new world however (specially about `digestif` and variants __and__ others provided `*.a`).

A clear limitation is that for any library which uses a C stub and do not get right information on `META` files, we will have a _link_ error about some symbols due to `-noautolink`.

dinosaure:
> Instead of package providing a library lib<package>_<target>.a and mucking about with META files to force the linker to use the non-standard library name, package will provide lib<package>.a in some target-specific location and something (possibly ocaml-freestanding, but see below) will provide information to mirage needed to set the linker's library search path to use only(!) libraries from the target-specific location(s).
> 
> Correct?

Correct at the beginning, not needed now because we use `-noautolink`. This is the way chosen before and `ocamlopt` with `-output-complete-obj` wants to follow flags given by `*.cm{,x}a` (like `zarith.cmxa` informs to put `-lgmp` at the link step).

Possible solution about that:
- `-noautolink` as I showed in my previous comment
- avoid the link by `dune`, I still continue to not fully understand the diff between `-output-obj` and `-output-complete-obj` and I suspect that the diff is probably the problem - and the main difference between `dune` and `ocamlbuild`.

> Without pkg-config we lose the ability to compose -L search paths -- what you write above seems to require ocaml-freestanding to know about every possible package in order to be able to emit the -L$(opam config var lib)/<package>. That does not scale, and is the biggest limitation of dropping pkg-config altogether.

`-L` flags if we look into details was currently given by 2 tools:
- `pkg-config` for Solo5 stuff (but Xen too)
- `ocamlfind` for libraries like `nocrypto` or `digestif`

About the current status of `mirage`, we still must have `ocaml-freestanding` for this purpose. My only update (even if it's little bit hard to follow) is to translate these information to some files (eg. mirage/ocaml-freestanding#66) according the `dune`'s format. The old part is available here: https://github.com/mirage/mirage/pull/1020/files#diff-34f96126d5975959270e9225bfd90fa0L29

Now, with `dune` I firstly moved that to the `configure` time and, from the request of @samoht, moved this step to the `build` time.

> As an alternative, could we instead have_package_-target install its libraries into a shared directory, such as $(opam config var lib)/mirage-<target>/lib, and this would be added to -L/provided to mirage by some suitable package? I write "some suitable" deliberately as this is probably not ocaml-freestanding's job, but should instead be done by mirage-<target>, i.e. mirage-xen or `mirage-solo5.

I think, you describe in some words what is the current problem. _layout_ of libraries with C stubs to be compatible with MirageOS is not documented. `nocrypto` has way and `digestif` has an other way. From this erratic eco-system, we currently want to move all of that to a `dune`'s _layout_ with variants - but still be compatible with old _layout_.

At the end, the story is only about how to correctly link an _unikernel_. But I will describe by through in another comment.

> Do we need this tool at all? Couldn't ocaml-freestanding or mirage-solo5 generate the information required in a format already suitable for passing to ocamlopt?

We currently need this tool not only for `ocaml-freestanding` and `mirage-solo5` but for libraries like `nocrypto` and `digestif` (or `bigstringaf`). So libraries outside the scope of MirageOS.

> Relying on directory precedence only in the presence of multiple libraries with the same name is incredibly fragile and will come back to bite us in the future. Can't we have one and one only specific search path per xen/freestanding/...?

I completely agree with you but the diff between now and what is proposed in this PR is to fix it a bit with what we can have from the eco-system. `dune` gives an other way but we need to keep in our mind that some packages will not be updated to the `dune`'s layout in a near future.

> If you are using ocamlopt to link instead of ld, does that mean you are no longer passing -nostdlib? That would be bad, see above. The unikernel must not accidentally link in host libraries.

`ld` still is used at the final step for Solo5 and Xen target and we use `-nostdlib` 👍 . The linker is used by `ocamlopt` to produce a `*.o` because `dune` emits `ocamlopt -output-complete-obj ...`.

dinosaure:
### Status of link step

So I just would like to clarify a non obvious situation about MirageOS.

Currently, most of problems highlighted by the _dunification_ of MirageOS is about the linking step. __This__ problem is currently fixed by __several__ ways:
- `pkg-config`
- `ocamlfind`
- the `cmxa` format

All of three can give to us information needed to correctly link libraries and C stubs/archives, etc. From what I know, the mostly popular way to get this information is `ocamlfind` with the `META` file but it does not fit correctly.

The `cmxa` format is the default way to get right C stubs/C archives, but it's possibly the wrong way (currently `nocrypto.cmxa` gives to us `-lnocrypto_stubs` by default, it's the same for `zarith.cmxa`).

`pkg-config` go further and can be more precise about what we needs to properly link our _unikernel_ but we have some concerns about this tool (cc @hannesm).

And, on top of that, we want to move to the `dune`'s variant _layout_.

### Concerns about what `mirage` does not do

In any way, we decided to let this task to the build system (currently to `ocamlbuild`/`ocamlfind`) but I'm not sure that is the right direction to do. MirageOS has a so specific link step and, even if I reach a status when `mirage-skeleton` is happy with `dune`, the cost is huge and I'm sure because we clearly use `dune` to a non-conventional way.

So I start to think about a `mirage link` to take care about the link step by ourselves with a deal with `ocamlopt` only. As @mato said, this part is incredibly fragile.

### Let's focus about the _dunification_

I'm not really happy with the status of https://github.com/mirage/mirage.git#clean-dunification-without-pkg-config but the underlying problem is not really about `dune` but about what the `mirage` should do on top of the build-system. But I will keep in my mind that and may be provide a better solution for all.

mato:
> I'm not really happy with the status of https://github.com/mirage/mirage.git#clean-dunification-without-pkg-config

Me neither. 

> but the underlying problem is not really about dune but about what the mirage should do on top of the build-system. But I will keep in my mind that and may be provide a better solution for all.

We should try and find a proper solution to the problem of "Mirage-compatible" packages, and we should do that now rather than later, since as far as I can tell the current state of the "dunification" just increases complexity and technical debt.

If we can replace pkg-config entirely and get other benefits such as a standardised way of building Mirage-compatible packages, we should consider such a solution "now" (with e.g. a "flag day" for the MirageOS release that includes it), _even if_ that means breaking compatibility with existing packages. Conversely, if the _tools_ (`dune`, `ocamfind`, ...) are not up to the task, we should also consider fixing the tools!

Given the complexity involved, the best way to progress on this would be to actually sit down together for a couple of days at a computer, possibly with some `dune` folks around, and figure it out. However, I don't see that happening this year, I already have too much on my plate with higher-priority work involving building systems that _use_ MirageOS.

Any ideas on how to progress this? @avsm @samoht @hannesm @dinosaure

mato:
Attaching today's discussion from Slack, so that we don't lose it.
[slack-log.txt](https://github.com/mirage/mirage/files/3944806/slack-log.txt)

PR #1022, opened 2019-12-08
embed proper build instructions and an install field into opam package for unikernel

hannesm:
in order to use opam tooling on unikernels, it is crucial to embed proper build and install instructions (atm: copy artifact to `%{prefix}%/bin` (I'm happy to improve the install location)

the build instructions now include a mirage configure -- with this, taking such a generated opam package, it only needs to have an embedded `src` section, and can be installable by opam.

The `binary_suffix` -- sorry I couldn't find whether this information is already available in mirage somewhere, neither did I bother to look what extensions are actually produced (apart from hvt and unix -- which better should install the actual binary and not a link to it).


hannesm:
I filled in the TODO from earlier commits, namely:
- discover git repository (and if successful, note origin and branch in opam file as source)
- discover git subdirectory and cd / prefix in build and install instructions
- collect and output artifact to install in bin and extra files to install into etc (libvirt, xl, xe)

while at it, I renamed mirage_configure_virtio to mirage_configure_libvirt to reflect what it is actually doing, and expose a `filename` function to avoid repeating `(name ^ "_libvirt") + "xml"` all over.

this is good to go (once CI passes) from my side.

PR #1025, opened 2019-12-18
Merge functoria back

samoht:
Rational: I have various plan to simplify/improve the API between functoria and mirage and having everything in one repository will make this kind of refactoring much easier. Also I don't know of any users of functoria outside of mirage, so I think we are making our life difficult here without valid reason.

/cc @Drup


hannesm:
to me this looks fine

avsm:
Fine by me.

PR #1027, opened 2019-12-19
Wrap library modules

samoht:


Drup:
I think this is a good idea, but it seems more complicated than initially planned. What's the issue ?

samoht:
Name clashes between `Functoria.Key` and `Mirage.Key`. I'm refactoring various things to make this go away.

samoht:
Closing, I'll upstream this on smaller patches.

PR #1028, opened 2019-12-20
configure, output opam: use git rev-parse instead of git branch --show-current (the latter requiring git 2.22+)

hannesm:
unfortunately I do not know since which git version the `rev-parse` is supported (pretty tricky to figure out). if this PR is fine, I plan to cherry-pick this commit on a 3.7.3 branch and cut a release 3.7.4 to "unbreak" mirage for people with old git installations.

the code in question is guarded - `find_git` is called in `configure_opam`, and if it errors this is handled gracefully, thus installations of 3.7.3 with an old git do _not_ lead to mirage that is not able to configure, just to mirage whose generated opam files are suboptimal and an error is presented on console (but return code of mirage is 0).

discussion how to get the current branch from git is at https://stackoverflow.com/questions/6245570/how-to-get-the-current-branch-name-in-git

### DIFF
--- a/lib/mirage_configure.ml
+++ b/lib/mirage_configure.ml
@@ -62,7 +62,7 @@ let find_git () =
   in
   Bos.OS.Dir.current () >>= fun cwd ->
   find cwd None >>= fun subdir ->
-  let git_branch = Bos.Cmd.(v "git" % "branch" % "--show-current") in
+  let git_branch = Bos.Cmd.(v "git" % "rev-parse" % "--abbrev-ref" % "HEAD") in
   Bos.OS.Cmd.(run_out git_branch |> out_string) >>= fun (branch, _) ->
   let git_remote = Bos.Cmd.(v "git" % "remote" % "get-url" % "origin") in
   Bos.OS.Cmd.(run_out git_remote |> out_string) >>| fun (git_url, _) ->


talex5:
In particular, it doesn't work on the latest Debian (10). Here's a Dockerfile to reproduce the problem:

```dockerfile
FROM ocurrent/opam:debian-10-ocaml-4.08
RUN git clone https://github.com/mirage/mirage-skeleton.git
RUN (cd ~/opam-repository && git pull && opam update)
RUN opam depext -i mirage
WORKDIR /home/opam/mirage-skeleton/tutorial/hello
RUN opam exec -- mirage configure
```

However, despite showing a huge red error, it doesn't return an error code and appears to generate the `main.ml` file.

hannesm:
@talex5 as I mentioned in the comment above:

> installations of 3.7.3 with an old git do not lead to mirage that is not able to configure, just to mirage whose generated opam files are suboptimal and an error is presented on console (but return code of mirage is 0).

does this PR (using `git rev-parse --abbrev-ref HEAD` instead) work in your debian docker?

talex5:
> does this PR (using git rev-parse --abbrev-ref HEAD instead) work in your debian docker?

Yes.

hannesm:
thanks for confirmation, I merged this PR into master and created a 3.7 branch with this commit on top of v3.7.3 tag. I released v3.7.4 from the 3.7 branch (https://github.com/ocaml/opam-repository/pull/15582) and pushed changes to both 3.7 and master branch.

PR #1030, opened 2019-12-21
Add --timestamp to fix the timestamp in generated files

samoht:
/cc @hannesm 

I know you don't need it for reproducibility but I need it in the upcoming expect tests. Would be nice if it could be useful for this too, though :-)


functoria/app/functoria_command_line.ml
```diff
@@ -51,6 +51,17 @@ let build_dir =
...
+let timestamp =
+  let doc =
+    Arg.info
+      ~docs:configuration_section
+      ~docv:"DATA"
+      ~doc:"The timestamp to use in generated files. If not provided, use \
+            the current date instead."
+      ["timestamp"]
+  in
+  Arg.(value & opt (some string) None & doc)
```
samoht: Do we have to protect against code injection here? Maybe we should only accept valid rfc3339 timestamps?

hannesm:
for reproducibility, I expected there'll be issues.. but it turns out the timestamp is only emit as comment into OCaml code, which does not end up in the binary. If you plan to emit the timestamp into a binary, please make sure to respect the `SOURCE_DATE_EPOCH` environment variable (as defined [by the reproducible builds project](https://reproducible-builds.org/specs/source-date-epoch/)).

hannesm:
I can see several ways forward:
- use an environment variable which contains the seconds since epoch (i.e. SOURCE_DATE_EPOCH);
- do not emit at all a timestamp into the code (even as comment, has it ever been turned out to be useful?) I'd imagine version information of the mirage tool is more useful for debugging/tracing;
- as proposed here, use an optional command-line argument.

I'm slightly biased towards 1 or 2.

functoria/app/functoria_command_line.ml
hannesm: as it looks atm, passing `--timestamp "*)"` will result in a `main.ml` which is not valid OCaml code

samoht: But you could pass `--timestamp "*) Sys.command \"echo toto\" ;; (*"`

samoht:
I like 2 :-) I'll change that patch to do that instead.

samoht:
Closing in favour of #1032

PR #1032, opened 2019-12-21
Remove timestamps in generated files.

samoht:
Also re-use the same code for generated dune files.

See #1030


functoria/lib/functoria_misc.ml
```diff
@@ -72,10 +72,9 @@ module Codegen = struct
 
   let main_ml = ref None
 
-  let generated_header ?(argv=Sys.argv) ?(time=Ptime_clock.now ()) () =
-    Format.asprintf "Generated by %s (%a)."
+  let generated_header ?(argv=Sys.argv) () =
+    Format.asprintf "Generated by '%s'."
```
hannesm: to avoid potential issues, why not use `%S` here, which does a String.escape and surrounds it with `"`?

hannesm:
looks good to me, and it removes code -- thus I'm in favour! :)

PR #1039, opened 2020-01-06
functoria: move packages functions into a Functoria_package module.

samoht:
This also abstracts the Functoria.package type, and rename the constructor
and accessors to use more sensible names.

We also vendor the function to compare opam version to simplify the code.


functoria/lib/functoria_package_version.ml
```diff
@@ -0,0 +1,168 @@
+(******************************************************************************)
+(*  This file is part of the Dose library http://www.irill.org/software/dose  *)
+(*                                                                            *)
+(*  Copyright (C) 2011 Ralf Treinen <ralf.treinen@pps.jussieu.fr>             *)
+(*                                                                            *)
+(*  All rights reserved. This file is distributed under the terms of the      *)
+(*  GNU Lesser General Public License version 2.1, with the special           *)
+(*  exception on linking described in the file LICENSE.                       *)
```
hannesm: I'm not sure whether we want to vendor LGPL code in mirage..

hannesm: and with recent functoria changes, we shouldn't need to compare version numbers... https://github.com/mirage/functoria/pull/183

samoht: hum ok I'll revert this change. I was not totally sure about the current code, but let's keep it as is :-)

hannesm: sorry for failing to document the current code. the main idea is to track the versions as strings around, and emit whatever constraints are there to the opam file. then opam will figure out whether there's a solution with all the constraints (and we don't need to handle version comparison).

samoht: makes sense, I've reverted my change and updated the tests to `Package.merge` to show what happens after a merge for min/max versions.

Drup:
Looks good!

PR #1042, opened 2020-01-07
Move basic implementations to Functoria (from Functoria_app)

samoht:
Also add (basic) tests for these.


test/functoria/key_gen.ml.expected
```diff
@@ -0,0 +1,5 @@
+(* Generated by "./test.exe". *)
+
+
+let runtime_keys = List.combine [] []
```
samoht: Not sure why this is empty...

test/functoria/test_impl.ml
```diff
@@ -0,0 +1,22 @@
+(* yiikes *)
+let () =
+  Functoria_misc.Codegen.set_main_ml "main.ml"
+
+let test_keys () =
+  let dev = Functoria.(keys sys_argv) in
+  let info =
+    Functoria.Info.create
+      ~packages:[]
+      ~context:Functoria_key.empty_context
```
Drup: My memory is a bit fuzzy, but I don't think that should be empty to test what you want to test.

test/functoria/key_gen.ml.expected
samoht: Actually it's empty because the default implementation that I am using do not define any keys :p

functoria/lib/functoria.ml
```diff
@@ -250,3 +247,161 @@ module type KEY =
...
+          Bos.Cmd.(v "opam" % "list" % "--installed" % "-s"
+                   % "--color=never" % "--depopts" % "--resolve" % pkgs_str
+                   % "--columns" % "name,version")
+        in
+        (Bos.OS.Cmd.run_out cmd |> Bos.OS.Cmd.out_lines) >>= fun (deps, _) ->
+        let deps =
+          List.fold_left (fun acc s ->
+              match String.cuts ~empty:false ~sep:" " s with
+              | [n; v] -> (n, v) :: acc
+              | _ -> acc
```
hannesm: this looks a bit brittle, to just drop anything without a whitespace (or with multiple ones). if it can't happen, I'd be fine with an `assert false` here, if it can happen, maybe a warning message should be emitted?

functoria/lib/functoria.ml
```diff
@@ -250,3 +247,161 @@ module type KEY =
...
+
+    method !configure _i = Ok ()
+
+    method !build i =
+      Log.info (fun m -> m "Generating: %a" Fpath.pp file);
+      (* this used to call 'opam list --rec ..', but that leads to
+         non-reproducibility, since this uses the opam CUDF solver which
+         drops some packages (which are in the repositories configured for the
+         switch), see https://github.com/mirage/functoria/pull/189 for further
+         discussion on this before changing the code below.  *)
```
hannesm: you may want to update the comment why `--resolve` is used now -- I tested the `--resolve` command manually below, and it seems to work fine for me

hannesm:
the `info_gen` modifications look good to me

samoht:
All green! Merging this to unlock the rest of the patches....

PR #1047, opened 2020-01-09
Functoria: rename Functoria.configurable into Functoria.Device

samoht:
Remove objects and classes and clean-up few related things.

This is not ready to be merged yet but `functoria` and `functoria-runtime` seems to compile and their tests pass.

### DIFF
--- a/functoria/app/functoria_app.ml
+++ b/functoria/app/functoria_app.ml
@@ -522,7 +522,8 @@ module Make (P: S) = struct
     Codegen.append_main "let _ = Printexc.record_backtrace true";
     Codegen.newline_main ();
     Cache.save ~argv (Info.build_dir i) >>= fun () ->
-    Engine.configure_and_connect ~init i jobs >>| fun () ->
+    Engine.configure i jobs >>| fun () ->
+    Engine.connect i ~init jobs;
     Codegen.newline_main ()
 
   let clean_main i jobs =
--- a/functoria/lib/functoria.mli
+++ b/functoria/lib/functoria.mli
@@ -34,6 +34,8 @@
 
 *)
 
+open Rresult
+
 (** {1:combinators Combinators} *)
 
 (** The type for values representing module types. *)
@@ -70,7 +72,7 @@ val abstract: _ impl -> abstract_impl
 
 (** {1:keys Keys} *)
 
-type key = Functoria_key.t
+type abstract_key = Functoria_key.t
 (** The type for command-line keys. See {!Functoria_key.t}. *)
 
 type context = Functoria_key.context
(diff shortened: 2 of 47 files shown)


test/functoria/gen-1/main.ml.expected
```diff
@@ -2,31 +2,31 @@ let (>>=) x f = f x
...
   return Info_gen.info
   )
 
-let f11 = lazy (
-  let __key1 = Lazy.force key1 in
-  let __info1 = Lazy.force info1 in
-  __key1 >>= fun _key1 ->
-  __info1 >>= fun _info1 ->
-  App1.start _key1 _info1
+let app_make = lazy (
```
samoht: might be better to match `impl_name` instead of `module_name` here.

Drup:
Hmm, have you tried to adapt mirage yet ? It uses inheritance in quite a few places.

samoht:
And now mirage compiles.

A good side-effect of that refactoring is that it should be easier to add unit tests ; I'll see what breaks in `mirage-skeleton` and I'll add the corresponding tests in this repository.

functoria/lib/functoria.mli
```diff
@@ -38,8 +38,8 @@
 
 (** The type for values representing module types. *)
 type _ typ =
-  | Type    : 'a -> 'a typ
-  | Function: 'b typ * 'c typ -> ('b -> 'c) typ
+  | Type   : 'a -> 'a typ
+  | Functor: 'b typ * 'c typ -> ('b -> 'c) typ
```
samoht: this should probably be done in another PR to avoid adding noise.

functoria/lib/functoria.mli
```diff
@@ -129,138 +129,139 @@ val package :
 (** {1:app Application Builder}
 
     Values of type {!impl} are tied to concrete module implementation
-    with the {!foreign} construct. Module implementations of type
+    with the {!device} construct. Module implementations of type
```
samoht: noise

functoria/app/functoria_app.ml
```diff
@@ -519,15 +519,16 @@ module Make (P: S) = struct
     Codegen.newline_main ();
     Codegen.append_main "%a" Fmt.text  P.prelude;
     Codegen.newline_main ();
-    Codegen.append_main "let _ = Printexc.record_backtrace true";
+    Codegen.append_main "let () = Printexc.record_backtrace true";
```
samoht: this should go in another PR which cleanup the generated code

functoria/lib/functoria_engine.ml
```diff
@@ -101,57 +84,49 @@ let find_device info g i =
...
       end
   in
-  let f v res = res >>= fun () -> f v in
-  Graph.fold f t @@ R.ok () >>| fun () ->
-  tbl
+  Graph.iter f t
 
 let meta_init fmt (connect_name, result_name) =
-  Fmt.pf fmt "let _%s =@[@ Lazy.force %s @]in@ " result_name connect_name
+  Fmt.pf fmt "let %s =@[@ Lazy.force %s @]in@ " result_name connect_name
```
samoht: this should go in another PR which cleanup the generated code

functoria/lib/functoria_engine.ml
```diff
@@ -101,57 +84,49 @@ let find_device info g i =
...
 
 let emit_connect fmt (iname, names, connect_string) =
-  (* We avoid potential collision between double application
-     by prefixing with "_". This also avoid warnings. *)
-  let rnames = List.map (fun x -> "_"^x) names in
   let bind ppf name =
-    Fmt.pf ppf "_%s >>= fun %s ->@ " name name
+    Fmt.pf ppf "%s >>= fun %s ->@ " name name
```
samoht: this should go in another PR which cleanup the generated code

functoria/lib/functoria_engine.ml
```diff
@@ -101,57 +84,49 @@ let find_device info g i =
...
-    (connect_string rnames)
+    Fmt.(list ~sep:nop meta_init) (List.combine names names)
+    Fmt.(list ~sep:nop bind) names
+    (connect_string names)
 
 let emit_run init main =
   (* "exit 1" is ok in this code, since cmdliner will print help. *)
   let force ppf name =
-    Fmt.pf ppf "Lazy.force %s >>= fun _ ->@ " name
+    Fmt.pf ppf "Lazy.force %s >>= fun () ->@ " name
```
samoht: this should go in another PR which cleanup the generated code

samoht:
Ok I think this is ready for a review, if anyone is interested and/or have an opinion :-)

Next, I'll make a PR on mirage-skeleton to use the new API.

samoht:
All green!

dinosaure:
So, I come after the battle, but this PR introduces something strange for, it deletes the `name` value of the device object which is used as a unique identifier for a "device". I just would like to know what is the semantic now. Before this PR, the name is used to represent the value given to the `start` function.  I suspect that we use now the name of the module but I need a confirmation about that.

samoht:
Devices are now uniquely identified using a unique [id](https://github.com/mirage/mirage/blob/master/lib/functoria/device.ml#L61). 

That `id` is generated when you use `Device.v` so that means two instantiations of a device with the same parameter will create 2 different devices. If you want to avoid this, you need to make sure you call `Device.v` once.

Do you have some code where this is a problem?

dinosaure:
It's perfect, thanks.

PR #1050, opened 2020-01-17
Use ocamlformat 0.12

samoht:


samoht:
I don't undertand why the CI is failing with:

```
#=== ERROR while compiling conf-perl.1 ========================================#
"perl": command not found.
```

What MirageOS package is using perl?

samoht:
So actually the error seems to be transient:

```
(4/7) Installing perl (5.28.2-r1)
ERROR: Failed to create usr/share/perl5/core_perl/Unicode/Collate/CJK/Zhuyin.pm: I/O error
ERROR: perl-5.28.2-r1: IO ERROR
```

I'm restarting the failed container.

PR #1053, opened 2020-02-05
Make Functoria.typ abstract

samoht:


Drup:
I agree it's nice to isolate the main type in its own module, but I'm not so convinced by making it abstract. What's your rational for that ?

samoht:
It's easier to change later if it's abstract. I agree that in that case it's not very important, so I'm fine to revert that change if you really want to keep it concrete :-)

samoht:
Updated -- I've removed the abstract type.

PR #1060, opened 2020-02-17
Add a `query opam` sub-command

samoht:
It's probably best to generate an `.install` file instead of having `cp` commands, but I've just copied what is there already. Improvements can be done in a subsequent PR.

Note also that this PR adds a new `install` property to devices: this allows individual devices to declare which `bin` or `etc` artefacts they want to install. Not sure if that's very useful to expose this, but here it it is anyway.


lib/mirage_configure.ml
```diff
@@ -29,97 +30,6 @@ let clean_myocamlbuild () =
...
-      append fmt "# %s" (generated_header ());
-      Info.pp_opam ~name fmt info;
-      append fmt {|maintainer: "dummy"|};
-      append fmt {|authors: "dummy"|};
-      append fmt {|homepage: "dummy"|};
-      append fmt {|bug-reports: "dummy"|};
-      Format.fprintf fmt {|build: [ "sh" "-exc" "|};
-      ( match subdir with
-      | None -> ()
-      | Some p -> Format.fprintf fmt "cd %a && " Fpath.pp p );
```
samoht: This is very ugly and fragile, so I haven't copied that in my PR.  Instead I'm using the `--config-file` argument here. Not sure if that's make a difference, I'll try to test this (not sure how yet...)

samoht:
The tests are failing with:
```
19 8.917 # -url { src: "<URL>#<BRANCH>" }
#19 8.917 # +url { src: "file:///src#master" }
#19 8.917 #         test alias test/functoria/runtest
```
I don't understand the auto-detection logics, which was added in #1022. @hannesm all the auto-detection logics sounds very fragile to me (and very hard to test!)-- do you mind explaining the bits that you absolutely need vs. the rest? e.g. do you need branch detection? What happens if that branch is moving?

lib/mirage_configure.ml
samoht: Hum actually I think I understand what you tried to do here -- I'll try to replicate and somehow test that behaviour automatically.

hannesm:
I've no clue what you're trying to fix/change here, but here is what I attempted to do with #1022 (which, as you describe, is not working in all cases -- but a best-effort solution that works for surprisingly many unikernels):
the mirage-generated opam file contains additional information, so it is fully usable as an opam package. this entails:
- (a) installation instructions (the resulting unikernel binary image is copied into `%{bin}%`), now `opam install <mirage-generated-opam-file>` can be used to preserve the binary image (additional artifacts, such as libvirt.xml etc. are as well installed).
- (b) embed a source url in the opam file - conventionally (see e.g. mirage-skeleton) unikernels are developed in subdirectories of git repositories -- mirage configure now attempts to find a `.git` repository, and if found executes `git rev-parse --abbrev-ref HEAD`to figure out the branch name, and `git remote get-url origin` to figure out the remote (plus some code that rewrites git+ssh of github.com to https). the subdirectory is as well tracked, and involved in the build and install instructions.

with these two changes, I can take such an mirage-generated opam-file (target-specific / configured) as the only metadata (not entirely true, there's as well the need for opam-repository and other potential pinned packages) to get the unikernel compiling using opam. I use this feature as a basis for reproducible builds (see https://github.com/roburio/reproducible-unikernel-repo/blob/master/packages/unikernel-hello-hvt.0.0.0/opam). the required metadata from opam-repository (and other pins), as well as replacing the branch with a specific unique commit id is actually handled by orb / https://github.com/ocaml/opam/pull/4055. You may want to look into https://hannes.nqsb.io/Posts/ReproducibleOPAM for further information. Having said that, I'm open for more mature solutions (my main motivation was that I'm lazy and didn't want to edit these mirage-generated opam files by hand).

>  Instead I'm using the `--config-file` argument here.

I did not because I still lack the understanding what it does (and tests with this argument only confused me furher) -- is it supposed to work in both configure and build? Where do the build results end up (in the directory where the config file resides or `cwd`)?

As said, I hope I was able to shed some light into what it is used for. And I'm sure there could be more mature solutions.

PR #1066, opened 2020-02-23
Use `version` instead of hard-coding a non-existing upper-bound.

samoht:
This makes pinning to a `dev` version possible.

/cc @hannesm 

It's a mix between what we had before and your [fix](https://github.com/mirage/mirage/pull/1019): we have some slack to release mirage and mirage-runtime separately while keeping the upper-bound to a known existing version (the current one!)

### DIFF
--- a/mirage.opam
+++ b/mirage.opam
@@ -25,7 +25,7 @@ depends: [
   "astring"
   "logs"
   "stdlib-shims"
-  "mirage-runtime"     {>= "3.7.0" & < "3.8.0"}
+  "mirage-runtime"     {>= "3.7.0" & <= version}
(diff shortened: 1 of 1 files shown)


samoht:
(this is needed to fix ocaml-ci ... we tried to not use a `dev` version for pinning and use the max existing version instead, but that breaks other projects)

hannesm:
hmm, I fail to see the big picture about which version of what should be compatible with what else. the generation of upper bounds were introduced in mirage 3.3.0 to allow further development of APIs (keeping in mind that old configured unikernels are out there which we don't want to break).

PR #1068, opened 2020-02-24
Revise random

hannesm:
please see discussion at https://github.com/mirage/mirage-crypto/issues/6

this is a PR against the 3.7 branch. this PR is backwards compatible, but the choice of `--prng` (=nocrypto/fortuna/stdlib) does not have any effect. instead, Fortuna will always be used.

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -1,5 +1,5 @@
 eval `opam config env`
 opam depext -uiy mirage
 cd ~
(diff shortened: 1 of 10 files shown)


hannesm:
now that conduit is released to opam-repository, the travis job should succeed (and if it does, I'll merge + release).

hannesm:
done in https://github.com/ocaml/opam-repository/pull/16021

PR #1077, opened 2020-03-09
Define a new Functoria.Action module and use it everywhere

samoht:
Follow up from https://github.com/mirage/mirage/pull/930

### DIFF
--- a/lib/functoria/DSL.mli
+++ b/lib/functoria/DSL.mli
@@ -33,8 +33,6 @@
 
 (** {1:combinators Combinators} *)
 
-open Rresult
-
 type 'a typ = 'a Type.t
 (** The type for values representing module types. *)
 
@@ -152,9 +150,9 @@ val impl :
   ?keys:Key.t list ->
   ?extra_deps:Impl.abstract list ->
   ?connect:(Info.t -> string -> string list -> string) ->
-  ?configure:(Info.t -> (unit, R.msg) result) ->
-  ?build:(Info.t -> (unit, R.msg) result) ->
-  ?clean:(Info.t -> (unit, R.msg) result) ->
+  ?configure:(Info.t -> unit Action.t) ->
+  ?build:(Info.t -> unit Action.t) ->
+  ?clean:(Info.t -> unit Action.t) ->
   string ->
   'a typ ->
   'a impl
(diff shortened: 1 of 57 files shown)


samoht:
What needs to be done:
- port the tests from #930 (and maybe add more?)
- update mirage-skeleton

samoht:
Updated, now the only module using `Bos` is `Functoria.Action`. There are too much `Action.run` in the current code, but they are a bit hard to remove (as doing will propagate `'a Action.t` a bit everywhere, especially inside `'a Cmdliner.Term.t` where they become a pain to manipulate).

Current use of `Action.run`:
```
lib/functoria/app.ml:    build_and_execute ~state ?help_ppf ?err_ppf argv |> Action.run |> function
lib/functoria/app.ml:      run_configure_with_argv ~state Action.run Sys.argv c
lib/functoria/app.ml:    run () |> Action.run |> exit_err
lib/functoria/opam.ml:    match Action.run @@ find_git () with
lib/mirage/mirage_key.ml:    ( match Action.run @@ Action.run_cmd_out Bos.Cmd.(v "uname" % "-s") with
test/f0/f0.ml:    match Action.run @@ split_root () with
test/functoria/gen-1/test.ml:  match Action.run (test ()) with Ok () -> () | Error (`Msg e) -> failwith e
test/functoria/gen-2/test.ml:  match Action.run (test ()) with Ok () -> () | Error (`Msg e) -> failwith e
test/mirage/gen-1/test.ml:  match Functoria.Action.run (test ()) with
```
The one in `functoria/opam.ml` are filesystem access to guess the repository root ; and the one in `mirage/mirage_key.ml` is to find the kind of OS we are running on. I think that's ok to keep them like this for now on. The important ones are the one in `functoria/app.ml` where they are used in `register` and `run_with_argv`. We could expose similar functions which are interpreted instead to ease testing, I'll experiment with this next.

samoht:
Tests have been added, but it seems that something is a bit off, as `test/mirage/action/test.expected` is truncated vs. https://github.com/mirage/mirage/pull/930/files#diff-7661350cdd5ae1288a23149c76a53107

samoht:
I've just pushed a few tests for `Functoria.Action`. It should now have a good test coverage.

Still need to inspect the mirage test results before merging this, as they look a bit empty to me.

dinosaure:
The goal of this PR is to track any operations on the system (like `copy`, `open_file`, etc.) ?

samoht:
Yes the goal is to never use `Bos` directly. `'a Action.t` actions can be introspected, so we can compute the set of generated files automatically. It's also makes the whole thing much easier to test (just print the trace instead of calling Bos) and it also allow to add a `--dry-run` command to list the commands before running them.

samoht:
Pushed a new version, with more tests and a small fix on mirage-skeleton which should make Travis happt.

samoht:
All green! I'll squash and merge here and in mirage-skeleton

PR #1082, opened 2020-03-16
functoria: add a way to check link-time conflicts

samoht:


samoht:
I'm actually not very convinced this is a very good idea (as we don't check for conflicts recursively).

PR #1085, opened 2020-03-18
Add a `mirage query dune` and `mirage query dune.inc` subcommands

samoht:
This will allow the generation of dune files to use dune instead of ocamlbuild.

What is missing for having a full dune workflow:
- [ ] the mirage device needs to be modified to register dune stanzas (#1024 kind of already does this but directly writing in the dune file)


test/functoria/query/install.expected
```diff
@@ -0,0 +1,7 @@
+bin: [
+  "test/functoria/query/main.exe" {"noop"}
```
samoht: `test/functoria/query/` should not be in the path

test/functoria/query/dune.inc.expected
```diff
@@ -0,0 +1,11 @@
+;; Generated by "test query dune.inc".
+
+(rule
+  (targets .mirage.config dune.build key_gen.ml)
```
samoht: dune.build should not be there

test/functoria/query/dune.base.expected
```diff
@@ -0,0 +1,12 @@
+;; Generated by "test query dune.base".
```
samoht: It would be nice to embed the version of the tool too

test/functoria/query/dune.full.expected
```diff
@@ -0,0 +1,23 @@
...
+  (libraries f0 functoria))
+
+(rule
+  (alias config)
+  (targets dune)
+  (mode promote-until-clean)
+  (action (with-stdout-to dune (run ./config.exe query dune))))
+
+(rule
+  (targets .mirage.config key_gen.ml)
```
samoht: .mirage.config is generate by `mirage configure`, not `./config.exe config`

test/functoria/query/install.expected
```diff
@@ -0,0 +1,7 @@
+bin: [
+  "./main.exe" {"noop"}
+]
+etc: [
+  "./vote" {"vote"}
+  "./warn_error" {"warn_error"}
+]
```
rgrinberg: Where do these come from? I don't see any install stanzas generated

samoht: the functoria tool is generating these (could be folded in dune-project later on I guess)

samoht:
I've added a rough [design document](https://github.com/samoht/mirage/blob/query-dune/DESIGN.md) to explain the danse of the various sub-commands...

lib/functoria/action.ml
```diff
@@ -39,13 +39,12 @@ type _ command =
...
   | Get_var : string -> string option command
   | Set_var : string * string option -> unit command
-  | With_dir : Fpath.t * (unit -> 'a t) -> 'a command
   | Pwd : Fpath.t command
   | Tmp_file : int option * tmp_name_pat -> Fpath.t command
   | Write_file : Fpath.t * string -> unit command
   | Read_file : Fpath.t -> string command
   | With_output :
-      int option * Fpath.t * string * (Format.formatter -> 'a)
+      int option * bool * Fpath.t * string * (Format.formatter -> 'a)
```
rgrinberg: Inline record perhaps? The `bool` param is a bit hard to understand.

test/functoria/query/dune.full.expected
```diff
@@ -0,0 +1,33 @@
...
+  (action (run ./config.exe build)))
+
+(rule
+  (alias clean)
+  (action (run ./config.exe clean)))
+
+(executable
+   (name      main)
+   (modules   (:standard \ config))
+ ; (promote   (until-clean))
```
rgrinberg: Why do we need a commented line?

samoht: If I uncomment this line, the rule is evaluated unconditionally for some reason. Will try to make a smaller repro when everything works.

samoht:
Now all the unit tests are passing, and the e2e tests are working on my machine (unfortunately, they can't run in the CI yet). 🎉 

Next steps:
- port mirage runes for dune in the new system (should be easy, the hard work has been done in #1024);
- add the missing features (dune-worskpace generation, opam/.install file generation at the root of the repo) if they are needed;
- think about a merging strategy -- there are a few piece which can be cut out and upstreamed separately but most of the changes are quite pervasive so I'm not sure it's really worth the effort; I'll do this anyway for small changes (like the removal of `Arg.required` that nobody seems to be using).
- test, test, test, test

samoht:
Rebased -- there is still a bit much noise (I'll make a few more cleanup PRs tomoorw) but any feedback/review is welcome.

The build system of mirage is still using ocamlbuild, this (oddely) seems to continue to work as I've managed to compile mirage-skeleton. But the paths of the binaries is a bit obscure now. Will add dune+mirage support tomorrow.

test/functoria/query/install.expected
```diff
@@ -1,7 +1,7 @@
 bin: [
-  "test/functoria/query/main.exe" {"noop"}
+  "main.exe" {"noop"}
```
samoht: this is not great, but should be fixed once everything use dune and we don't have to worry with .install files anymore.

samoht:
Closing in favour of #1153

PR #1089, opened 2020-03-22
xen backend now uses pvh

hannesm:


hannesm:
superseeded by https://github.com/mirage/mirage/issues/1159

PR #1090, opened 2020-03-25
Remove Functoria.Key.Arg.required

samoht:
This was not used anywhere and was adding uneeded complexity.


hannesm:
hold on for a bit, I use this [here](https://github.com/roburio/unikernels/blob/master/certificate/config.ml), and think it is a good idea to keep it.

samoht:
Good to know that it is used :p I'm closing this then and I'll add a few tests.

hannesm:
thanks @samoht

PR #1094, opened 2020-03-25
Add Functoria.Key.find

samoht:
This allow to test if a key is present on the command-line (default values
are not added to the context anymore). Extracted from #1085 

/cc @Drup -- I don't think I have broken anything with this, but I'm not 100% sure :-)

### DIFF
--- a/lib/functoria/key.ml
+++ b/lib/functoria/key.ml
@@ -136,6 +136,19 @@ module Arg = struct
   let required ?(stage = `Both) conv info =
     { stage; info; kind = Required conv }
 
+  let default (type a) (t : a t) =
+    match t.kind with Opt (d, _) -> d | Flag -> false | Required _ -> None
+
+  let eq conv x y =
+    let str = Fmt.to_to_string conv.serialize in
+    str x = str y
+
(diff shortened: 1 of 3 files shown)


samoht:
I'm merging this a bit quickly to cleanup the review for #1085, but feel free to review/comment anyway whenever you have time.

Drup:
I'm pretty sure you haven't broken anything (and there are lot's of tests now! :D) but the use of `is_default` bothers me quite a bit, because it means if you pass `--foo false` (or `--bar default_value`, it's considered the same as if you hadn't given an argument at all.

I suggest we change the transformation into cmdliner term to always return options (just like we do in the `Opt` case, and we shave off the option in the `context` function. Does that make sense ?

samoht:
> I suggest we change the transformation into cmdliner term to always return options (just like we do in the Opt case, and we shave off the option in the context function. Does that make sense ?

That sounds good too -- I am indeed not totally happy with collapsing the default and "no argument" case.

samoht:
I've implemented the optional default keys in https://github.com/mirage/mirage/pull/1145

PR #1103, opened 2020-03-26
functoria: do not treat main.ml as a special file anymore.

samoht:
This allows main.ml to appear when doing `<cmd> query files-configure`

This is extracted from #1085


lib/functoria/app.ml
```diff
@@ -415,20 +415,16 @@ module Make (P : S) = struct
...
-    Codegen.set_main_ml file;
-    Codegen.append_main "(* %s *)" (Codegen.generated_header ());
-    Codegen.newline_main ();
-    Codegen.append_main "%a" Fmt.text P.prelude;
-    Codegen.newline_main ();
-    Codegen.append_main "let _ = Printexc.record_backtrace true";
-    Codegen.newline_main ();
+    let main = Info.main i in
+    Log.info (fun m -> m "Generating: %a" Fpath.pp main);
+    Action.with_output ~path:main ~append:false ~purpose:"Create main.ml"
```
hannesm: shouldn't `purpose` use `main` then as well, instead of refering to `main.ml`?

samoht: Thanks, fixed

PR #1108, opened 2020-03-26
functoria: remove the --build-dir command-line option

samoht:
This is not used anywhere and it makes it very hard to change the build logic.

This is extracted from #1085 

/cc @hannesm to double-check that this option is indeed not used/useful.

### DIFF
--- a/lib/functoria/app.ml
+++ b/lib/functoria/app.ml
@@ -29,7 +29,6 @@ module Config = struct
   type t = {
     name : string;
     build_cmd : string list;
-    build_dir : Fpath.t;
     packages : package list Key.value;
     keys : Key.Set.t;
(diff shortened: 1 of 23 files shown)


hannesm:
I have not used the `--build-dir` option at all (but thought you intended to use it for the `build` instructions in https://github.com/mirage/mirage/pull/1060 ?)

samoht:
The build-dir is now inferred from the path of config.ml: https://github.com/mirage/mirage/pull/1108/files#diff-67172689a45280c3e8a837bd93f1f991R108 and https://github.com/mirage/mirage/pull/1108/files#diff-67172689a45280c3e8a837bd93f1f991R105

PR #1110, opened 2020-03-26
functoria: split App into the Tool and Lib modules

samoht:
Extracted from #1085 and on top of #1108 

The e2e tests are failing, need to investigate tomorrow.


samoht:
The e2e tests are now passing. 

The split introduces a bit of code duplication (and the logic is still a bit more convoluted) so I'll push a few more commits to simplify this before merging (for instance Tool doesn't use `Cli.peek` yet).

PR #1118, opened 2020-03-27
WIP: move opam file generation from ./config.exe to the tool

samoht:
Extracted from #1085 

We want to extract opam and .install file generation from `./config.exe` as the goal is that command to always run in the build dir (and hence it doesn't know the dune workspace root). We want the `mirage` tool do generate all the `dune-projet`, `dune-workspace`, `opam, `Makefile`, etc which are at the root of the project as well as context caches (which should be set only by configure or by the user).

To do this `mirage` will query `./config.exe query` to get useful information but for most of these files this is not needed.

We also need to continue to clean-up the `Tool` API as currently it doesn't parse the sub-command properly and always re-exec (we need a special case for clean, but can as well import the full logic from #1085 which is much simpler)


test/functoria/tool/simple-help.err.expected
```diff
@@ -10,4 +10,9 @@
                Write to dune.build (28 bytes)
                Is_file? dune -> true
                Read dune (221 bytes)]
+* Run_cmd 'dune exec --root . -- ./config.exe query "name" --dry-run' (ok: out="[...]" err="")
+* Run_cmd 'dune exec --root . -- ./config.exe query "opam" --dry-run' (ok: out="[...]" err="")
+* Is_file? [...].opam -> false
```
samoht: `mirage help` should not call `./config.exe`

test/functoria/tool/version.err.expected
```diff
@@ -10,5 +10,12 @@
                Write to dune.build (28 bytes)
                Is_file? dune -> true
                Read dune (221 bytes)]
+* Run_cmd 'dune exec --root . -- ./config.exe query "name" --version a b c
```
samoht: `mirage --version` should not call  `./config.exe`

samoht:
The different parts of this PR have been merged individually, so closing.

PR #1131, opened 2020-04-01
functoria: use --depext by default (as the mirage tool did)

samoht:
This should fix mirage-skeleton


samoht:
There is still an issue with mirage-skeleton (due to `mirage query` not reading the cached context), fixing this next.

PR #1132, opened 2020-04-01
functoria: `<tool> query' should read the cached context

samoht:
Should fix the build of mirage-skeleton

PR #1133, opened 2020-04-03
functoria: fix reading of cached arguments

samoht:
This commit ensure that all commands are properly reading the cache
and merging it with the current command-line arguments.

From #1132


samoht:
mirage-skeleton is still broken for an unknown reason, I'll continue to investigate...

PR #1148, opened 2020-04-15
Upgrade to OCamlformat v0.14.1

CraigFe:
- reformats the code to be compliant with OCamlformat v0.14.1
- updates the `.ocamlformat` file accordingly


samoht:
I like that diff :-) Thanks!

PR #1154, opened 2020-05-12
Next 3.7

hannesm:
while I appreciate work on MirageOS 4, I don't want to have it blocking releases and uses of other libraries, in particular:
- TLS 1.3 support in tls 0.12.0 (https://github.com/ocaml/opam-repository/pull/16423)
- conduit with a dns-client that times out requests ~~(PR'ed https://github.com/mirage/ocaml-conduit/pull/314)~~ (https://github.com/ocaml/opam-repository/pull/16425)
- general observation that our `Bos.OS.Cmd.run_out` were too sloppy, leading to errors (exit <> 0) translated to success
- the ability to run mirage (link) without an `opam` (used to figure out the prefix), if `PREFIX` is defined. this goes well with how `PKG_CONFIG_PATH` is handled

if nobody objects, I'd test more thoroughly and release a 3.7.7. Thanks for reading.

### DIFF
--- a/lib/mirage_impl_random.ml
+++ b/lib/mirage_impl_random.ml
@@ -1,28 +1,25 @@
 open Functoria
-open Mirage_impl_misc
+open Mirage_impl_mclock
+open Mirage_impl_time
 
 type random = RANDOM
 let random = Type RANDOM
 
 let random_conf = object
   inherit base_configurable
-  method ty = random
+  method ty = time @-> mclock @-> random
   method name = "random"
-  method module_name = "Mirage_crypto_rng"
+  method module_name = "Mirage_crypto_rng_mirage.Make"
   method! keys = [ Mirage_key.(abstract prng) ]
   method! packages =
-    Mirage_key.(if_ is_unix)
-      [ package ~sublibs:["unix"] "mirage-crypto-rng" ]
-      [ package "mirage-crypto-entropy" ]
-  method! connect i _ _ =
-    match get_target i with
-    | #Mirage_key.mode_unix ->
-      "Lwt.return (Mirage_crypto_rng_unix.initialize ())"
-    | _ ->
-      (* here we could use the boot argument (--prng) to select the RNG! *)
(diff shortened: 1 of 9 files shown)


dinosaure:
CI fails due to `conduit`, otherwise, I'm ok with this release.

samoht:
LGTM

And yes we should not block 3.7.* releases. I'll backport your changes in my tree.

hannesm:
cool, as mentioned on wednesday, I'll merge this after mirage-crypto 0.7.0 is out (which involves some rng changes, sorry for the hazzle)

lib/mirage_impl_random.ml
```diff
@@ -1,28 +1,25 @@
...
+    Mirage_key.pure [
+      package ~sublibs:["mirage"] ~min:"0.7.0" "mirage-crypto-rng"
+    ]
+  method! connect _i modname _ =
+    (* here we could use the boot argument (--prng) to select the RNG! *)
+    Fmt.strf "%s.initialize (module Mirage_crypto_rng.Fortuna)" modname
 end
 
-let default_random = impl random_conf
+let default_random = impl random_conf $ default_time $ default_monotonic_clock
```
hannesm, line 23: for those more into the mirage code base: I attempted
```OCaml
let default_random ?(time = default_time) ?(clock = default_monotonic_clock) () = impl random_conf $ time $ clock
```

which is what is done elsewhere as well, but there are some questions:
- who uses non-default labeled arguments? (if there isn't anyone, why bother?)
- on the use-site in other devices, we can no longer `?(random = default_random)`, but now require `time` and `mclock` explicitly and then `let random = match random with Some r -> r | None -> default_random ~time ~clock () in ...`
--> does, if there are now multiple calls to `default_random`, they all end up with the same device?
- since this is a minor release, I'd like to retain API compatibility, what is the supposed way to do so? expose a `Mirage.default_random_expert` with the new signature?

PR #1163, opened 2020-06-09
Forward-port changes for the new TLS stack

samoht:
See https://github.com/mirage/mirage/pull/1154 for the PR against the 3.7 branch.

### DIFF
--- a/lib/mirage/mirage.ml
+++ b/lib/mirage/mirage.ml
@@ -60,6 +60,8 @@ let stdlib_random = Mirage_impl_random.default_random
 
 let default_random = Mirage_impl_random.default_random
 
+let rng = Mirage_impl_random.rng
+
 let nocrypto = Mirage_impl_random.nocrypto
 
 let nocrypto_random = Mirage_impl_random.default_random
--- a/lib/mirage/mirage.mli
+++ b/lib/mirage/mirage.mli
@@ -125,6 +125,9 @@ val default_random : random impl
 (** Default PRNG device to be used in unikernels. It uses getrandom/getentropy
     on Unix, and a Fortuna PRNG on other targets. *)
 
+val rng : ?time:time impl -> ?mclock:mclock impl -> unit -> random impl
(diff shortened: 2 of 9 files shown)


hannesm:
> I've also added an `rng` device

that is nice, though I'm still a bit confused how to use it. From my point of view, a MirageOS unikernel should
- use a "default" time device
- use a "default" clock device
- use a "default" random device

Now, if the user select some special from above, all reverse-dependent devices should use that -- i.e. the ARP layer, the IP layer, the TCP layer should use the specially selected random device. Or, if a specific clock device is used (e.g. where time passes faster or slower), the default random device (used by all devices/layers above) should use that. And I'm not sure whether (how?) this is possible without writing lots of manual `~clock:my_favourite_clock` when instantiating the stack. It may as well be that feature in my head is ill-advised (and/or not in scope for MirageOS4).

dinosaure:
> Now, if the user select some special from above, all reverse-dependent devices should use that -- i.e. the ARP layer, the IP layer, the TCP layer should use the specially selected random device. Or, if a specific clock device is used (e.g. where time passes faster or slower), the default random device (used by all devices/layers above) should use that. And I'm not sure whether (how?) this is possible without writing lots of manual ~clock:my_favourite_clock when instantiating the stack. It may as well be that feature in my head is ill-advised (and/or not in scope for MirageOS4).

An other possibility is to enforce some implementations to be unique into the `functoria` graph. BY this way, we ensure that any `Mclock` refer to only one and unique implementation - note that with this view, it still is possible to use another `Mclock` but we enforce the fact that 2 implementations of it is something wrong.

samoht:
> Now, if the user select some special from above, all reverse-dependent devices should use that -- i.e. the ARP layer, the IP layer, the TCP layer should use the specially selected random device. 

I've just pushed 4a301c7 which builds a custom random generator with the provided mclock and time devices (when that's possible). It is what you had in mind?

We could indeed probably extend the functoria graph with "unique" properties but that's a bit more work that we can indeed do after MirageOS 4.0.

PR #1166, opened 2020-06-13
xen: fix for xen 4.10+ (as seen in LTS Ubuntu)

djs55:
I notice that current mirage doesn't work on xen 4.10+ (as used in e.g. LTS Ubuntu) because they removed the `builder` argument from the `.xl` file and replaced it with `type`.

See https://wiki.xen.org/wiki/Type_Config_Option

Obviously we're also in the process of refactoring Xen support to use pvh as in #1089 

I don't know whether we should change it to `type = pv` in the meantime. I'm making this PR just in case -- I needed to make this change anyway for my local testing.


hannesm:
thanks, this is included in mirage 3.8.0 (this PR should still be merged on master so that it'll be part of the mirage 4 release as well).

hannesm:
thanks, included in #1185

PR #1174, opened 2020-06-17
Fix cmdliner term generation for keys with the same name but differen…

samoht:
…t defaults

Since https://github.com/mirage/mirage/pull/1162 the (CLI) name is not enough
to distinguish between keys. This means that we build a cmdliner term with
multiple keys having the same name. In that case cmdliner will explode at
runtime with something like:

    Fatal error: exception Invalid_argument("option name --interface defined twice
      (doc strings are 'The network interface listened by the unikernel.  ' and
      'The network interface listened by the unikernel.  ')")

This happens when we call `mirage --help`, `mirage describe`, `mirage clean`
with partial implementation choices (e.g. without a corresponding
`mirage configure`).

### DIFF
--- a/lib/functoria/key.ml
+++ b/lib/functoria/key.ml
@@ -257,6 +257,17 @@ and compare_setters : type a b. a setter list -> b setter list -> int =
   | Setter (x, _) :: tx, Setter (y, _) :: ty -> (
       match compare (Any x) (Any y) with 0 -> compare_setters tx ty | i -> i )
 
+(* Set of keys, without runtime name conflicts. This is useful to create a
+   valid cmdliner term. *)
+module Names = Stdlib.Set.Make (struct
+  type nonrec t = t
+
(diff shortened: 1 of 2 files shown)


Drup:
I'm still wondering if we should even accept keys with same names and different defaults. It seems to me we should just reject those (but with a proper error message)....

samoht:
So how would you handle  https://github.com/mirage/mirage/blob/master/lib/mirage/mirage_impl_network.ml#L36-L52 ?

Drup:
hugh, yes, alright.

Honestly, I'm not sure. This is the case where the non-applicativeness of the cmdliner term we are building comes back to bite us. It is not the first time. :)
Your proposed change is a good band-aid, but I agree with you that we should try to merge the keys, rather than arbitrary keep the last inserted.

samoht:
I'll merge the workaround and I've created an issue to remember that this could be improved.

PR #1180, opened 2020-07-03
add ocaml runtime arguments (OCAMLRUNPARAM)

hannesm:
the main motivation behind this PR is that I'd like to pass on the GC strategy (since 4.10 there's best-fit) without recompiling the unikernel (since the OCaml runtime allows this as a rutime configurable option).

Now, I was as well unsatisfied with the hardcoded emission of `let _ = Printexc.record_backtrace true` into every `main.ml` file (which can now be safely removed :).

This PR defines several boot parameters (also configure stage --> both), mostly equivalent of what the `OCAMLRUNPARAM` environment variable allows. The documentation has been inspired by gc.mli and http://caml.inria.fr/pub/docs/manual-ocaml/runtime.html#s%3Aocamlrun-options .

For how the code is generated: my goal is to avoid repeating to hardcode defaults from the OCaml runtime, thus the keys are of type `int option key` -- now there's some dirty `match` generated for them, I'd be fine if there's a better way to achieve this (I looked into the codebase, and only found things for `int key option`).

I'd be happy to get review + feedback on this PR (it targets the 3-series, but I'm fine to develop one for master as well, once accepted).

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -251,6 +251,73 @@ end
 
 include Functoria_app.Make (Project)
 
+let backtrace = impl @@ object
+    inherit base_configurable
+    method ty = job
+    method name = "ocaml_backtrace"
+    method module_name = "Printexc"
+    method! keys = [ Key.abstract Key.backtrace ]
+    method! connect _ _ _ =
+      Fmt.strf "Lwt.return (Printexc.record_backtrace %a)"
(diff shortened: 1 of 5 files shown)


dinosaure:
Just a quick look, and it seems fine. I will try to test it on some of my unikernels :+1:.

PR #1187, opened 2020-09-13
IPv6 stack support for the mirage utility

hannesm:
missing is a dual socket stack, requires https://github.com/mirage/mirage-tcpip/pull/433


hannesm:
on the application side, the changes for mirage-skeleton's network unikernel can be seen at https://github.com/hannesm/mirage-skeleton/commit/8616199da9fbeb4e860fd99ae8202ac8379cfff0

dinosaure:
So, if I configure my unikernel with `--ipv4=10.0.0.42/28`, my stack will be a `StackV4` and only a `StackV4` - or I need to add `--no-ipv6` (just to clear in mind the behavior)?

hannesm:
if you use `stackv4v6` in `config.ml`, you'll get a `Mirage_stack.V4V6` in `unikernel.ml` (where type ipaddr = Ipaddr.t) which respects `--ipv4=...` and `--ips=..` boot arguments (i.e. it combines the ipv4 and the ipv6 stack).

for ipv4-only unikernels, use `stackv4`, for ipv6-only unikernels `stackv6`.

PR #1189, opened 2020-09-28
mirage: use best-fit as default allocation policy, bump lower bound to 4.10.1

hannesm:
as discussed recently, the best-fit collector has a great impact on MirageOS unikernels. Also, we agreed to support the most recent two major OCaml versions only.


hannesm:
not a good idea to do sth along this lines with mirageos3, eventually mirageos4 will do this (or since 4.13 changed the default allocation policy, maybe this is not needed at all).

PR #1194, opened 2020-10-26
remove mirage-types-lwt

hannesm:
since master is meant for 4.0, let's remove the things we marked deprecated earlier. first mirage-types-lwt.


dinosaure (approved):

avsm (approved):
Looks good to me, but could we get a CHANGES entry as well telling users that it's no longer necessary as well. Helps when grepping for why the package has disappeared.

hannesm:
> but could we get a CHANGES entry as well telling users that it's no longer necessary as well

Sure. Looking at CHANGES, this is empty for the upcoming 4.0 release so far. AFAICT currently changes are collected and introduced at release time (instead of at merge time).  I find it a bit weird to change this for this PR (or within this PR), and would rather sit down -- once 4.0 is in its final stage -- and collaboratively work on the changes. Looking at the diff between the last release and the last commit of the main branch, I'm a bit scared by the sheer amount of commits / PRs that landed in between (and hope Thomas has an overview of expected breakages for users).

samoht:
yes I'm happy to help write the 4.0 changelog :-) (although I'm a bit scared at looking back in the history).

Would have been much simpler to keep adding entries to CHANGES

dinosaure:
#1232 was merge so I believed that we can close this PR, @samoht?

hannesm:
indeed

PR #1196, opened 2020-10-29
WIP conduit 3.0

hannesm:
//cc @samoht @dinosaure 

This is my attempt at the conduit 3.0 changes for the mirage utility. It does not yet work well:
- tcp and tls conduit are provided and compose well (tls conduit requires another conduit, stacking is achieved)
- http devices are removed (since `'a conduit` is useless)
- issue is that if in config.ml, I have `  foreign ~packages ~keys "Unikernel.Main" (tls_conduit @-> job)`, I don't know what to write in YYY of unikernel.ml: "module Main (C : YYY) = struct let start (proto, service) = ... end`

I'm happy if others take this as inspiration and actually get it to work. It may need some modifications to the conduit library (to expose concrete module types). It's also not clear to me how support for conduit clients should look like.


hannesm:
let's close this since conduit will be reiterated.

Issue #1198, opened 2020-11-19
Obscure failure in "mirage configure" when unrelated .opam files exist


mato:
When unrelated `.opam` files exist in the unikernel source, `mirage configure` gets confused and fails with an obscure error message, for example:

```
File "3.8.0.opam", line 1, characters 0-0:
Error: "3.8.0" is not a valid Dune project name.
mirage: unknown option `-t'.
         unknown option `--allocation-policy'.
         unknown option `--retry'.
```

This should be fixed, and/or `mirage` should emit a warning if such files exist.

hannesm:
the root of this issue is dune -- it errors out with `Error: "3.8.0" is not a valid Dune project name.` -- I'm not sure how to fix this in mirage (since any mirage subcommand uses ``dune exec --root `pwd` -- ./config.exe ..``, i.e. the dune code is executed before mirage code. Maybe report this issue to the dune repository?

PR #1202, opened 2020-11-27
Makefile: check that a single opam file is present

hannesm:
In #1198 a failure was observed if there are spurious opam files present
in the current working directory. While we can't inside of mirage handle this
properly, we can protect the `opam install -y --deps-only .` (which installs
dependencies of all present opam files) by checking that exactly one opam
file is present in the current working directory.

as noted in #1198 I don't see how we can -- inside mirage -- handle the case that dune exits due to an opam file in the working directory which name is not a dune project name.


hannesm:
that's way too brittle, and with mirageos4 on the horizon, eventually a different solution may be viable.

PR #1208, opened 2021-01-20
WIP: [Mirage 3.x] Adapt to Solo5 and ocaml-freestanding 0.7.0

mato:
This change adapts the Solo5-based targets supported by the mirage tool to the new cross-compilation toolchain provided by the combination of the Solo5 "bottom half" and ocaml-freestanding "top half" changes, both of which will be released as version 0.7.0 when ready.

Thanks to @TheLortex for doing the bulk of the work on the ocaml-freestanding changes and for being patient with my requests to do things in a way that lets us have a single ocaml-freestanding supporting both Mirage 3.x and the new dune-based Mirage 4.x (master).

Depends on:

- [ ] Solo5 "bottom half" (Solo5/solo5#488)
- [ ] ocaml-freestanding "top half" (mirage/ocaml-freestanding#90)
- [ ] mirage-solo5 0.6.5 (mirage/mirage-solo5#74)
- [ ] mirage-xen (TBA)

Questions/Notes:

1. mirage-solo5 is deliberately kept in the 0.6.x series to avoid having to cut new releases of mirage-*-solo5, all of which have an upper bound on `mirage-solo5 <= 0.7.0`. I think this is fine since it's APIs have not changed.
2. mirage-xen is not done yet, but the changes should be analogous to mirage/mirage-solo5#74.
3. do we use `ldpostflags` for anything?
4. Is there a better way to drop the `List` in usage of `find_ld` and `cc`?

Please review. If you'd like to test the full stack, the following set of pins is required:

```
opam pin add solo5.0.7.0 git+https://github.com/mato/solo5#split-packages
opam pin add solo5-config-ocaml-static.0.7.0 git+https://github.com/mato/solo5#split-packages
opam pin add solo5-bindings-hvt.0.7.0 git+https://github.com/mato/solo5#split-packages
opam pin add solo5-bindings-muen.0.7.0 git+https://github.com/mato/solo5#split-packages
opam pin add solo5-bindings-spt.0.7.0 git+https://github.com/mato/solo5#split-packages
opam pin add solo5-bindings-virtio.0.7.0 git+https://github.com/mato/solo5#split-packages
opam pin add solo5-bindings-xen.0.7.0 git+https://github.com/mato/solo5#split-packages
opam pin add ocaml-freestanding.0.7.0 git+https://github.com/TheLortex/ocaml-freestanding#mirage-4
opam pin add mirage-solo5.0.6.5 git+https://github.com/TheLortex/mirage-solo5#revert-makefile
opam pin add mirage git+https://github.com/mato/mirage#3-adapt-to-ocaml-freestanding
```

Note that the bindings packages are co-installable now, if you're testing pick the ones that you want/are available on your system.

### DIFF
--- a/lib/mirage_impl_misc.ml
+++ b/lib/mirage_impl_misc.ml
@@ -54,8 +54,22 @@ let pkg_config pkgs args =
   in
   Bos.OS.Env.set_var var (Some value) >>= fun () ->
   let cmd = Bos.Cmd.(v "pkg-config" % pkgs %% of_list args) in
-  Bos.OS.Cmd.(run_out cmd |> out_string |> success) >>| fun data ->
-  String.cuts ~sep:" " ~empty:false data
+  Bos.OS.Cmd.(run_out cmd |> out_string |> success)
+
+(* Invoke solo5-config for target and return output if successful. *)
+let solo5_config target args =
+  let target_string = match target with
+    | `Virtio -> "--target=virtio"
+    | `Muen -> "--target=muen"
+    | `Hvt -> "--target=hvt"
+    | `Genode -> "--target=genode"
+    | `Spt -> "--target=spt"
+    | `Xen | `Qubes -> "--target=xen"
+    | _ ->
+      invalid_arg "solo5_config only defined for solo5 targets"
+  in
+  let cmd = Bos.Cmd.(v "solo5-config" % target_string %% of_list args) in
+  Bos.OS.Cmd.(run_out cmd |> out_string |> success)
 
 (* Implement something similar to the @name/file extended names of findlib. *)
 let rec expand_name ~lib param =
--- a/lib/mirage_link.ml
+++ b/lib/mirage_link.ml
@@ -6,25 +6,16 @@ module Key = Mirage_key
 module Info = Functoria.Info
 
 let static_libs pkg_config_deps =
-  pkg_config pkg_config_deps [ "--static" ; "--libs" ]
+  pkg_config pkg_config_deps [ "--static" ; "--libs" ] >>| fun s ->
+  String.cuts ~sep:" " ~empty:false s
 
-let ldflags pkg = pkg_config pkg ["--variable=ldflags"]
+let ld target =
+  solo5_config target ["--ld"] >>= fun s ->
+  Rresult.R.open_error_msg (Bos.Cmd.of_string s)
 
-let ldpostflags pkg = pkg_config pkg ["--variable=ldpostflags"]
-
-let find_ld pkg =
-  match pkg_config pkg ["--variable=ld"] with
-  | Ok (ld::_) ->
-    Log.info (fun m -> m "using %s as ld (pkg-config %s --variable=ld)" ld pkg);
-    ld
-  | Ok [] ->
-    Log.warn
-      (fun m -> m "pkg-config %s --variable=ld returned nothing, using ld" pkg);
-    "ld"
-  | Error msg ->
-    Log.warn (fun m -> m "error %a while pkg-config %s --variable=ld, using ld"
-                 Rresult.R.pp_msg msg pkg);
-    "ld"
+let ldflags target =
+  solo5_config target ["--ldflags"] >>= fun s ->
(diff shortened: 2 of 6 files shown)


hannesm:
> do we use ldpostflags for anything?

not that I'm aware of, they were added (by you ;) 4 years ago (see https://github.com/mirage/mirage/commit/7b1c75310ff01e8b1765e4fb7e357e4776f318cd) but never been used. I don't think it is worth keeping them.

lib/mirage_impl_misc.ml
```diff
@@ -57,6 +57,22 @@ let pkg_config pkgs args =
...
+    | `Hvt -> "--target=hvt"
+    | `Genode -> "--target=genode"
+    | `Spt -> "--target=spt"
+    | `Xen | `Qubes -> "--target=xen"
+    | _ ->
+      invalid_arg "solo5_config only defined for solo5 targets"
+  in
+  let cmd = Bos.Cmd.(v "solo5-config" % target_string %% of_list args) in
+  Bos.OS.Cmd.(run_out cmd |> out_string |> success) >>| fun data ->
+  String.cuts ~sep:" " ~empty:false data
```
hannesm, line 74: if we drop the last line `String.cuts ...` here (together with the `>>| fun data ->` in the previous line, the result of `solo5_config` will be a `string` (instead of a `string list`).

lib/mirage_link.ml
```diff
@@ -8,23 +8,11 @@ module Info = Functoria.Info
 let static_libs pkg_config_deps =
   pkg_config pkg_config_deps [ "--static" ; "--libs" ]
 
-let ldflags pkg = pkg_config pkg ["--variable=ldflags"]
+let find_ld target = solo5_config target ["--ld"]
 
-let ldpostflags pkg = pkg_config pkg ["--variable=ldpostflags"]
+let ldflags target = solo5_config target ["--ldflags"]
```
hannesm, line 13: then we can re-add that `>>| fun data -> String.cuts ...` for ldflags here

lib/mirage_link.ml
```diff
@@ -34,21 +22,20 @@ let link info name target _target_debug =
...
-    ldpostflags bindings >>= fun ldpostflags ->
+    find_ld target >>= fun ld ->
+    ldflags target >>= fun ldflags ->
+    (* ldpostflags bindings >>= fun ldpostflags -> *)
     let extension = Mirage_configure_solo5.bin_extension target in
     let out = name ^ extension in
-    let ld = find_ld bindings in
     let linker =
-      Bos.Cmd.(v ld %% of_list ldflags % "_build/main.native.o" %
+      Bos.Cmd.(v (List.hd ld) %% of_list ldflags % "_build/main.native.o" %
```
hannesm, line 35: and don't need a `(List.hd ld)` here.

lib/mirage_link.mli
```diff
@@ -1,9 +1,9 @@
 open Rresult
 
 val static_libs : string -> (string list, [> R.msg ]) result
-val ldflags : string -> (string list, [> R.msg ]) result
+val ldflags : Mirage_key.mode -> (string list, [> R.msg ]) result
 val ldpostflags : string -> (string list, [> R.msg ]) result
 
-val find_ld : string -> string
+val find_ld : Mirage_key.mode -> (string list, [> R.msg ]) result
```
hannesm, line 7: and this wold be `string` on the right hand side again :)

mato:
@hannesm Thanks for the suggestions and help; take a look at the latest version. I've re-worked the command line fragment construction to use `Bos.Cmd.of_string` where appropriate, among other things this means that:

1. We rely on the `Bos` tokenizer rather than a plain conversion to `List` splitting on spaces; this seems sensible.
2. If `solo5-config [--cc | --ld]` ever change to output a command with arguments, then things will continue to work.
3. The final `Cmd` construction feels cleaner to me this way.

Let me know what you think.

hannesm:
looks good to me

mato:
This needs to be re-worked for Solo5/solo5#494 and the corresponding upcoming ocaml-freestanding PR by @TheLortex. Closing this one, will come back with a new PR when ready.

PR #1210, opened 2021-02-02
Worse is better

Drup:
When we created functoria, I had this fancy idea of using graphs everywhere.
With hindsight, It wasn't such a good idea.

This commit remove all the formal graph structure, and use DAGs implemented as terms. The type of `Impl.t` is expanded to cover all we need, and most of the pre-computations are done on `Impl.t` directly. `Device_graph.t` now only contain devices, and a new function `Impl.eval` removes everything except the devices, which are used for code output.

All this removes quite a bit of code, and more importantly, quite a lot of `assert false` which have been the cause of numerous issues ...

I only have tested this very loosely (using the testsuite), so I would appreciate if someone can run this on their favorite unikernel. The code output *will* be different (because the order and the ids are not particularly stable), but it should be equivalent.

Unfortunately, I had to touch `Lib` in the process, in order to insert the calls to `Impl.eval` appropriately. I seem to have changed something on the way (see the test change in the last commit). I'm not sure what.

The dot output is not implemented yet, I'll do that next.


lib/functoria/impl.ml
```diff
@@ -53,84 +82,316 @@ let main ?packages ?packages_v ?keys ?extra_deps module_name ty =
...
+  type tbl = abstract Tbl.t
+  let create () = Tbl.create 50
+  let add tbl a b = Tbl.add tbl (abstract a) (abstract b)
+  let get (type a) tbl (a : a t) =
+    if Tbl.mem tbl @@ abstract a then
+      let Abstract b = Tbl.find tbl (abstract a) in
+      Some (Obj.magic b : a t)
+    else
+      None
+end
```
Drup, line 203: This is rather dirty. If anyone has a proposal to make that safe in a concise manner. I'm all ears. So far my plan is to use equality witnesses.

dinosaure:
I'm not really aware about this codebase but:
1) Travis CI try to compile `mirage-skeleton`, as far as Travis CI is green, we can say that this PR is fine for most of our cases
2) I can play a bit with some of my unikernels
3) it can be interesting to integrate [hannesm/functoria-playground](https://github.com/hannesm/functoria-playground) (/cc @hannesm) and see if such cases work (I suspect that these _unikernels_ reached the same `assert false` as me)
4) I'm wondering about MirageOS 4 and MirageOS 3.10. Currently, the merge will be apply on MirageOS 4 and we probably backport this patch on MirageOS 3 (branch 3 but I'm not sure)

Otherwise, `hmap` is missing.

lib/functoria/impl.ml
dinosaure: Please use equality witnesses 👍

Drup: Done!

Drup:
dot output is implemented, and I added equality witnesses to implement the polymorphic hash consing on `impl`s in a safe way. It seems to work well.

We still have 3 walks over `impl`s. The typing gets in the way of simplifying/factorizing further. I though about getting rid of preserving sharing on `impl`s, and only do that on device graphs (where it's essential, to avoid double inits), but that would result in horrible dots and compromise collections. Maybe someone else with fresh eyes will have an idea.

The last commit changes a test: the last version of dune doesn't create `.merlin` files by default anymore.

I think it's ready for review.

dinosaure:
Thanks, I will do the end-to-end test by hands on some unikernels with UNIX and `hvt` and let Travis CI tell to us that all works :+1:. I will figure out how to patch-back your PR into MirageOS 3.10 (and cut a release with that)

samoht:
I like the approach very much. Give me a few days to review the code in details before merging :-)

lib/functoria/device.mli
```diff
@@ -51,6 +51,9 @@ val pp : 'b Fmt.t -> ('a, 'b) t Fmt.t
 val equal : ('a, 'b) t -> ('c, 'd) t -> bool
 (** [equal] is the equality function for devices. *)
 
+val witness : ('a, _) t -> ('b, _) t -> ('a, 'b) Misc.Eq.eq
```
samoht, line 54: Maybe move `Eq` to a top-level module, that's not very nice to have `Misc` leaking in the public API.

lib/functoria/impl.ml
```diff
@@ -53,84 +60,376 @@ let main ?packages ?packages_v ?keys ?extra_deps module_name ty =
...
 
-let with_left_most_device ctx t { f } =
+  let get (type a) tbl (oldv : a t) : a t option =
+    if Tbl.mem tbl @@ abstract oldv then
+      let (Abstract newv) = Tbl.find tbl (abstract oldv) in
+      match equal oldv newv with Eq -> Some newv | NotEq -> None
+    else None
+end
+
+let simplify ~partial ~context (Abstract t) =
```
samoht, line 201: Can you add a bit of documentation here? Especially the logics to decide whether to use default condition values depending  on `partial` and which key is in `context`. It always take me a while to remember why it's working like this :-)

lib/functoria/impl.ml
```diff
@@ -53,84 +60,376 @@ let main ?packages ?packages_v ?keys ?extra_deps module_name ty =
...
+        HC.add tbl impl acc;
+        acc
+  and aux_abstract (Abstract a) = Abstract (aux a)
+  and aux_tl : type a v. (a, v) tl -> (a, v) tl = function
+    | Nil -> Nil
+    | Cons (h, t) -> Cons (aux h, aux_tl t)
+  in
+  Abstract (aux t)
+
+let eval ~context (Abstract t) =
```
samoht, line 238: That `eval` function is so simple now :-) 💯 ❤️

lib/functoria/impl.mli
```diff
@@ -81,17 +84,20 @@ module Tbl : Hashtbl.S with type key = abstract
...
-(** The type for iterators in [if] nodes. *)
-
-type 'a f_app = f:'a -> x:'a -> 'a
-(** the type for iterators on [app] nodes. *)
-
 val with_left_most_device : Key.context -> _ t -> 'a f_dev -> 'a
 (** [with_left_most_device ctx t f] applies [f] on the left-most device in [f].
     [If] node are resolved using [ctx]. *)
 
-val map :
```
samoht, line 94: did you remove `map` on purpose?

lib/functoria/lib.ml
```diff
@@ -47,32 +54,28 @@ module Config = struct
...
+  let v ?(keys = []) ?(packages = []) ?(init = []) ~build_cmd ~src name jobs =
     let name = Name.ocamlify name in
-    let jobs = Device_graph.create main_dev in
     let packages = Key.pure @@ packages in
+    let jobs = Impl.abstract jobs in
     let keys = Key.Set.(union (of_list keys) (get_if_context jobs)) in
     { packages; keys; name; init; jobs; build_cmd; src }
 
-  let eval ~partial context
+  let simplify ~partial context
```
samoht, line 64: you probably don't need to rename that function

lib/functoria/lib.ml
```diff
@@ -108,8 +111,8 @@ module Make (P : S) = struct
 
   module Log = (val Logs.src_log src : Logs.LOG)
 
-  let eval_cached ~partial ~with_required ~output ~cache context t =
-    let info = Config.eval ~partial context t in
+  let simplify_cached ~partial ~with_required ~output ~cache context t =
```
samoht, line 114: same here, `eval_cached` still works fine on that context I think

lib/functoria/lib.ml
```diff
@@ -119,16 +122,18 @@ module Make (P : S) = struct
...
     in
     Cmdliner.Term.(pure f $ context)
 
   (* FIXME: describe init *)
   let describe (t : _ Cli.describe_args) =
-    let (_, jobs), _ = t.Cli.args.context in
+    let { Config.jobs; _ } = t.args.Cli.context in
     let f fmt =
-      Fmt.pf fmt "%a\n%!" (if t.dot then Config.pp_dot else Config.pp) jobs
+      Fmt.pf fmt "%a\n%!" (if t.dot then Config.pp_dot else Fmt.nop) jobs
```
samoht, line 136: Is it possible to continue to show something sensible in text mode (instead of `Fmt.nop`) ?

lib/functoria/misc.ml
```diff
@@ -48,3 +48,58 @@ module Name = struct
...
+   fun r s ->
+    let module R = (val r : ID with type t = r) in
+    let module S = (val s : ID with type t = s) in
+    match R.Tid with S.Tid -> Eq | _ -> NotEq
+
+  let equal' a b = to_bool @@ equal a b
+
+  let pp (type a) ppf ((module M) : a t) = Fmt.int ppf M.id
+
+  let hash (type a) ((module M) : a t) = M.id
```
samoht, line 104: `hash` is a bit of a weird name here -- as we ensure that we return a unique ID. Maybe just call it `id`? 

Or maybe don't use it for equality (as it is done in one other part of the code -- I think when trying to equalise devices).

lib/functoria/misc.mli
```diff
@@ -35,3 +35,25 @@ end
 module Name : sig
   val ocamlify : string -> string
 end
+
+module Eq : sig
+  type (_, _) eq = Eq : ('a, 'a) eq | NotEq : ('a, 'b) eq
+
+  val andeq : ('a, 'b) eq -> ('c, 'd) eq -> ('a * 'c, 'b * 'd) eq
```
samoht, line 42: do we need to expose `andeq` and `andb` ?

test/functoria/test_graph.ml
```diff
@@ -10,34 +10,36 @@ let z = Device.v "Bar" job
...
 
-let ident s i x = Fmt.strf "%s__%d" s (i + x)
+let ident s i = Fmt.strf "%s__%d" s i
 
 let test_var_name () =
   let id = id () in
-  Alcotest.(check string) "x" (ident "foo_bar" id 1) (var_name x);
-  Alcotest.(check string) "y" (ident "x_y" id 2) (var_name y);
-  Alcotest.(check string) "z" (ident "bar" id 3) (var_name z)
+  Alcotest.(check string) "x" (ident "foo_bar" id) (var_name x);
```
samoht, line 35: I am confused: why is this different now? How do we ensure devices have unique names (and variable names)?

samoht:
Ok so: many thanks for doing this, it's a big cleanup on the existing code and simplifies a lot of things. I'm looking forward to merge it asap :-)

The only thing that I'm not super sure is the changes in `var_name`. I guess the previous generated IDs were the for all the nodes in the impl graphs while the new ones are only for the nodes in the device graph? But why do the test are re-using the same IDs in different variables?

lib/functoria/impl.mli
Drup: Yes, it's not used outside the module now.

lib/functoria/lib.ml
Drup: Fair enough. I though the name was a bit misleading, since it doesn't call `eval`.

lib/functoria/lib.ml
Drup: `Config.pp` was defined as `Device_graph.pp` ... which was `Fmt.nop`. It was printing nothing all along.

Drup: We could technically use `Impl.pp` here, but that's more of a debug output than something you want to show to users.

lib/functoria/misc.mli
Drup: They are used in the implementation of `Impl.equal`

lib/functoria/impl.ml
Drup: I flipped the bool to `simplify ~full` at its slighly more intuitive, and added a comment.

lib/functoria/device.mli
Drup: Done, it's called `Typeid`

lib/functoria/misc.ml
Drup: Done, renamed to `id`.

lib/functoria/misc.mli
Drup: aaand, they are gone.

test/functoria/test_graph.ml
Drup: Okay, so the point of the test was indeed rendered a bit moot. I modified it partially: Now, we create a full implementation, pass it through the eval function, ask for a topo sort and tests the IDs on that. What do you think of that ?

Drup:
I think I solved all your points, I also added documentations in various places.

samoht (approved):

samoht:
All green! @dinosaure did you test it on mirage-skeleton already? Happy to merge that PR once that works ok.

TheLortex:
I've set up a temporary mirage CI pipeline to test the PR: https://ci.mirage.io/main/pipeline.html
The outcome is that mirage-skeleton fails on the same places as with the main branch (except for issues that should be fixed if you rebase your PR on master to pick up the latest fixes).

Drup:
@TheLortex Thanks! From what I can see, those failure are not related to the current PR. Is that correct ? Otherwise, I just rebased, so we can have a clearer picture.

TheLortex:
Yes, now we can see that the errors are the same on `master` and with your PR, so this is looking good !

dinosaure:
Ready to merge, we will discuss internally if we want to patch-back that in MirageOS 3 or not, thanks your great work both 👍 !

Drup:
Thanks, I'm happy it turned out so well, this code has been bothering me for a few years now, and now hopefully it will be more reliable, and people will be able to look and modify it!

lib/functoria/engine.ml
```diff
@@ -81,28 +83,24 @@ let files info t stage =
...
 
 let find_all_devices info g i =
   let ctx = Info.context info in
   let id = Impl.with_left_most_device ctx i { f = Device.id } in
-  let p = function
-    | Device_graph.Dev d -> Device.id d = id
-    | App | If _ -> false
+  let f x l =
+    let (Device.Graph.D { dev; _ }) = x in
+    if Device.id dev = id then x :: l else l
```
samoht, line 94: @Drup are you sure we really want to select all devices with the same id here? I don't remember why we need to need to find "all devices" here. The `connect` code mainly try to associate the init devices by their counterpart in the graph.

Drup: `find_all_devices` was added rather recently with the `~init` parameter. I'm personally not super convinced we need to code it like that, but I wasn't sure, so I preserved the same semantics as was present before.

Since we are very careful never to duplicate any devices, I think this list always ever have one element ...

samoht: but what happens if you add the same device twice in the graph?

samoht: That function should probably just be a call to `with_left_most_device` ...

PR #1211, opened 2021-02-05
Fix arp

TheLortex:
I guess this was a typo, the change for `arp 2.3.0` is that `arp-mirage` becomes `arp.mirage`.


hannesm:
yes indeed this was a typo, thanks for the PR. LGTM.

PR #1212, opened 2021-02-18
Move impls and targets in separate folders

TheLortex:
As a preparation for the big switch to dune, here's a PR that re-organises the `lib/mirage/` code. It separates 4 components into libraries:
- `mirage.key`: key definitions
- `mirage.impl`: device implementations
- `mirage.target`: target-specific rules
- `mirage`: the mirage library

There's no API or behavior change.  

## Why

I based my mirage 4 work on this PR, because it increases structure and readability. The main advantage is that target-specific instructions (configure/build-time) are clearly isolated, and the requirements to create a new target are unified under a single signature instead of being scattered across files.


dinosaure:
LGTM for me, however OCaml 4.12 does not work due to a non-erased optional argument, it should be a good opportunity to fix that in this PR (where it's mostly about styles).

TheLortex:
Thank you! Added the fix for 4.12, and also fixed the `lint-opam` job in the ocaml-ci check.

dinosaure (approved):

dinosaure:
To move a bit forward about MirageOS 4 and because it should not change anything at the end (only internals layout), I take the opportunity to merge this PR. An other point about CI, due to the work of @TheLortex, @avsm and @talex5 on https://ci.mirage.io/ we will trust on them now and `ocaml-ci` instead of Travis CI.

PR #1213, opened 2021-02-18
Fix ipv6 key

TheLortex:
This PR fixes the ipv6 key, which has the wrong label (`ipv6_prefix`) so the generated code wouldn't build. 

The `ping6` mirage-skeleton unikernel is failing because of this:
```
File "key_gen.ml", line 103, characters 48-78:
103 |   (Functoria_runtime.Arg.opt (Cmdliner.Arg.some Mirage_runtime.Arg.ipv6_prefix) 
```


hannesm:
LGTM, sorry for introducing this issue.

PR #1214, opened 2021-02-23
Adapt to functoria 3.2

samoht:
Adapt to https://github.com/mirage/functoria/pull/197

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -232,7 +232,7 @@ module Project = struct
   let ignore_dirs = Mirage_build.ignore_dirs
 
   let create jobs = impl @@ object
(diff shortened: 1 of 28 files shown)


dinosaure:
I would like to know if it still is interesting to integrate it into the next release of MirageOS 3. Currently, from what I see, https://github.com/mirage/functoria/pull/197 is a bit too big (at least, the REVDEPS shows up a deep change). So I'm not sure that we want to spend our time on that or not (when the patch is already integrated for MirageOS 4).

samoht:
it's probably not worth pushing this further given the introduced API change.

dinosaure:
Ok, thanks!

PR #1215, opened 2021-03-03
Use let* syntax for the Action API

TheLortex:
Instead of using infix operators, use custom `let*` / `let+` operators. This allows code to be written in a more _direct style_. 

(based on https://github.com/mirage/mirage/pull/1212)


dinosaure:
Can you rebase this PR from master to clear a bit the diff?

PR #1220, opened 2021-04-14
Use cram tests in place of the dune-generated expect tests

TheLortex:
Cram tests is a dune feature since 2.7 (see https://dune.readthedocs.io/en/stable/tests.html#cram-tests). It is a format in which expectation tests can be written, its biggest advantage being simplicity: tests consists in shell commands, followed by the expected output. 

The motivation behind that is that currently, it's not straightforward to make the link between a `.expected` file and which command is supposed to produce that output. (it's often defined in `gen.ml`).

Hopefully, the cram testing framework should help us maintain the tests and track changes more easily.


dinosaure:
CI seems good (minor `ocamlformat` & Travis CI), I will take a look but I completely agree with this PR 👍 Nice polishing!

test/mirage/help/describe.t
```diff
@@ -0,0 +1,332 @@
...
+             Represented as circles. Branches are dotted, and the default
+             branch is in bold.       Configurables
+                                          Represented as rectangles. The order
+                                          of the output arrows is the order of
+                                          the functor arguments.       
+                                                                Data
+                                                                dependencies
+                                                                           
+                                                                Represented as
+                                                                dashed arrows.
```
dinosaure, line 190: I think that such layout is not expected and it's probably a bug. We should make an issue about that because I don't think that your PR introduce this bug.

dinosaure:
So I did not check that if you miss something but the PR seems fine for me. Can you just make an issue about the man output (see [my comment](https://github.com/mirage/mirage/pull/1220#discussion_r613959996))?

test/mirage/help/describe.t
TheLortex: Yes, looking at the diff the layout issue was already present in the `test/mirage/help/describe.expected` file.

dinosaure:
Thanks for this work, I will merge it :+1:.

PR #1240, opened 2021-10-18
Modify app_info to use `opam monorepo`

### LINKED ISSUE #1236: mirage4: unreliable `app_info` device
In Mirage3, `app_info` uses `opam list --resolve ...` to find which opam packages are installed (at configuration time).

I am not sure this will work for MirageOS 4 as the dependencies are now downloaded to be compiled at the same time as unikernel code. The exact dependencies to download are not computed by `opam` directly (but by a plugin), so the real list of actual packages might be different from `opam list --resolve`.

samoht:
Fix #1236

Note: this is not tested yet.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,11 @@ Breaking changes:
   $ mirage configure --extra-repo https://github.com/dune-universe/opam-overlays
   ````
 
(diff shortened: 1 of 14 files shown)


samoht:
I've changed `app_info` to only query linked libraries via `dune-build-info`. All the other solutions are not reliable enough.

Issue #1249, opened 2021-10-26
Add use public names to binaries, and use the right names in targets


samoht:
Right now, the target names are ocamlliifed:

```
$ cat tutorial/hello-key/dune.build
(rule
 (target hello_key)
 (enabled_if (= %{context_name} "default"))
 (action
  (copy main.exe %{target})))
```

This should be `hello-key` to match the original unikernel name.

Also, to run this one, one has to type:

```
$ dune exec -- tutorial/hello-key/main.exe
```

whereas it would be better to simply write:

```
$ dune exec -- hello-key
```

Not that this probably wouldn't work with other build artefacts?

dinosaure:
Due to #1251, should we close this PR?

PR #1251, opened 2021-10-27
Fix ocamlification of unikernel names

TheLortex:
This PR fixes the first question of https://github.com/mirage/mirage/issues/1249

What it does is remove ocamlification of the name in `Config.v` and introduce _opamification_ for Makefile generation and opam file generation. _opamification_ is a less constrained transformation from arbitrary names to opam-friendly package names. 

First commit adds tests to highlight the changes. Second commit actually perform the changes.


dinosaure:
It seems that the CI fails to compile `base` (which was recently released, I think we need to bump an other version on our overlays).

TheLortex:
I'm on it

Issue #1250, opened 2021-10-26
simplify the generated Makefile


samoht:
The generated should follow what is now done on `mirage skeleton`, e.g. it would be nice to have the following workflow:

```
$ mirage configure ...
$ make lock # this one is a bit involved as it requires adding the right opam overlays
$ make depends # this one is a bit involved as it requires calling `opam monorepo depext` + `opam install` on the opam switch
$ opam monorepo pull # == make pull
$ dune build # == make build
$ mirage clean # == make clean
```

We are not very from this, but there are still a bit of differences that'd be nice to fix.

PR #1252, opened 2021-10-27
Simplify makefile

TheLortex:
On top of https://github.com/mirage/mirage/pull/1251

Implement the workflow suggested in https://github.com/mirage/mirage/issues/1250.

* Separate Makefile `depends` target in three steps: `lock`, `depends` and `pull`.
* `make all` does all of them + `build` 
* The new `opam monorepo depext` command is used to install depexts from the lockfile

One question remains: what should be the behavior of `make` by default: `make all` or `make build` ? For now it's `make all`.

### DIFF
--- a/lib/functoria/lib.ml
+++ b/lib/functoria/lib.ml
@@ -55,7 +55,6 @@ module Config = struct
     Key.Set.fold f all_keys skeys
 
   let v ?(keys = []) ?(packages = []) ?(init = []) ~build_cmd ~src name jobs =
-    let name = Name.ocamlify name in
     let packages = Key.pure @@ packages in
     let jobs = Impl.abstract jobs in
     let keys = Key.Set.(union (of_list keys) (get_if_context jobs)) in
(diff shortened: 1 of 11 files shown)


samoht:
I'm fine with `make all` being the default target. People call still use `dune build` to have incremental builds :-)

dinosaure:
I'm ok with this PR, I would like to get a review from @hannesm who asked about:
>  is opam depext still used, I thought some recent opam monorepo had a builtin depext command? If it is not used anymore, the documentation of --no-depext should be updated. :)

hannesm:
this is very fine to merge (I have not looked into the code, but from the high-level description this is wonderful).

IIUC there are now additional Makefile targets, and producing a unikernel from scratch involves:
- mirage configure 
- make lock
- make depend
- make pull
- (mirage) dune build

is this correct? is the order of "make lock" and "make depend" relevant? alternatively, one can use mirage configure followed by make all.

TheLortex:
> producing a unikernel from scratch involves:
> 
>     * mirage configure
>     * make lock
>     * make depend
>     * make pull
>     * (mirage) dune build
> 
> 
> is this correct? 

Yes it is. I'm thinking that to keep things simple, we could have the following:

* `make lock`, `make install-switch`, `make pull`: the three separate operations

* `make depend`: all-in-one

> is the order of "make lock" and "make depend" relevant? alternatively, one can use mirage configure followed by make all.

Currently `make depend` also takes care of installing the lockfile's external dependencies using `opam monorepo depext`. 
That's why it needs to be run after `make lock`.

dinosaure:
LGTM, but we should fix the CI.

dinosaure:
So I'm ok to merge this PR but it seems that the last `opam-monorepo` broke something and the CI can not build unikernels anymore.

dinosaure:
Thanks!

PR #1256, opened 2021-12-06
mirage 3: allow newer mirage-clock

hannesm:


TheLortex:
CI failure looks unrelated:
```
ld: /home/opam/.opam/4.13/lib/mirage-crypto/libmirage_crypto_freestanding_stubs.a: file not recognized: file truncated
```
I couldn't reproduce the issue but I'm continuing to investigate..

I think this is safe to merge because of the CI success on the Mirage 4 side.

PR #1266, opened 2022-01-12
Align the context_name with the toolchain used

dinosaure:


dinosaure:
The error on Xen is related to the new release of `ocaml-freestanding` which does not provide `pkg-config` anymore. I think it's fine to merge, I will try to close all PRs today about `mirage-xen`/`mirage-solo5` packages.

PR #1268, opened 2022-01-21
Fix url given by the src field into the OPAM file

dinosaure:
This PR wants to fix an warning produced by `opam` when we pin the generated package. Note that I added a new dependency `uri` but it seems fine. WDYT @mirage/core?


dinosaure:
This PR requires 39 dependencies instead of 31 (`bigstringaf` + `bigarray-compat`, `pecu`, `angstrom`, `base64` & `ipaddr` + `uri` and `emile`).

reynir:
This introduces an extra `/` in the beginning of the path of http(s) URIs. I'll see if I can come up with a fix.

PR #1270, opened 2022-01-26
Rename shell variable MIRAGE_EXTRA_REPO to EXTRA_REPOS (we can add multiple repositories)

dinosaure:


lib/functoria/cli.ml
```diff
@@ -89,9 +89,9 @@ let extra_repo doc_section =
     let pp ppf (name, repository) = Fmt.pf ppf "%s:%s" name repository in
     Arg.conv (parser, pp)
   in
-  let env = Arg.env_var "MIRAGE_EXTRA_REPO" in
+  let env = Arg.env_var "MIRAGE_EXTRA_REPOS" in
   let doc =
-    Arg.info ~docs:doc_section ~docv:"URL" ~env
+    Arg.info ~docs:doc_section ~docv:"(<name>:<url>)+" ~env
```
TheLortex, line 94: `(<name>:<url>)+` doesn't really make it clear that the repos are comma-separated

TheLortex (changes requested):
I agree that it should be plural, but in that case the command line argument name should also be changed (`--extra-repo` => `--extra-repos`)

lib/functoria/cli.ml
dinosaure: hmmhm, do you prefer: `<name>:<url>(','<name>:<url>)*`?

TheLortex: Maybe `NAME1:URL1,NAME2:URL2,...` ?

TheLortex (approved):

PR #1272, opened 2022-01-26
Prepare v4.0.0~beta1

dinosaure:
/cc @mirage/core

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-### v4.0.0+beta1 (2022-01-26)
+### v4.0.0~beta1 (2022-01-26)
 
 Refactor build process to use [Dune](https://dune.build/) build system. The
 motivation is to drop `ocamlbuild`-induced technical debt and to obtain
@@ -28,10 +28,10 @@ Main changes:
   (or cross-compiler) toolchain, as defined by `findlib`, it can be tuned with
   environment variables and custom flags for the OCaml or underlying C compiler.
 
-* The usual workflow: `mirage configure && make depends && mirage build` does
-  not change however. Generated files are located into the new directory
-  `./mirage` (OPAM files, `main.ml`, `key_gen.ml` or `manifest.json`) and
-  the final artifact is created into the `./dist` directory.
+* The usual workflow `mirage configure && make depends && mirage build` does
+  not change. However, files are now generated in the `./mirage` directory
+  (OPAM files, `main.ml`, `key_gen.ml` or `manifest.json`), and
+  the final artefact is created in the `./dist` directory.
 
 Breaking changes:
 
@@ -39,22 +39,20 @@ Breaking changes:
   systems may be sandboxed in a Dune package, but the recommended way is to
   switch to `dune`. A set of packages have been ported to dune but the changes
   are not upstreamed, they are available in the `dune-universe`
-  [overlay repository](https://github.com/dune-universe/opam-overlays). A
-  specific package must be _dunified_ in a _certain_ way to allow the
-  cross-compilation: `zarith`. It is maitained by the MirageOS core team and
-  available into our own
+  [overlay repository](https://github.com/dune-universe/opam-overlays).
+  The MirageOS core team maintains packages that do not support cross-compilation yet in
   [mirage overlay repository](https://github.com/dune-universe/mirage-opam-overlays).
-  These repositories are configured by default.
-  To remove them, use `mirage configure --no-extra-repo`.
+  The mirage tool uses these two opam overlays by default. To stop using these (and use the default packages 
+  provided by opam), use `mirage configure --no-extra-repo`.
 
 * `Functoria_runtime.info` and `Mirage_runtime.info` now list all the libraries
   that are statically linked against the unikernel. The `packages` field have
   been removed and the `libraries` field is now accurate and contains the
   versions computed by `dune-build-info`.
 
-* The DSL to describe devices into the `config.ml` was updated. We don't use
-  anymore objects and we replace it by an usage of `Mirage.impl` which expects
-  same _fields_ than before.
+* Update the DSL to describe devices into the `config.ml`.  We don't use
+  objects anymore, and we replace it with the usage of `Mirage.impl` that
+  expects the same _fields_ as before.
 
 ### v3.10.8 (2021-12-17)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-### v4.0.0~beta1 (2022-01-26)
+### v4.0.0~beta1 (2022-01-29)
 
 Refactor build process to use [Dune](https://dune.build/) build system. The
 motivation is to drop `ocamlbuild`-induced technical debt and to obtain
@@ -19,7 +19,8 @@ Main changes:
   `opam-monorepo` tool. This tool reads the `<unikernel>-monorepo.opam` file and
   make use of the opam solver to compute the transitive dependency set, saves
   that as a _lockfile_, and fetch sources in the `duniverse/` subfolder.
-  More info on the [Github repository](https://github.com/ocamllabs/opam-monorepo).
+  More info on the
+  [Github repository](https://github.com/ocamllabs/opam-monorepo).
 
 * The compilation scheme use `dune`'s concept of a _workspace_: it's a set of
   libraries that are built together using the same _context_. For each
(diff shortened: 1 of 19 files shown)


CHANGES.md
```diff
@@ -28,25 +28,34 @@ Main changes:
...
+  [mirage overlay repository](https://github.com/dune-universe/mirage-opam-overlays).
+  These repositories are configured by default.
+  To remove them, use `mirage configure --no-extra-repo`.
 
 * `Functoria_runtime.info` and `Mirage_runtime.info` now list all the libraries
   that are statically linked against the unikernel. The `packages` field have
   been removed and the `libraries` field is now accurate and contains the
   versions computed by `dune-build-info`.
 
+* The DSL to describe devices into the `config.ml` was updated. We don't use
```
samoht, line 55:
```suggestion
* Update the DSL to describe devices into the `config.ml`.  We don't use
```

CHANGES.md
```diff
@@ -28,25 +28,34 @@ Main changes:
...
+  anymore objects and we replace it by an usage of `Mirage.impl` which expects
```
samoht, line 56:
```suggestion
  objects anymore, and we replace it with the usage of `Mirage.impl` that
```

CHANGES.md
```diff
@@ -28,25 +28,34 @@ Main changes:
   (or cross-compiler) toolchain, as defined by `findlib`, it can be tuned with
   environment variables and custom flags for the OCaml or underlying C compiler.
 
+* The usual workflow: `mirage configure && make depends && mirage build` does
```
samoht, line 31:
```suggestion
* The usual workflow `mirage configure && make depends && mirage build` does
```

CHANGES.md
```diff
@@ -28,25 +28,34 @@ Main changes:
...
+  not change however. Generated files are located into the new directory
```
samoht, line 32:
```suggestion
  not change. Generated files are located in the new directory
```

CHANGES.md
```diff
@@ -28,25 +28,34 @@ Main changes:
...
+  `./mirage` (OPAM files, `main.ml`, `key_gen.ml` or `manifest.json`) and
```
samoht, line 33:
```suggestion
  (OPAM files, `main.ml`, `key_gen.ml` or `manifest.json`), and
```

CHANGES.md
```diff
@@ -28,25 +28,34 @@ Main changes:
...
+  the final artifact is created into the `./dist` directory.
```
samoht, line 34:
```suggestion
  the final artefact is created in the `./dist` directory.
```

CHANGES.md
```diff
@@ -28,25 +28,34 @@ Main changes:
...
-  $ mirage configure --extra-repo https://github.com/dune-universe/opam-overlays
-  ````
+  are not upstreamed, they are available in the `dune-universe`
+  [overlay repository](https://github.com/dune-universe/opam-overlays). A
+  specific package must be _dunified_ in a _certain_ way to allow the
+  cross-compilation: `zarith`. It is maitained by the MirageOS core team and
+  available into our own
+  [mirage overlay repository](https://github.com/dune-universe/mirage-opam-overlays).
+  These repositories are configured by default.
+  To remove them, use `mirage configure --no-extra-repo`.
```
samoht, line 48:
```suggestion
  [overlay repository](https://github.com/dune-universe/opam-overlays).
  The MirageOS core team maintains packages that do not support cross-compilation yet in
  [mirage overlay repository](https://github.com/dune-universe/mirage-opam-overlays).
  The mirage tool uses these two opam overlays by default. To stop using these (and use the default packages 
  provided by opam), use `mirage configure --no-extra-repo`.
```

CHANGES.md
```diff
@@ -1,4 +1,4 @@
-### v4.0.0 (unreleased)
+### v4.0.0+beta1 (2022-01-26)
```
samoht, line 1:
```suggestion
### v4.0.0~beta1 (2022-01-26)
```

functoria-runtime.opam
```diff
@@ -13,7 +13,6 @@ license:      "ISC"
 tags:         ["org:mirage"]
 
 build: [
-  ["dune" "subst"] {pinned}
```
hannesm, line 16: Uhm, I'm not in favour of this change. The execution of `dune subst` in a development build helps to track down version numbers (otherwise `%%VERSION%%` is reported). Can we please keep that as is? What is the reasoning behind that change?

The line should be (otherwise some linter complain):
`["dune" "subst"] {dev}`

mirage.opam
```diff
@@ -28,6 +28,7 @@ depends: [
   "alcotest" {with-test}
   "fmt" {with-test}
 ]
+available: opam-version >= "2.1.0"
```
hannesm, line 31: isn't this (or the functoria one above) sufficient? I'd expect the `-runtime` packages to _not_ require any specific opam version.

functoria-runtime.opam
dinosaure: For sure :+1:.

mirage.opam
dinosaure: yeah, I can delete that for `-runtime` package. I did a systemic patch :+1:.

functoria-runtime.opam
hannesm: for the release in opam-repository, this does not matter (since the watermarks are substituted during the release process anyways) -- but for our main branch I find the `dune subst` really helpful. Thanks a lot :)

dinosaure:
Hmmhmm, that's a shame that `opam monorepo` was not able to downgrade `awa` to `0.0.5` and let us to compile unikernels even if it does not use the last version of `awa`. @TheLortex do you have any clues about this solver behavior?

functoria-runtime.opam
```diff
@@ -13,22 +13,19 @@ license:      "ISC"
...
 ]
 
 depends: [
   "ocaml" {>= "4.08.0"}
-  "dune" {>= "2.0.0"}
+  "dune" {>= "2.8.0"}
   "cmdliner" {>= "0.9.8"}
-  "fmt"
+  "fmt" {>= "0.9.0"}
   "functoria" {with-test & >= "2.2.0"}
```
hannesm, line 26:
```suggestion
  "functoria" {with-test & = version}
```

mirage-runtime.opam
```diff
@@ -11,16 +11,17 @@ tags:         ["org:mirage" "org:xapi-project"]
...
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
 ]
 
 depends: [
   "ocaml" {>= "4.08.0"}
-  "dune" {>= "2.0.0"}
+  "dune" {>= "2.8.0"}
   "ipaddr"             {>= "5.0.0"}
   "functoria-runtime"  {= version}
+  "functoria"          {= version & with-test}
```
hannesm, line 24: is this really needed? do we have mirage-runtime tests?

TheLortex:
> Hmmhmm, that's a shame that `opam monorepo` was not able to downgrade `awa` to `0.0.5` and let us to compile unikernels even if it does not use the last version of `awa`. @TheLortex do you have any clues about this solver behavior?

The issue isn't about `awa`, but the dependency from `functoria-runtime` to `fmt.0.9.0`, which is not present in opam-overlays:
```
              - functoria-runtime -> (problem)
                   Rejected candidates:
                     functoria-runtime.4.0.0: Requires fmt >= 0.9.0
```

mirage-runtime.opam
dinosaure: I'm not sure indeed - do you prefer to delete (on both sides, `opam-repository` and this PR)?

dinosaure:
> The issue isn't about `awa`, but the dependency from `functoria-runtime` to `fmt.0.9.0`, which is not present in opam-overlays:

Ah ok, may be we probably downgrade `fmt`, I'm not sure that we really requires `fmt.0.9.0` (`fmt.0.8.4` should be enough I think).

mirage-runtime.opam
hannesm: yes -- I can see a test/mirage-runtime directory, but there the test is attached (in the dune file) to the mirage package, not mirage-runtime

hannesm:
maybe fmt 0.8.4 or 0.8.5 (which revised the signature for Fmt.failwith and Fmt.invalid_arg) -- or 0.8.7 if the newer Fmt.str etc. are used (instead of the deprecated Fmt.strf) :)

hannesm:
> to `fmt.0.9.0`, which is not present in opam-overlays

as reported https://github.com/dune-universe/opam-overlays/issues/89

dinosaure:
It seems that `available: opam-version >= "2.1"` does not allow us to use `ocaml-ci` 😕 .

mirage-runtime.opam
```diff
@@ -11,14 +11,14 @@ tags:         ["org:mirage" "org:xapi-project"]
 doc:          "https://mirage.github.io/mirage/"
 
 build: [
-  ["dune" "subst"] {pinned}
+  ["dune" "subst"] {dev}
   ["dune" "build" "-p" name "-j" jobs]
   ["dune" "runtest" "-p" name "-j" jobs] {with-test}
```
hannesm, line 16:
```suggestion
```

there aren't any tests, anyways :)

hannesm:
I opened https://github.com/dinosaure/mirage/pull/1 with further test nitpicks

samoht:
Tests are broken on macOS, I'm fixing those.

dinosaure:
> Tests are broken on macOS, I'm fixing those.

Should I redo the tarball?

mirage.opam
```diff
@@ -29,7 +29,6 @@ depends: [
   "alcotest" {with-test}
   "fmt" {>= "0.8.9" & with-test}
 ]
-available: opam-version >= "2.1.0"
```
dinosaure: Note that it is required on `opam-repository` for two reasons:
1) we only able to use `opam.2.1`
2) the release on `opam-repository` uses the `hidden-flags` which is available only on `opam.2.1`

The real issue is mostly about `ocaml-ci` which fails to solve packages.

samoht: This unblocks ocaml-ci so there is an issue here. I am not sure to understand why this is needed, I'll comment on the main PR.

hannesm: I agree that what we release to opam-repository should include the available constraint, but for this repository - if ocaml-ci is not yet on opam 2.1 - I see more value in allowing the CI to run here than to be pedantic about the availability.

would be great to also open an issue with ocaml-ci to update opam to 2.1.

PR #1276, opened 2022-01-31
Fix minimal version required for mirage{,-runtime}

dinosaure:
According to OPAM and for a beta release, the minimal version
must be 4.0 instead of 4.0.0. Indeed, OPAM concludes that
4.0.0~beta1 < 4.0.0 (but 4.0.0~beta1 > 4.0).

Try to fix mirage/mirage-dev#381


dinosaure:
I'm not sure to understand why `ocaml-ci` fails now 😕 .

kit-ty-kate:
it might be worth doing for the upperbound constraints as well

PR #1283, opened 2022-02-02
Add Git devices

dinosaure:
We need a release of `git` which will be done as soon as I can but it's my initial draft to provide Git devices. Tell me if the documentation is good enough.


lib/mirage/mirage.mli
```diff
@@ -764,6 +764,89 @@ val default_argv : argv impl
...
+  tcpv4v6 impl ->
+  mimic impl ->
+  mimic impl
+(** [git_ssh ?authenticator ~key tcpv4v6 ctx] returns a new {!mimic} context
+    which is able to allocate an SSH connection with the given private [key]
+    with a remote Git repository. You can ensure the identity of the remote Git
+    repository via [authenticator].
+
+    The format of the private key is: [<type>:<base64>] The format of the
+    authenticator is: [<domain>:<hash>:<fingerprint>] *)
```
hannesm, line 831: hmm - is it type:base64? wasn't it type and then either seed or base 64 encoded private key? which types are supported?

for authenticator - why is there a domain? which public key should be used?

e.g. from https://github.com/roburio/unipi/ :
```
--ssh-authenticator to authenticate the ssh remote (format is SHA256:b64-encoded-public-key hash, the output of ssh-keygen -lf <(ssh-keyscan -t rsa|ed25519 remote-host 2>/dev/null))
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,89 @@ val default_argv : argv impl
...
+val git_http :
+  ?tls_key_fingerprint:string key ->
+  ?tls_cert_fingerprint:string key ->
+  ?headers:(string * string) list key ->
+  ?time:time impl ->
+  ?pclock:pclock impl ->
+  tcpv4v6 impl ->
+  mimic impl ->
+  mimic impl
+(** [git_http ?tls_key_fingerprint ?tls_cert_fingerprint ?headers tcpv4v6 ctx]
```
hannesm, line 842: I still find it kind of strange what authenticator is used:
- by default a chain of trust with ca-certs-nss
- if tls_key_fingerprint is provided, this value (unclear which format this should be in - likely hex?)
- if tls_cert_fingerprint is provided, that value

is it possible to provide a custom trust anchor? it should be. what will happen if both tls_key_fingerprint and tls_cert_fingerprint are provided? is there still a possibility to have no validation?

so. similar to the ssh above, I'd be in favour of a single `authenticator` (optional) labeled argument, and its value should include:
- key;hash?:fingerprint
- cert:hash?:fingerprint
- trust-anchor:pem-encoded cert (list?)
- none

dinosaure: Yeah we can improve that but about `trust-anchor:pem-encoded cert`, you mean a file (which must be read) or a serialized form? And if we have `none`, that means Mozilla's NSS trust anchors?

lib/mirage/mirage.mli
dinosaure: Yes, I wrote without a check of format, I need to look about `authenticator` but you are right about the key.

lib/mirage/mirage.mli
hannesm: > you mean a file

no, just the raw pem-encoded certificate (if argument length allows that - at least EC certificates should fit nicely into the limit). a file would be rather complex to support

> `none`, that means Mozilla's NSS trust anchors?

no, it should mean none --> no, thanks, no validation whatsoever should be done. mainly for debugging purposes, or for scenarios where you don't care.

dinosaure: but how you want then to handle Mozilla's NSS trust anchors?

hannesm: as default? also an explicit `nss-trust-anchors` value for `authenticator` could be supported which uses the nss trust anchors. My main arguments are:
- why have multiple optional keyword arguments, and what is the semantics if all are present?
- there are at least two use cases not supported:
   - a custom trust anchor (self-signed certificate in environments where you have your own CA), and 
   - "no validation of the remote server certificate at all".

samoht:
I am a bit confused by these combinators. Are they used for client-side connection or server-side configuration? Are they are just useful for Git or also to set up different kinds of connections?

As far as I am aware, there is no consensus on mimic replacing conduit yet. If so, could we try to avoid exposing a new concept/name "mimic" here, and just call it a `git_config` device or whatever? I am also a bit confused by the `merge` combinator. Why do we need to expose this to our users?

dinosaure:
> I am a bit confused by these combinators. Are they used for client-side connection or server-side configuration?

We don't have the server implementation yet. So it's only about the client.

>  Are they are just useful for Git or also to set up different kinds of connections?

`mimic` is useful for any connections. I did not yet implement device for `paf` (the `http/af` support) as a client but it will use the same logic.

> As far as I am aware, there is no consensus on mimic replacing conduit yet.

It's not a replacement. `ocaml-git`/`irmin` needs `mimic` to be able to `fetch`/`push`.

> If so, could we try to avoid exposing a new concept/name "mimic" here, and just call it a git_config device or whatever?

The concept of `mimic` is not directly related to what Git does. Again, the usage of `mimic` is relevant for `paf` as a client too.

If I decided to propose these devices, it's mostly because @hannesm and me still continue to copy/paste the same bunch of code into our `config.ml` where we want to use a remote Git store (for instance, our SMTP stack, `dns-primary-git` and some other unikernels). I'm pretty confident that these devices should not change a lot - `mimic` has a stable API and expected features are implemented. About composition (and specially `merge`), the user should be able to merge the SSH client `git_ssh` with, for instance, `git_http` if he/she wants to communicate _via_ these protocols.

samoht:
I understand that these combinators are useful and I think we should definitely integrate those in the mirage tool. I just find the exposed API confusing and I was just wondering if there is a way to simplify the surfaced API.

dinosaure:
> (ie. it's confusing that you are calling them Git combinators but their types are named differently)

I agree that names and/or documentations are probably not the best (it's why this PR is a draft), feel free to propose a better name and/or documentation :) but keep in your mind that `mimic` can be used for something else than Git.

samoht:
> but keep in your mind that mimic can be used for something else than Git.

do you plan to add combinators for this something else too? If yes, that'd be useful to see what they look like before settling on an API.

dinosaure:
> do you plan to add combinators for this something else too? If yes, that'd be useful to see what they look like before settling on an API.

Yes, for `paf` and it will probably be exactly the same as `git_http` & `git_happy_eyeballs` - it will make a `ctx` which can be pass to `Client.{get,post,put,...} ~ctx`.

dinosaure:
Ok this PR is ready to be merged, if you are able to take a look on the documentation, it will be super useful.

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,104 @@ val default_argv : argv impl
 val no_argv : argv impl
 (** [no_argv] Disable command line parsing and set argv to [|""|]. *)
 
+(** {2 Mimic device} *)
+
+(** A {i mimic} device is a {i context} which contains few informations to let
```
yomimono, line 769: rewording suggestion: "A {i mimic} device is a {i context} allowing the user to [connect]/{!Mimic.resolve}/allocate a new resource to communicate with a peer" . If this is too drastic a change, "few informations" is strange and I'd at least change that to "information".

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,104 @@ val default_argv : argv impl
...
+    the user to [connect]/{!Mimic.resolve}/allocate a new resource to
+    communicate with a peer. Depending on what the context contains, the
+    protocol used to communicate can be TCP/IP, TLS or SSH (or something else).
```
yomimono, line 772: It has to be network-based though? It can't be, for example, Vchan?

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,104 @@ val default_argv : argv impl
...
+
+    Few packages deliver a {i context} which allows the user to communicate to a
```
yomimono, line 774: I'd say "Some" rather than "Few" unless you mean to emphasize that not many do this.

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,104 @@ val default_argv : argv impl
...
+val mimic : mimic typ
+(** A simple device which define a {i certain} Mimic's context. *)
+
+val mimic_merge : mimic impl -> mimic impl -> mimic impl
+(** [mimic_merge ctx0 ctx1] merges given contexts into one. *)
+
+(** {2 Git configuration} *)
+
+(** The user is able to initiate a connection with a Git repository. However, it
+    exists multiple ways to communicate with a Git repository:
```
yomimono, line 791: "... there are many ways to communicate..." reads a bit better

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,104 @@ val default_argv : argv impl
...
+
+    - TCP/IP
+    - HTTP
+    - HTTP + TLS
+    - SSH
+
+    These devices define these composable ways and deliver a Mimic's context
+    which can be pass to Git or Irmin in order to be able to {i fetch} and
+    {i push} from/into a Git repository. *)
```
yomimono, line 800: Not sure whether this is clearer, but here's another run at this section: "The devices defined below define these composable ways. The [mimic impl] returned from them can be passed to Git or Irmin in order to be able to {i fetch} and {i push} from/into a Git repository."

If that's not better, "... which can be pass ..." should be "which can be passed".

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,104 @@ val default_argv : argv impl
...
+val git_happy_eyeballs :
+  ?random:random impl ->
+  ?time:time impl ->
+  ?mclock:mclock impl ->
+  ?pclock:pclock impl ->
+  stackv4v6 impl ->
+  mimic impl
+(** [git_happy_eyeballs stackv4v6] provides a {!mimic} device which is {b only}
+    able to resolve domain-name. The user must {!mimic_merge} this device with
+    another which provides a context with a {i certain} protocol. *)
```
yomimono, line 811: I'm not sure I understand what "certain" means here? From the context I'm guessing it's a communication protocol that supports arbitrary messages (e.g. TCP, SSH?)

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,104 @@ val default_argv : argv impl
...
+    - key(:<hash>)?:<b64-encoded fingerprint> to authenticate via the key
+      fingerprint
+    - cert(:<hash>)?:<b64-encoded fingerprint> to authenticate via the cert
+      fingerprint
+    - trust-anchor(:<der-encoded cert>)+ to authenticate via a list of
+      certificates
+    - By default, we use X.509 trust anchors extracted from Mozilla's NSS *)
+
+val git_tcpv4v6_of_stackv4v6 : stackv4v6 impl -> tcpv4v6 impl
+(** [git_tcpv4v6_of_stackv4v6 stackv4v6] is an helpers to extract the TCP/IP
```
yomimono, line 862: s/an helpers/a helper/

yomimono:
Once I got to the documentation about what a mimic is, I thought the description was fairly clear.

samoht:
I'd definitely suggest renaming the `mimic` types and values to `git` as this is describing the signature for Git devices. I am just a bit confused as they all seem to generate a module type with a slightly different signature? Usually a MirageOS device corresponds to a unique module type.

Similarly, I am not sure to understand how does the `mimic_merge` function reflect in term of module types or implementations. Could you give an example of how you are using this in existing unikernels?

lib/mirage/mirage.mli
dinosaure: When I say _certain_ protocol, I mean any protocols. A concrete example is, of course, SSH or TCP/IP or even HTTP but it can be something else.

lib/mirage/mirage.mli
dinosaure: It can be `vchan` 👍 . As far as the protocol provides a `Mirage_flow.S`, it can be used as a _transmission protocol_ for the Smart protocol.

dinosaure:
I force-pushed on this PR to try to respond to @samoht's concerns and @yomimono comments. The `mimic` device is gone and I made 2 devices:
- `git_client_with_dns`
- `git_client`

The first one can be better and _can_ expects few values such as a `Dns_mirage.t` (which defines _nameservers_) but we need to implement these devices (and I think it's orthogonal to this PR). Then, I provided a little example to be clear how to use these devices and how to compose them.

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,114 @@ val default_argv : argv impl
...
+    its application. For instance, the user is able to restrict only the SSH
+    connection to communicate with a Git repository or the user can handle
+    TCP/IP and SSH as possible protocols to communicate with a peer.
+
+    For instance, a device which is able to communicate {i via} TCP/IP and SSH
+    can be implemented like:
+
+    {[
+      let git_client =
+        let dns = git_happy_eyeballs stackv4v6 in
```
samoht, line 791: Should we call this device `dns`? Why does `git_happy_eyeballs` is related to `git` at all?

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,114 @@ val default_argv : argv impl
...
+        let ssh = git_ssh ~key (git_tcpv4v6_of_stackv4v6 stackv4v6) dns in
+        let tcp = git_tcp (git_tcpv4v6_of_stackv4v6 stackv4v6) dns in
+        git_merge_clients ssh tcp
+    ]} *)
+
+type git_client
+(** Type of a device which implements a {i certain} protocol. *)
```
samoht, line 798: `certain` is not clear here: I guess you mean "Git"?

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,114 @@ val default_argv : argv impl
...
+  git_client_with_dns impl
+(** [git_happy_eyeballs stackv4v6] provides a {!git_client_with_dns} device *
+    which is {b only} able to resolve domain-name. Then, the user is able to
+    create a new device {!git_client} from it which will provide a {i certain}
+    protocol. The resolution of domain-name is done {i via} a {i happy-eyeballs}
+    query. *)
+
+val git_tcp : tcpv4v6 impl -> git_client_with_dns impl -> git_client impl
+(** [git_tcp tcpv4v6 ctx] returns a new {!mimic} context which is able to
+    allocate a TCP/IP connection with a remote Git repository. *)
```
samoht, line 825:
```suggestion
(** [git_tcp dns ctx] is a device able to connect to a remote Git repository using TCP/IP. *)
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,114 @@ val default_argv : argv impl
...
+  key:string option key ->
+  ?mclock:mclock impl ->
+  ?time:time impl ->
+  tcpv4v6 impl ->
+  git_client_with_dns impl ->
+  git_client impl
+(** [git_ssh ?authenticator ~key tcpv4v6 ctx] returns a new {!mimic} context
+    which is able to allocate an SSH connection with the given private [key]
+    with a remote Git repository. You can ensure the identity of the remote Git
+    repository via [authenticator].
```
samoht, line 838:
```suggestion
(** [git_ssh ?authenticator ~key tcpv4v6 dns] is device able to connect to a
   remote Git repository using an SSH connection with the given private [key].
    You can ensure the identity of the remote Git repository via [authenticator].
```

samoht: What should the string provided to `authenticator` contain? Is it the name of a directory in a KV-RO store?

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,114 @@ val default_argv : argv impl
...
+  ?time:time impl ->
+  ?pclock:pclock impl ->
+  tcpv4v6 impl ->
+  git_client_with_dns impl ->
+  git_client impl
+(** [git_http ?authenticator ?headers tcpv4v6 ctx] returns a new {!mimic}
+    context which is able to allocate an HTTP(S) connection with a remote Git
+    repository. The user is able to append few {i meta-data} {i via} the
+    [headers]. The user is able to verify the identity of the remote Git
+    repository via an [authenticator]. The format of it is:
```
samoht, line 862:
```suggestion
(** [git_http ?authenticator ?headers tcpv4v6 dns] is a device able to connect
      to a remote Git repository via an HTTP(S) connection, using the provided
      HTTP  [headers]. The identity of the remote Git can be verified using [authenticator]. 
      
      The format of it is:
```

lib/mirage/mirage.mli
dinosaure: The best will be: the Git protocol over a _certain_ transmission protocol. WDYT?

lib/mirage/mirage.mli
hannesm: from a brief look into "Git_mirage_happy_eyeballs", this seems to depend and use "mimic". I think the code should be part of mimic. `happy-eyeballs{-mirage}` is pretty concise about its footprint, and useful in other settings than mimic. The scope and API is basically https://datatracker.ietf.org/doc/html/rfc8305.

samoht: ok, so would renaming `type git_client_with_dns` into `type happy_eyeball` make sense? We could move the code later on (I'm just focusing on the API surfaced in mirage.mli here)

lib/mirage/mirage.mli
samoht: I'd go with a simpler form as `certain` doesn't mean the same thing in French and English so it's confusing here.

```suggestion
(** The type for devices that implement the Git protocol. *)
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,113 @@ val default_argv : argv impl
...
+        git_merge_clients ssh tcp
+    ]} *)
+
+type git_client
+(** Type of a device which implements a {i certain} protocol. *)
+
+val git_client : git_client typ
+
+type git_client_with_dns
+(** Type of a device which is able to {b only} resolve domain-name. *)
```
samoht, line 803:
```suggestion
(** The type for devices that resolve domain-name using the happy-eyeballs protocol (RFC 8305). *)
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,112 @@ val default_argv : argv impl
...
+        let tcp = git_tcp (git_tcpv4v6_of_stackv4v6 stackv4v6) dns in
+        git_merge_clients ssh tcp
+    ]} *)
+
+type git_client
+(** The type for devices that implement the Git protocol. *)
+
+val git_client : git_client typ
+
+type git_client_with_dns
```
samoht, line 801:
```suggestion
type happy_eyeballs
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,112 @@ val default_argv : argv impl
...
+val git_merge_clients : git_client impl -> git_client impl -> git_client impl
+(** [git_merge_clients a0 a1] merges given [git_client] into one. *)
+
+val git_happy_eyeballs :
+  ?random:random impl ->
+  ?time:time impl ->
+  ?mclock:mclock impl ->
+  ?pclock:pclock impl ->
+  stackv4v6 impl ->
+  git_client_with_dns impl
```
samoht, line 815:
```suggestion
  dns impl
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,112 @@ val default_argv : argv impl
...
+    [ssh-keygen -lf <(ssh-keyscan -t rsa|ed25519 remote
+    2>/dev/null)]. *)
+
+val git_http :
+  ?authenticator:string option key ->
+  ?headers:(string * string) list key ->
+  ?time:time impl ->
+  ?pclock:pclock impl ->
+  tcpv4v6 impl ->
+  git_client_with_dns impl ->
```
samoht, line 853:
```suggestion
  dns impl ->
```

lib/mirage/mirage.mli
samoht: you don't have to do any changes in `mimic` yet, let's try to have a nice API in mirage.mli first (and merge this PR). I think we are getting closer now :-)

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,115 @@ val default_argv : argv impl
...
+
+val git_client : git_client typ
+
+type dns
+(** The type for devices that resolve domain-name using the happy-eyeballs
+    protocol (RFC 8305). *)
+
+val happy_eyeballs : dns typ
+
+val git_happy_eyeballs :
```
samoht, line 807:
```suggestion
val happy_eyeballs :
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,115 @@ val default_argv : argv impl
...
+  ?time:time impl ->
+  ?mclock:mclock impl ->
+  ?pclock:pclock impl ->
+  stackv4v6 impl ->
+  dns impl
+(** [git_happy_eyeballs stackv4v6] provides a {!git_client_with_dns} device *
+    which is {b only} able to resolve domain-name. Then, the user is able to
+    create a new device {!git_client} from it which will provide a {i certain}
+    protocol. The resolution of domain-name is done {i via} a {i happy-eyeballs}
+    query. *)
```
samoht, line 818:
```suggestion
(** [happy_eyeballs stackv4v6] is a {!dns} device that resolve domain-name using the happy-eyeballs
    protocol (RFC 8305). *)
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,115 @@ val default_argv : argv impl
...
+    ]} *)
+
+type git_client
+(** The type for devices that implement the Git protocol. *)
+
+val git_client : git_client typ
+
+type dns
+(** The type for devices that resolve domain-name using the happy-eyeballs
+    protocol (RFC 8305). *)
```
samoht, line 803:
```suggestion
(** The type for devices that resolve domain-names. *)
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,115 @@ val default_argv : argv impl
...
+  ?pclock:pclock impl ->
+  stackv4v6 impl ->
+  dns impl
+(** [git_happy_eyeballs stackv4v6] provides a {!git_client_with_dns} device *
+    which is {b only} able to resolve domain-name. Then, the user is able to
+    create a new device {!git_client} from it which will provide a {i certain}
+    protocol. The resolution of domain-name is done {i via} a {i happy-eyeballs}
+    query. *)
+
+val git_merge_clients : git_client impl -> git_client impl -> git_client impl
```
samoht, line 820:
```suggestion
val merge_git_clients : git_client impl -> git_client impl -> git_client impl
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,115 @@ val default_argv : argv impl
...
+(** [git_merge_clients a0 a1] merges given [git_client] into one. *)
```
samoht, line 821:
```suggestion
(** [merge_git_clients a b] is a device that can connect to remote Git repositories using either the device [a] or the device [b]. *)
```

samoht:
I believe that my last suggested edits to `mirage.mli` are the last one (phew). Could you squash and maybe have a last pair of eyes look at this to make sure the resulted docstring make sense (maybe @yomimono ?)

dinosaure:
I will double check that `mirage-skeleton/applications/git` works on my side.

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,112 @@ val default_argv : argv impl
...
+(** {2 Git client configuration} *)
+
+(** Users can connect to a remote Git repository in many ways:
+
+    - TCP/IP
+    - HTTP
+    - HTTP + TLS
+    - SSH
+
+    The devices defined below define these composable ways. The
```
samoht, line 776:
```suggestion
    The devices defined below define these in composable ways. The
```

lib/mirage/mirage.mli
```diff
@@ -764,6 +764,112 @@ val default_argv : argv impl
...
+    its application. For instance, the user is able to restrict only the SSH
+    connection to communicate with a Git repository or the user can handle
+    TCP/IP and SSH as possible protocols to communicate with a peer.
+
+    For instance, a device which is able to communicate {i via} TCP/IP and SSH
+    can be implemented like:
+
+    {[
+      let git_client =
+        let dns = git_happy_eyeballs stackv4v6 in
```
samoht, line 790:
```suggestion
        let dns = happy_eyeballs stackv4v6 in
```

dinosaure:
I can confirm that the current status of this PR work with `mirage-skeleton/applications/git`.

yomimono:
I think it looks great :)

hannesm:
I've to admit, similar to `Mirage_impl_resolver` I'm very confused if you choose names for types and values that are then actually much more than the basic implementation:
- `type dns` to me should be the basic Dns_client_mirage.Make().t -- not something wrapped that is embedded in git or mimic
- happy_eyeballs the same -- there is a Happy_eyeballs_mirage.Make().t (constructed by `create`) that is pretty useful. Now, the happy_eyeballs in mirage is something deep inside git depending on mimic -- to me at least this is not straightforward and easy to follow what to do and what to use.
- The same was and is the case for `conduit` wrapping some `resolver` - which is mostly the Dns_client_mirage again, but some boilerplate around it.

To me, these devices are rather cumbersome and incomplete: what if you use conduit _and_ git, can't you use the same dns_client? How to provide the recursive resolver (the Mirage_impl_resolver has a nameserver argument exposed as `--resolver` IIRC)?

dinosaure:
> * type dns to me should be the basic Dns_client_mirage.Make().t -- not something wrapped that is embedded in git or mimic

I agree with you. I prefer to prefix this type with `mimic` to ensure that such type is in purpose of `mimic` (with `git` and `paf`) than a DNS thing. Moreover, `git_happy_eyeballs`/`happy_eyeballs` can take a `Dns_client_mirage.Make().t` into its `connect` - I did not go further on this side because `Dns_client_mirage.Make` does not provides (yet?) a signature.

> * happy_eyeballs the same -- there is a Happy_eyeballs_mirage.Make().t (constructed by create) that is pretty useful. Now, the happy_eyeballs in mirage is something deep inside git depending on mimic -- to me at least this is not straightforward and easy to follow what to do and what to use.

Same as above, if we can prepend the name with `mimic`. According to @samoht's comments however, it becomes inconsistent with Git where, to be able to allocate a resource to communicate with a Git repository, we must play with `mimic_*` things and `git_*` things if we prefix `dns` & `happy_eyeballs`. Again, these resources should:
1) be available from the `mimic` distribution (it's a non-sense that the `ocaml-git` distribution provide a Mimic context which is able to only resolve domain-name
2) be usable by another higher protocol than Git (for instance, HTTP but we can talk about SMTP too)

In my head, the graph is:
```
mimic: empty context
 `- context with: domain-name -> Mimic.flow (TCP/IP) (happy-eyeballs)
      |- context with: domain-name -> Mimic.flow (TCP/IP) -> Git protocol (git+tcp/ip)
      |- context with: domain-name -> Mimic.flow (TCP/IP) -> SSH -> Git protocol (git+ssh)
      |- context with: domain-name -> Mimic.flow (TCP/IP) -> HTTP -> Git protocol (git+http)
      |- context with: domain-name -> Mimic.flow (TCP/IP) -> TLS -> HTTP -> Git protocol (git+https)
      |- context with: domain-name -> Mimic.flow (TCP/IP) -> HTTP (http)
      `- context with: domain-name -> Mimic.flow (TCP/IP) -> TLS -> HTTP (https)
```

And all of them are interchangeable/mergeable (the `priority` mechanism in `mimic` ensure then to choose, into the `Mimic.resolve`, the best one). Note we talk about `Mimic.flow (TCP/IP)`, but, as I said, we can replace it by `Mimic.flow (Vchan)` or `Mimic.flow (SOCKS)` if we want (as long as we can talk to someone with such protocol). The design is a bit strange but for me, it's a reverse about what we can do with `functoria`.

Indeed, `functoria` say that from some arguments, we build implementations. `Mimic` says that from some meta-data, we build some values which are required for some implementations - if these values exists into the context and/or are resolvable from the context, we are able to choose the right implementation.

My best wish will be to have a `Dns_mirage_client.S` and have a `Happy_eyeballs_mirage.S`, be sure that you can build the second with the first one and pass them into a `Mimic`'s device to produce a `context` which can be used by `Git.fetch`/`Git.push`/`Http.get`/`Http.post`/... regardless implementations used and selected by `functoria`.

> The same was and is the case for conduit wrapping some resolver - which is mostly the Dns_client_mirage again, but some boilerplate around it.

I don't have opinion about this.

> To me, these devices are rather cumbersome and incomplete: what if you use conduit and git, can't you use the same dns_client?

For me, it's incompatible. `git` does not use `conduit` anymore and expects explicitly a `Mimic.ctx`. The case can appear if you mix an HTTP client and a Git client. But in such case, the best will be to use `git` and `paf` who share the same design - and can use only one and unique instantiated `happy_eyeballs` value `mimic_happy_eyeballs`.

> How to provide the recursive resolver (the Mirage_impl_resolver has a nameserver argument exposed as --resolver IIRC)?

The question is more related to: do we provide a `Dns_mirage_client.S` which can be concretized to a `functoria` type `dns` and built from several arguments such as the `nameserver`, `timeout`, etc? I really would like to have this kind of devices regardless Git or Mimic and re-use them then. For me, the `Mirage_impl_resolver.resolver` is a `conduit` relative device. Again, I can agree with you that `ocaml-dns` should provide its own device regardless any stuffs around such as `conduit` and `mimic`.

samoht:
Merging as I think we have converged to something which, while not being perfect, is a good improvement. Let's make further improvements (if ant) in upcoming PRs!

samoht:
Done in https://github.com/mirage/mirage/pull/1290

PR #1286, opened 2022-02-04
Preview: upgrade to ocamlformat.0.21.0

gpetiot:
This is a preview of the formatting of your project codebase with the upcoming `ocamlformat.0.21.0`. Note that this package is not yet released in opam (so the linting/formatting task of your CI won't pass), the formatting might also change slightly in the released version as we try to integrate feedback from our users.

The changes on this project are due to:
- the separators of codeblocks `{[]}` in doc-comments are now placed on their own line, this is the new syntax agreed upon with other tools handling odoc comments.

As always, we are happy to hear your feedback!  Thank you for using ocamlformat!

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version = 0.20.1
+version = 0.21.0
 profile = conventional
 break-infix = fit-or-vertical
 parse-docstrings = true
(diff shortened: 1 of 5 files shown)


dinosaure:
Thanks, seems fine for me.

PR #1288, opened 2022-02-08
Add `docteur` device

dinosaure:
`docteur` is close to be released. May be a good time to add it into our `mirage-skeleton`. This PR is a draft to have review about documentation. Currently, the workflow is simple:
- the user decide what he/she wants to download (it can be a remote Git repository or `file://path/`)
- the device (as `tar_archive`) makes the image _via_ dune
- we install the image (`disk` by default - Solo5 does not like `.`) into the `dist/` directory
- from `unix`, a simple launch is enough (the block device will `open "disk"` by itself)
- from `solo5`, we must pass `--block:disk=<path-to-the-disk-file>`

### DIFF
--- a/lib/mirage/mirage.mli
+++ b/lib/mirage/mirage.mli
@@ -294,6 +294,28 @@ val generic_kv_ro :
 
     If no key is provided, it uses {!Key.kv_ro} to create a new one. *)
 
+val docteur :
+  ?mode:[ `Fast | `Light ] ->
+  ?disk:string Key.key ->
+  ?analyze:bool Key.key ->
+  string ->
+  kv_ro impl
+(** [docteur ?mode ?disk ?analyze remote] creates a Docteur image which can be
+    used by the produced {i unikernel} as an external block-device. It permits
+    to attach a read-only file-system. [analyze] checks the integrity of the
+    given block-device at the boot-time (it can take a time). It's possible to
+    use the file-system into 2 modes:
+
+    - [`Light] an access requires that we reconstruct the path to the requested
+      object. That mostly means that we probably need to extract few objects
+      before the extraction of the requested one. This choice does not require a
+      huge usage of memory but it can be slower is deep in your filesystem
+    - [`Fast] reconstructs the layout of your file-system at the boot time (so
+      it can take a time). Then, if the user wants a specific file, we ensure
+      that only the requested object is extracted. However, depending on your
+      image, this mode can require a huge amount of memory to keep the structure
+      of your file-system. *)
+
 type kv_rw
 (** Abstract type for read-write key/value store. *)
 
--- a/lib/mirage/mirage.mli
+++ b/lib/mirage/mirage.mli
@@ -300,12 +300,11 @@ val docteur :
   ?analyze:bool Key.key ->
   string ->
   kv_ro impl
-(** [docteur ?mode ?disk ?analyze remote] creates a {i docteur} image which can
-    be used by the produced {i unikernel} as an external block-device. It
-    permits to attach a read-only (probably huge) file-system. The format of the
-    file-system is the Git PACK file (version 2). It permits to store a huge
-    amount of files into a small image (for instance, 14Gb of documentation can
-    be compiled into a {i docteur} image of 240Mb).
+(** [docteur ?mode ?disk ?analyze remote] is a read-only, key-value store
+    device. Data is stored on that device using the Git PACK file format,
+    version 2. This format has very good compression factors for many similar
+    files of relatively small size. For instance, 14Gb of HTML files can be
+    compressed into a disk image of 240Mb.
 
     Unlike {!crunch}, [docteur] produces an external image which means that less
     memory is used to keep and get files. The image can be produced from many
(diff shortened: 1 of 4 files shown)


lib/mirage/mirage.mli
```diff
@@ -294,6 +294,28 @@ val generic_kv_ro :
 
     If no key is provided, it uses {!Key.kv_ro} to create a new one. *)
 
+val docteur :
+  ?mode:[ `Fast | `Light ] ->
+  ?disk:string Key.key ->
+  ?analyze:bool Key.key ->
+  string ->
+  kv_ro impl
+(** [docteur ?mode ?disk ?analyze remote] creates a Docteur image which can be
```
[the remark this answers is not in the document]
dinosaure: It's a Git PACK file, should we define it as is?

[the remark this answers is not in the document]
dinosaure: I improved the documentation to explain (in the opposite of `crunch`) why `docteur` can be interesting.

dinosaure:
This PR is ready to be merged. I improved the documentation and, if you want, I can add a new unikernel into `mirage-skeleton` as an example such device (like `ocaml-crunch`).

lib/mirage/mirage.mli
```diff
@@ -294,6 +294,54 @@ val generic_kv_ro :
...
+  ?disk:string Key.key ->
+  ?analyze:bool Key.key ->
+  string ->
+  kv_ro impl
+(** [docteur ?mode ?disk ?analyze remote] creates a {i docteur} image which can
+    be used by the produced {i unikernel} as an external block-device. It
+    permits to attach a read-only (probably huge) file-system. The format of the
+    file-system is the Git PACK file (version 2). It permits to store a huge
+    amount of files into a small image (for instance, 14Gb of documentation can
+    be compiled into a {i docteur} image of 240Mb).
```
samoht, line 308:
```suggestion
(** [docteur ?mode ?disk ?analyze remote] is a read-only, key-value device.
    Data are stored on that device using the Git PACK file format, version 2.
    This format has very good compression factors for many similar files of relatively
    small size. For instance, 14Gb of HTML files can be compressed into a disk 
    image of 240Mb.
```

lib/mirage/mirage.mli
```diff
@@ -294,6 +294,54 @@ val generic_kv_ro :
...
+
+    Unlike {!crunch}, [docteur] produces an external image which means that less
+    memory is used to keep and get files. The image can be produced from many
+    sources:
+
+    - A local Git repository (like [file://path/to/the/git/repository/])
+    - A simple directory (like [file://path/to/a/simple/directory/])
+    - A remove Git repository (via SSH, HTTP(S) or TCP/IP as what [git clone]
```
samoht, line 316:
```suggestion
    - A remote Git repository (via SSH, HTTP(S) or TCP/IP as what [git clone]
```

lib/mirage/mirage.mli
```diff
@@ -294,6 +294,54 @@ val generic_kv_ro :
...
+    For a Solo5 target, users must {i attach} the image as a block device:
+
+    {[ $ solo5-hvt --block:<name>=<path-to-the-image> -- unikernel.{hvt,...} ]}
+
+    For the Unix target, the program [open] the image at the beginning of the
+    process.
+
+    [docteur] becomes useful to load some files without the memory restriction
+    when the unikernel is launched (~ 1Go) with Solo5.
+
```
samoht, line 328:
```suggestion

```

lib/mirage/mirage.mli
```diff
@@ -294,6 +294,54 @@ val generic_kv_ro :
...
+    - [`Light]: an access requires that we reconstruct the path to the requested
+      file. That mostly means that we probably need to extract few objects
+      before the extraction of the requested one. This choice does not require a
+      huge usage of memory but it can be slower if the requested file is deep in
+      the structure.
+    - [`Fast]: reconstructs the layout of your file-system at the boot time (so
+      it can take a time at the beginning). Then, if the user wants a specific
+      file, we ensure that only the requested object is extracted. However,
+      depending on your image, this mode can require a huge amount of memory.
+      Indeed, it keeps the structure of your file-system in memory. *)
```
samoht, line 343:
```suggestion
    - [`Light]: any access requires that we reconstruct the path to the requested
      file. That means that we will need to extract a few additional objects
      before the extraction of the requested one. [`Light] does not cache
      anything in memory but it can be slower if the requested file is deep in
      the directory structure.
    - [`Fast]: reconstructs and cache the layout of the directory structure 
      when the unikernel starts: it might increase boot-time and bigger 
      memory requirements. However, [`Fast] allows the device to decode
      only the requested object so it is faster than the [`Light] mode. *)
```

samoht:
Could you fix the formatting? Happy to merge after this.

dinosaure:
Done 👍

lib/mirage/mirage.mli
```diff
@@ -294,6 +294,47 @@ val generic_kv_ro :
...
     If no key is provided, it uses {!Key.kv_ro} to create a new one. *)
 
+val docteur :
+  ?mode:[ `Fast | `Light ] ->
+  ?disk:string Key.key ->
+  ?analyze:bool Key.key ->
+  string ->
+  kv_ro impl
+(** [docteur ?mode ?disk ?analyze remote] is a read-only, key-value device. Data
+    are stored on that device using the Git PACK file format, version 2. This
```
samoht, line 304:
```suggestion
(** [docteur ?mode ?disk ?analyze remote] is a read-only, key-value store device.
    Data is stored on that device using the Git PACK file format, version 2. This
```

samoht:
(sorry I broke formatting again... this time it should be the last time..)

dinosaure:
I force-pushed to clean the history. It should be good now.

PR #1289, opened 2022-02-09
Update to cmdliner 1.1.1

samoht:
/cc @dbuenzli  I'm not sure if I've done the "right thing" so your feedback would be great! 

`Cmd.eval_value` seems to be a bit more complex than the previous `eval` but that's manageable. The main change is that peek/eval do not work on the same types anymore (as it's not possible to extract a `'a Term.t` from `'a Cmd` due to grouping). The rest is pretty straitghforward.

Something very minor is that the man pages seem to generate a bit more whitespace than before. Not sure what changed (maybe just a side-effect of changes in `Fmt`)?

PR #1291, opened 2022-02-10
Add git devices

dinosaure:
Re-open the PR to continue the discussion from #1283 and try to finalize it. For my point of view:
- [x] we should move `Git_mirage_happy_eyeballs` into the `mimic` distribution
- [x] this module should compose with an `Happy_eyeballs_mirage.S` which should compose with a `Dns_client_mirage.S`
 * [x] so `ocaml-dns` should provides `Dns_client_mirage.S` and a type `dns_client`
 * [x] and we `happy-eyeballs` should provides a `Happy_eyeballs_mirage.S` and a type `happy_eyeballs`
- [x] finally, `mimic` should take the final `happy_eyeballs` `impl` to provide a `mimic_dns`
- [x] and this `mimic_dns` will be used by `git_client` and `paf_client`

PR #1292, opened 2022-02-10
Add Key.opt_all to allows usage of an argument multiple times

dinosaure:
It's a draft to try to implement the `opt_all` combinator. I know that it's probably in conflict with #1289 but it's may be nice to integrate it before and do the upgrade then, The idea behind is to let the user to use multiple times an argument such as:
```sh
$ ./unikernel.exe --dns-nameserver 1.1.1.1 --dns-nameserver 8.8.8.8
```
It's probably not the best example but I'm sure that we can easily imagine many use-cases.


dinosaure:
I added a test about `opt_all` and it seems to work on the `functoria` level, please take a look on that, it becomes needed for #1291

Drup (approved):
This looks good! Maybe add one test for emptyness and that'll be good to go. :)

dinosaure:
Thanks for the review.

PR #1293, opened 2022-02-12
Add tcpv4v6_of_stackv4v6

dinosaure:
Related to mirage/mirage-tcpip#474. It helps us to start to implement protocols which expects only a TCP/IP stack. By this way, it easy to:
1) implement protocols regardless the UDP/IP stack
2) have an easy switch between `mirage-tcpip` which provides a full-stack and `utcp` which provides only a `Tcpip.Tcp.S`

### DIFF
--- a/lib/mirage/mirage.ml
+++ b/lib/mirage/mirage.ml
@@ -197,6 +197,19 @@ let static_ipv4v6_stack = Mirage_impl_stack.static_ipv4v6_stack
 let direct_stackv4v6 = Mirage_impl_stack.direct_stackv4v6
 let socket_stackv4v6 = Mirage_impl_stack.socket_stackv4v6
 
+let tcpv4v6_of_stackv4v6 v =
+  let impl =
+    let packages =
+      [ package "tcpip" ~sublibs:[ "stack-direct" ] ~min:"7.1.0" ]
+    in
+    let connect _ modname = function
+      | [ stackv4v6 ] -> Fmt.str {ocaml|%s.connect %s|ocaml} modname stackv4v6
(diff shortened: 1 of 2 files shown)


hannesm:
Thanks for your PR, I'm not sure I understand the purpose of this "device", taking into account that `Tcpip.Stack.V4V6` signature provides a function `val tcp : t -> TCP.t`?

dinosaure:
> Thanks for your PR, I'm not sure I understand the purpose of this "device", taking into account that Tcpip.Stack.V4V6 signature provides a function val tcp : t -> TCP.t?

Yeah, it's not very clear but currently for an implementation which expects a `Tcpip.Tcp.S`, the only way to create it from an unikernel is:
```ocaml
module Make (Stack : Tcpip.Stack.V4V6) = struct
  module My_super_module = My_super_functor.Make (Stack.TCP)

  let start stackv4v6 =
    let t = My_super_module.connect (Stack.tcp stackv4v6) in
    My_super_module.run t
end
```

The goal of this PR is to let `functoria` to handle such case and be able to write this:
```ocaml
module Make (My_super_module : My_super_signature) = struct
  let start t =
    My_super_module.run t
end
```

Then, `functoria` will care about `let __t = My_super_module.connet __tcpv4v6` and `let __tcpv4v6 = My_stack.tcp __stackv4v6` automatically. The final point is about `utcp` which provides `Tcpip.Tcp.S` and `mirage-tcpip` which provides `Tcpip.Stack.S`. the change from one to the other should easier when it operates only on the `functoria` level.

samoht (approved):

PR #1296, opened 2022-02-23
Upgrade tests with dune.3.0.0

dinosaure:


functoria.opam
```diff
@@ -23,6 +23,7 @@ build: [
 depends: [
   "ocaml" {>= "4.08.0"}
   "dune" {>= "2.8.0"}
+  "dune" {with-test & >= "3.0.0"}
```
samoht, line 26: could you put this on one line?

PR #1300, opened 2022-03-06
Add chamelon device

dinosaure:
/cc @yomimono, it's a draft, feel free to complete/change/update the signature/documentation.

### DIFF
--- a/lib/mirage/impl/mirage_impl_block.ml
+++ b/lib/mirage/impl/mirage_impl_block.ml
@@ -1,6 +1,7 @@
 open Functoria
 open Mirage_impl_misc
 open Mirage_impl_kv
+open Mirage_impl_pclock
 module Key = Mirage_key
 
 type block = BLOCK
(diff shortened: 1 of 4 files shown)


yomimono:
TL;DR - sure, go ahead!

I talked a bit about why I hadn't proposed this myself with @talex5 -- I don't want to contribute to low-quality default solutions in `mirage` that newcomers use and then stumble over.  I'd be more comfortable including it knowing that others using MirageOS 4 before release have some intention to use it, so it's been through testing by more people than me before the MirageOS 4 release.

I had a quite-long set of thoughts here, but then I revisited https://github.com/mirage/mirage-www/pull/562 , which contains a lot of discussion around this exact subject.  Here's my evaluation of `chamelon` against the levels described there. Only shortcomings are highlighted; other subrequirements are :heavy_check_mark: to the best of my knowledge.

:heavy_check_mark: `ready`
:x: `integrated` - the project uses global mutable state in the block allocator (this is guarded by an `Lwt_mutex`). I haven't tried generating documentation with `odig`.
:x: `automated` - there are a lot of tests but there's not automatic CI on push anywhere that users can see
:x: `approved` - the integer overflow protection is definitely not 100%; a hacky attempt at it has broken 32-bit support, which is pretty embarrassing for an explicitly 32-bit format.  I don't know whether the builds are reproducible.
:x: `verified` - Nope. I took a stab at a small subset in an early version and hope to revive this someday.

This is pretty comforting to me, actually! I mean, it says "ready" right there :smile_cat: So go for it!

dinosaure:
> ❌ automated - there are a lot of tests but there's not automatic CI on push anywhere that users can see

Actually, I'm sure that @talex5 can add you as an allowed user to use `ocaml-ci` 🙂 .

yomimono:
As a companion to including `chamelon` here, should we add something exercising it to `mirage-skeleton`? I was going to propose [url-shortener](https://github.com/yomimono/url-shortener) but we could probably just use one of the existing unikernels that operates on KV, and have it use `chamelon`.

dinosaure:
> we could probably just use one of the existing unikernels that operates on KV, and have it use chamelon.

Yes, `url-shortener` is a final application which will be improve (like, switch to `paf` is into my TODO list 🙂). A simple example of `chamelon` where we use `read` and `write` will be the best but I will try to propose that as soon as I can.

dinosaure:
Let's merge and cut a release. Thanks!

PR #1302, opened 2022-03-12
Add dns_client device

dinosaure:
It's a draft due to the fact that `dns.6.2.0` requires `cmdliner.1.1.0` (so we must wait #1289). However, we can start to talk about the interface & the documentation and improve it. Currently. the `dns_client` device can be used as:

**config.ml**:
```ocaml
open Mirage

let domain_name =
  let doc = Key.Arg.info ~doc:"The domain-name to resolve." [ "domain-name" ] in
  Key.(create "domain-name" Arg.(required string doc))

let nameservers =
  let doc = Key.Arg.info ~doc:"Nameserver." [ "nameserver" ] in
  Key.(create "nameserver" Arg.(opt_all string doc))

let unikernel = foreign "Unikernel.Make"
  ~keys:[ Key.v domain_name; Key.v nameservers ]
  (console @-> dns_client @-> job)

let stackv4v6 = generic_stackv4v6 default_network

let () = register "resolve" [ unikernel $ default_console $ generic_dns_client ~nameservers stackv4v6 ]
```

**unikernel.ml**:
```ocaml
module Make (Console : Mirage_console.S) (DNS : Dns_client_mirage.S) = struct
  let log console fmt =
    Fmt.kstr (Console.log console) fmt

  let start console dns =
    let ( >>= ) = Result.bind in
    match Domain_name.(of_string (Key_gen.domain_name ()) >>= host) with
    | Error (`Msg err) -> failwith err
    | Ok domain_name ->
      let open Lwt.Infix in
      DNS.gethostbyname dns domain_name >>= function
      | Ok ipv4 ->
        log console "%a: %a" Domain_name.pp domain_name Ipaddr.V4.pp ipv4
      | Error (`Msg err) -> failwith err
end
```

The unikernel can be used as: `./resolve --nameserver tcp:8.8.8.8 --domain-name google.com`

/cc @hannesm


hannesm:
> `dns.6.2.0` requires `cmdliner.1.1.0`

Actually, only `dns-cli` (a separate opam package) requires cmdliner at all. `dns` does not have such a dependency.

lib/mirage/impl/mirage_impl_dns.ml
```diff
@@ -0,0 +1,32 @@
...
+open Mirage_impl_stack
+open Mirage_impl_random
+
+type dns_client = Dns_client
+
+let dns_client = Type.v Dns_client
+
+let generic_dns_client ?(timeout = 5000000000L) nameservers =
+  let packages =
+    [ package "dns-client" ~sublibs:[ "mirage" ] ~min:"6.2.0" ~max:"6.3.0" ]
```
hannesm, line 14:
```suggestion
    [ package "dns-client" ~sublibs:[ "mirage" ] ~min:"6.2.0" ~max:"7.0.0" ]
```

Breaking API changes will increase the major version.

lib/mirage/impl/mirage_impl_dns.ml
```diff
@@ -0,0 +1,32 @@
...
+open Mirage_impl_mclock
+open Mirage_impl_pclock
+open Mirage_impl_stack
+open Mirage_impl_random
+
+type dns_client = Dns_client
+
+let dns_client = Type.v Dns_client
+
+let generic_dns_client ?(timeout = 5000000000L) nameservers =
```
hannesm, line 12: could we not have a default for the timeout, please? It is an optional argument in Dns_client_mirage.Make().connect.

Having the default only once, and in the dns client library, is much nicer than sprinkling around this number.

hannesm:
Apart from two minor things this looks great! Thanks for doing this work.

lib/mirage/impl/mirage_impl_dns.ml
hannesm: Good question, I'm in favour of a command-line argument - similar to `nameservers`

dinosaure: I just added a commit where `timeout` becomes a required key to build a `dns_client` device (note that it can be an optional argument). WDYT?

hannesm: I'm not sure, maybe an optional keyword argument -- i.e. that a user of that device is not required to pass the key? I'm not sure I understand the design space -- if you think this is a good solution, go with it :)

dinosaure: Yes, I agree with you so I changed and the user can optionally define a `dns-timeout` argument if he/she wants but it's not needed.

Issue #1287, opened 2022-02-07
We should not allow packages with a dot which leads a spurious solver error on the `opam monorepo` side


dinosaure:
We probably add a simple linter which checks that packages don't have dots in their names. I already get this error several times but it becomes hard to find it now where `opam monorepo` silently tries to solve the malformed OPAM file. Related to ocamllabs/opam-monorepo#255.

dinosaure:
So it's fixed but I let the issue open. We probably need a message which explain why the package name is bad (and we should probably advise the user to use `sublibs`).

hannesm:
There recently was a report on discord where a non-existing package (named "Zarith" (instead of "zarith")) was put into the config.ml -- which resulted in hard-to-read opam-monorepo/0install-solver error message. This makes me think whether we should do a bit more careful checking whether whatever you have in config.ml is actually nice in your local opam switch?

reynir:
Yes, I got a similar not at all helpful error message when I had an unpublished library in config.ml that I had forgot to pin in that opam switch. The error message made me think there was a dependency on OCaml>=5 IIRC.

PR #1304, opened 2022-03-15
Check that the package name respects opam conventions

TheLortex:
A fix for https://github.com/mirage/mirage/issues/1287

So this way, an invalid package name yields `Invalid_argument` at configure-time. We could add a hint (for example if there is a dot in the name) but at least the error is not propagated to opam-monorepo !


dinosaure:
Thanks! Seems good enough. Indeed, we probably should add a message if such error appear but that fine for the moment.

Issue #1308, opened 2022-03-21
`mirage build -f foo.ml` does not work


dinosaure:
The generated `dune.build` file wants to discard `config.ml` in any cases, we should consider it according to our `-f` option.

dinosaure:
Close by #1309

PR #1309, opened 2022-03-21
Propagate the user specified config.ml to the generated dune file

dinosaure:
A simple fix for #1308

### DIFF
--- a/lib/mirage/target/solo5.ml
+++ b/lib/mirage/target/solo5.ml
@@ -167,7 +167,7 @@ let main i =
 |}
     (context_name i) main (pp_list "libraries") libraries (pp_list "link_flags")
     flags (solo5_abi target) Fpath.pp
-    (Fpath.rem_ext (Info.config_file i))
+    (Fpath.rem_ext (Fpath.base (Info.config_file i)))
 
 let subdir name s = Dune.stanzaf "(subdir %s\n %a)\n" name Dune.pp (Dune.v s)
 
--- a/lib/mirage/target/solo5.ml
+++ b/lib/mirage/target/solo5.ml
@@ -161,12 +161,13 @@ let main i =
  (modes (native exe))
  (libraries %a)
  (link_flags %a -cclib "-z solo5-abi=%s")
- (modules (:standard \ config manifest))
+ (modules (:standard \ %a manifest))
  (foreign_stubs (language c) (names manifest))
 )
 |}
     (context_name i) main (pp_list "libraries") libraries (pp_list "link_flags")
-    flags (solo5_abi target)
+    flags (solo5_abi target) Fpath.pp
+    (Fpath.rem_ext (Info.config_file i))
 
 let subdir name s = Dune.stanzaf "(subdir %s\n %a)\n" name Dune.pp (Dune.v s)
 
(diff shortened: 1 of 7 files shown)


test/functoria-test/functoria_test.ml
```diff
@@ -17,8 +17,8 @@ let run ?(keys = []) ?init context device =
   let keys = keys @ Key.Set.elements (Engine.all_keys t) in
   let packages = Key.eval context (Engine.packages t) in
   let info =
-    Functoria.Info.v ~packages ~context ~keys ~build_cmd:[ "build"; "me" ]
-      ~src:`None "foo"
+    Functoria.Info.v ~config_file:(Fpath.v "config.ml") ~packages ~context ~keys
```
samoht, line 20: Do you still need this now that we have a default argument?

lib/mirage/target/solo5.ml
```diff
@@ -152,12 +152,13 @@ let main i =
...
  (link_flags %a -cclib "-z solo5-abi=%a")
- (modules (:standard \ config manifest))
+ (modules (:standard \ %a manifest))
  (foreign_stubs (language c) (names manifest))
 )
 |}
     (context_name i) main (pp_list "libraries") libraries (pp_list "link_flags")
-    flags Key.pp_target target
+    flags Key.pp_target target Fpath.pp
+    (Fpath.rem_ext (Fpath.base (Info.config_file i)))
```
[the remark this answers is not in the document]
dinosaure: > Is the issue that -f should never have a directory part?

It's because the `mirage-skeleton` test gives `some/directory/config.ml` as the configuration file for unikernels but we want to refer into the generated `dune.build` to the `config.ml` located into the same directory.

test/functoria-test/functoria_test.ml
dinosaure: Yes, what I did in [8128e28](https://github.com/mirage/mirage/pull/1309/commits/8128e28fcb562beaed588ff3660cfbbcabb7d7a7) is about `Functoria.Lib.v`, not `Functoria.Info.v` which does not define optional values.

samoht: could you define an optional value there too? That way we don't have to break the API.

dinosaure: Hmmhmm, I don't have strong opinion about that but it seems clear that any arguments expected by `Functoria.Lib.v` are required (and so non-optional). It seems spontaneous to stick on this design.

samoht: I'm even tempted to force all the calls to this function to strip out the directory component. I remember why I removed it from the 3.* design (as well as the build-dir option): it was really unclear what that directory component is about. I think we should make this clearer in the API:
- always keep `config_file` as an optional argument with default being "config.ml": we don't expect people to change this at all.
- always check that users of that functions that are providing a custom `config_file` should call it with an empty directory (otherwise the semantics is really unclear)

Ideally add a new test-case for this. I really remember that something was a bit tricky about this (maybe the interaction between `--build-dir` and `--config-file`).

dinosaure: I don't see where `build-dir` is defined (as a command-line argument). From what I see, `build_dir = Fpath.parent args.Cli.config_file`. Can you describe which workflow can be problematic?

dinosaure:
Let's merge, `build-dir` command-line argument was deleted and this PR fix #1308

PR #1311, opened 2022-03-23
mirage.opam & functoria.opam: require opam 2.1.0, fix #1239

### LINKED ISSUE #1239: warn the user if mirage is used with opam<2.1
As `make depends` will not work as expected if opam-2.0 is used, we should make sure we tell the user if the wrong version of opam is used.

hannesm:


dinosaure:
Thanks!

PR #1313, opened 2022-03-23
revise flags for dune file: remove -safe-string, remove custom warnings

hannesm:
just use dune's defaults.

### DIFF
--- a/lib/mirage/target/mirage_dune.ml
+++ b/lib/mirage/target/mirage_dune.ml
@@ -7,12 +7,9 @@ let flags i =
   let warn_error = Key.(get ctx warn_error) in
   [
     "-g";
-    "-w";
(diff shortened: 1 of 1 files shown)


dinosaure:
> I've to admit I'm not sure about this. Maybe the warn-error should as well be removed, and everything should be left to dune? IIUC a developer could just dune build --release and dune build for builds?

I think it's a relic of `ocamlbuild` and @TheLortex kept these flags to ensure that everything compile. Note that you must `dune promote` & `dune build @fmt` on this PR.

samoht:
I'm in favour of removing all of the custom warnings and only keep the dune's default.

samoht:
I've done something a bit more drastic with #1316

PR #1314, opened 2022-03-24
Rename ocaml-freestanding to ocaml-solo5

### LINKED ISSUE #1294: Rename `ocaml-freestanding` to `ocaml-solo5`
This issue is to track a deep rename of `ocaml-freestanding` to `ocaml-solo5`. Indeed, it's related to the issue #1284 where `ocaml-freestanding` is inside a complex dependency graph. An installation of `mirage.4.0.0~beta3` and a compilation of an unikernel for Solo5 targets leads a recompilation of the whole user's ecosystem due to `mirage-no-solo5`/`mirage-no-xen`. These projects are not needed anymore and libraries should not depends on `mirage-*-{solo5,xen}`  in anyways.

The easiest solution to move forward is to rename `ocaml-freestanding` into `ocaml-solo5`. By this fact, `mirage-no-solo5`/`mirage-no-xen` will just be forgotten by any of us (and libraries) and we don't ask the user to recompile all _upstream_ dependencies )because `ocaml-freestanding` is an optional dependency of these libraries).

The second step will be, of course, a deletion of `freestanding` artifact but it's more related to: #1262. This solution take more times and it will be hard to synchronize everyone about that.

Finally, the name `ocaml-solo5` is mostly because `ocaml-freestanding` is currently slightly dependent to what Solo5 provide for Solo5 & Xen targets.

This issue want to track such move:
- [x] prepare the `mirage` tool to emit `ocaml-solo5` instead of `ocaml-freestanding`
- [x] clean `mirage-*-{solo5,xen}` if they need about the new dependency path
- [x] rename `ocaml-freestanding`
- [x] release all of them: `ocaml-solo5` -> `mirage` -> `mirage-*-{solo5,xen}`

dinosaure:
Assume that `ocaml-solo5.0.8.0` is released (which should be just a GitHub rename), this PR wants to fix #1294. The rename can be quick or _clean_: indeed, we have several things which takes the `freestanding` name (such as the `dune`'s context, `ocamlfind`'s toolchain). What should be the best for us according to our deadline (/cc @mirage/core)?

_to test with_ https://github.com/mirage/mirage-dev/pull/387

### DIFF
--- a/lib/mirage/impl/mirage_impl_block.ml
+++ b/lib/mirage/impl/mirage_impl_block.ml
@@ -69,7 +69,7 @@ let block_conf file =
     Key.match_ Key.(value target) @@ function
     | #Mirage_key.mode_xen -> xen_block_packages
     | #Mirage_key.mode_solo5 ->
-        [ package ~min:"0.7.0" ~max:"0.8.0" "mirage-block-solo5" ]
+        [ package ~min:"0.8.0" ~max:"0.9.0" "mirage-block-solo5" ]
     | #Mirage_key.mode_unix ->
         [ package ~min:"2.12.0" ~max:"3.0.0" "mirage-block-unix" ]
   in
--- a/lib/mirage/impl/mirage_impl_console.ml
+++ b/lib/mirage/impl/mirage_impl_console.ml
@@ -15,7 +15,7 @@ let console_xen str =
   impl ~packages ~connect:(connect str) "Console_xen" console
 
 let console_solo5 str =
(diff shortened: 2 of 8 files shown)


dinosaure:
To clarify the error, `mirage-{net,block,console}-solo5` must be updated to expect the `solo5` `%{context_name}` instead of the `freestanding`. So the work is bigger than what we said at the beginning.

dinosaure:
https://github.com/mirage/mirage-dev/pull/387

PR #1324, opened 2022-04-14
Be able to make a docteur image from a relative path

dinosaure:
This fix is related to the recent release of `docteur.0.0.5` where we are able to craft an image from a **relative** path.


dinosaure:
Assume that it's correct, the `mirage-ci` fails for 2 reasons:
1) `apt install` on `arm64` fails
2) it seems that we have a problem on `xen` about constraints (specially about `shared-memory-ring`, /cc @hannesm) but the upcoming release will fix that I thinl.

Issue #1323, opened 2022-04-14
Strictly build what we need to deploy the unikernel and build block device into the default context


dinosaure:
Currently, `mirage build` is equivalent to `dune build --root .`. In such case, we don't really know what we build. It will be nice to expand `mirage build` to what we really want to build such as: `dune build --root . dist/unikernel.hvt` for example.

The question becomes tricky for artifact such as block-device produced by something like `tar` or `docteur`: `dune build --root dist/unikernel.hvt dist/disk.img`. But, as far as I can say, we already collect into `mirage_impl_block` these block-device.

Then and a last point, it seems a non-sense to try to produce a block-device from the Solo5/cross-compilation context. Indeed, we probably want to produce some files like some `*.js` from some OCaml code and pack them into an image then. In that situation, `js_of_ocaml` does not play well with our cross-compilation toolchain and it seems more accurate to limit the production of block-devices into the `default` context (not the cross-compilation context).

If we run the production of block-devices into the default context AND we explicitly asks `dune` to produce our unikernel and block-devices, we finally are able to:
- prepend the production of our block-devices by some host computations
- still produce everything needed by our unikernel with a simple `mirage build`

Such design is not a big deal but I probably missed some implications and this is why I would like to discuss about that first. /cc @TheLortex who is the most aware about all of that.

PR #1325, opened 2022-04-15
Build non-executable artifacts in the host context

dinosaure:
A context is needed to explain this PR and it's related to #1323.

The goal is to let the user to define some JS files which will be produced by `js_of_ocaml`. Then, a block-device should be generated to the user which includes JS files. A concrete example is [pasteur](https://github.com/dinosaure/pasteur) where the JS script comes from an OCaml code - by this way, we are able to make a website only in OCaml 🚀 .

However, I encountered several problems specially about the "cross-compilation" context. Indeed, it's a non-sense to let the user to fetch and compile `js_of_ocaml` into our (current) Solo5 context. From such situation, we must precise that some artifacts needed for our unikernel (such as block-device) must be buildable only into the default context/with the host compiler.

This PR has 3 commits:
- One is about the deletion of `(alias ...` for the Solo5 target which restrict `dune` to build only what is possible to build _via_  the `solo5` context. If we want to be able to build everything even if we have **multiple** contexts, we must let `dune` to build **everything**. The commit explain another solution which was a no-go due to the disability to pre-compute final artifacts.
- The second is about the context specialization of some artifacts which are installable as `etc` files (configuration files, artifacts). With this patch, only `bin` (such as our `unikernel`) need the cross-compilation context. `etc` objects will require the default context and they will use the host compiler
- The last commit is more specific about `docteur` where we needs to add an `extra_deps` argument to notice to `dune` that, even if a file into the directory `dir/` can be produced, it is really needed to craft the our `docteur` image.

Finally, with this PR, I'm able to describe a complex compilation pipeline which includes `js_of_ocaml`!

### DIFF
--- a/lib/functoria/install.mli
+++ b/lib/functoria/install.mli
@@ -20,7 +20,7 @@ type t
 
 val v : ?bin:(Fpath.t * Fpath.t) list -> ?etc:Fpath.t list -> unit -> t
 (** [v ~bin:\[(src,dst),...\] ~etc ()] is the installation of [src] as [dst] as
-    binary files, and [etc] as configuration. *)
+    binary files, and [etc] as configuration/artifact. *)
 
 val union : t -> t -> t
 (** [union a b] merge to sets of installation rules. *)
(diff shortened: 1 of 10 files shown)


lib/functoria/install.mli
```diff
@@ -34,9 +34,12 @@ val pp : t Fmt.t
 val pp_opam : t Fmt.t
 (** Print the opam rules to install [t] *)
 
-val dune : context_name:string -> t -> Dune.t
+val dune : context_name:string * string -> t -> Dune.t
```
TheLortex, line 37: `string * string` is a bit confusing, we don't know which one is the host and which one is the target context name

TheLortex (changes requested):
Hi, thanks for this PR.

Looks good overall, I only have a small comment about the use `string * string` that might be confusing in the future.

PR #1327, opened 2022-04-20
Adapt to opam-monorepo 0.3.0

TheLortex:
Opam-monorepo 3.0.0 has 3 changes that impacts the mirage tool:
- the `--root` option takes an absolute path for argument
- some packages can be marked as opam-installed: that means that the mirage tool can generate a single opam file that is understood both by opam and opam-monorepo
- by using the `--require-cross-compile` option, we'll be able to remove `mirage-opam-overlays` and move the mirage-specific packages (`zarith`) in `opam-overlays`

Mirage-skeleton changes: https://github.com/mirage/mirage-skeleton/pull/340

### DIFF
--- a/lib/functoria/cli.ml
+++ b/lib/functoria/cli.ml
@@ -31,7 +31,7 @@ let description_section = "DESCRIBE OPTIONS"
 type query_kind =
   [ `Name
   | `Packages
-  | `Opam of [ `Switch | `Monorepo ]
+  | `Opam
   | `Files
   | `Dune of [ `Config | `Build | `Project | `Workspace | `Dist ]
   | `Makefile ]
@@ -40,8 +40,7 @@ let query_kinds : (string * query_kind) list =
   [
     ("name", `Name);
     ("packages", `Packages);
-    ("monorepo.opam", `Opam `Monorepo);
-    ("switch.opam", `Opam `Switch);
+    ("opam", `Opam);
     ("files", `Files);
     ("Makefile", `Makefile);
     ("dune.config", `Dune `Config);
--- a/lib/functoria/cli.mli
+++ b/lib/functoria/cli.mli
@@ -62,7 +62,7 @@ type 'a help_args = 'a args
 type query_kind =
   [ `Name
   | `Packages
-  | `Opam of [ `Monorepo | `Switch ]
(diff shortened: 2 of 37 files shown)


TheLortex:
There are issues with regards to the install of the switch packages. The question is whether we:
- install them from the lockfile, requiring more extensive changes in the switch (in particular requiring opam-overlays to be available at all time)
- install them from the opam file, mutating the switch as less as possible and allowing to keep temporarily adding opam-overlays during the lock operation (as done currently). The cost here is that more discrepancies can happen (for example `dune 2` in the switch and `dune 3` in the monorepo, causing a build failure).

Related issue: https://github.com/ocamllabs/opam-monorepo/issues/289

dinosaure:
> install them from the opam file, mutating the switch as less as possible and allowing to keep temporarily adding opam-overlays during the lock operation (as done currently). The cost here is that more discrepancies can happen (for example dune 2 in the switch and dune 3 in the monorepo, causing a build failure).

I already noticed this kind of problem when a package explicitly needs `dune.3.0` where my switch provide only `dune.2.0`. I think, the second solution (install switch dependencies from the OPAM file as we currently do) seems the best. However, we probably should ask @hannesm about such workflow and be aligned with reproducible with as much as we can.

TheLortex:
@dinosaure Maybe give it a go with your unikernels but it should be good !

hannesm:
"install of the switch packages" -- so what would be the actual case of a package not released to opam-repository, but only to "opam-overlays"? I can't think of anything tbh, and would prefer to not use the "opam-overlays" for the "switch packages".

TheLortex:
> "install of the switch packages" -- so what would be the actual case of a package not released to opam-repository, but only to "opam-overlays"? I can't think of anything tbh, and would prefer to not use the "opam-overlays" for the "switch packages".

With the current state of the PR it's not doing that, and instead it's keeping the original behaviour.

The situation would occur if the packages were installed as specified in the _lockfile_, because its generated with opam-overlays enabled.

yomimono:
Sorry to have blocked you! Just merged the block size change in chamelon and will cut a release shortly; the version will be 0.0.8 .

dinosaure (approved):

PR #1332, opened 2022-05-19
various makefile and opam file tweak

hannesm:
this is a follow up to bdb8921ff6d23a10243743daac2073279bbf01a1

To test with: https://github.com/mirage/mirage-skeleton/pull/343

### DIFF
--- a/lib/functoria/lib.ml
+++ b/lib/functoria/lib.ml
@@ -321,9 +321,7 @@ module Make (P : S) = struct
     let build_dir = build_dir args in
     let name = P.name_of_target info in
     let opam_name = Misc.Name.opamify name in
-    let* () =
-      generate_makefile ~build_dir ~depext ~extra_repo (Misc.Name.opamify name)
-    in
+    let* () = generate_makefile ~build_dir ~depext ~extra_repo opam_name in
     let* _ = Action.mkdir (mirage_dir args) in
     let* () =
       Action.with_dir (mirage_dir args) (fun () ->
--- a/lib/functoria/makefile.ml
+++ b/lib/functoria/makefile.ml
@@ -31,7 +31,7 @@ let depext_rules =
   {|
 depext-lockfile: $(MIRAGE_DIR)/$(UNIKERNEL_NAME).opam.locked
 	echo " ↳ install external dependencies for monorepo"
-	$(OPAM) monorepo depext -y -l $<
+	env OPAMVAR_switch="monorepo" $(OPAM) monorepo depext -y -l $<
 |}
 
 let opam_repo_add_rule extra =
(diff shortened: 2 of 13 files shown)


dinosaure (approved):

hannesm:
I pushed further commits, best to be read commit-by-commit, and would be interested if these are fine to be merged. The behaviour shouldn't have changed.  /cc @samoht @TheLortex @dinosaure

dinosaure:
It seems indeed nice to explicitely write `{switch = "monorepo"}` into the OPAM file. If the CI is green, I don't have strong opinion against this PR :+1:.

hannesm:
I dropped the `remove name field from opam` since it was causing CI issues.

Now the name is `hello-key`, but the opam file name is `hello-key-hvt` -- is this expected or an inconsistency? Also, in MirageOS3 we used the `mirage-unikernel` prefix for the package name, was this removed for a good reason?

hannesm:
This requires https://github.com/mirage/mirage-skeleton/pull/343 but I have no clue how to educate our nice CI system about that fact.

dinosaure:
> This requires https://github.com/mirage/mirage-skeleton/pull/343 but I have no clue how to educate our nice CI system about that fact.

A mention of this PR is enough to tell `mirage-ci` to take this version of `mirage` but it seems that `mirage-ci` is _down_ - or it did not catch the request to recompile everything with what you did. /cc @TheLortex.

TheLortex:
To ask the CI to build two PRs together, you have to edit the original message to mention the other PR Github link. It's not looking at comments.

I will do it for you for this one

hannesm:
Thanks @TheLortex. Any chance you may do a review of the technical changes and express your opinion about them (esp. [f3639e3](https://github.com/mirage/mirage/pull/1332/commits/f3639e34ec4e139bcd345ae65af0f8f6cf3b124a))?

TheLortex:
The change sounds sensible. I didn't expect `opam-monorepo` to understand the `OPAMVAR_*` environment variables but it makes sense as they are using the opam libraries. My initial implementation was convoluted in particular because of that, so it's good if we can improve it.

Now I can suggest an even clearer change, based on the fact that we probably should **rename the `switch` variable**, as the `switch` variable name is an existing opam variable. In particular using `{ switch = "monorepo"}` would create a collision when used in a switch whose name is  "monorepo". I used it in the first place because I thought we _couldn't_ introduce new variables names in opam-monorepo.

Now that we can, we define a `monorepo` variable, and use a filter like `{ ?monorepo }` along with `OPAMVAR_monorepo=` to enable the variable:
- `opam` would ignore these fields as the `monorepo` variable doesn't exist
- `OPAMVAR_monorepo= opam-monorepo` would require these fields and work as expected.

hannesm:
@TheLortex done in [cae07da](https://github.com/mirage/mirage/pull/1332/commits/cae07dac1dc60466e8799077b881cd18a4941c5e)

Maybe the `opam-monorepo` developers are fine to introduce such an opam-environment variable (`OPAMVAR_monorepo="opam-monorepo"`) in their tool, so we don't have to prefix all invocations with it? //cc @NathanReb @Leonidas-from-XIV

TheLortex:
For info the content of the `OPAMVAR_monorepo="opam-monorepo"` environment variable doesn't matter as the check `{ ?monorepo }` only checks for its existence. We can use `OPAMVAR_monorepo=true` as a simpler solution for example.

Leonidas-from-XIV:
May I inquire about the significance of the `monorepo` variable? We're open to adding all kinds of features that make sense for users of opam-monorepo, but for that we'd need a suggestion/spec to figure out what is required and what's supposed to be solved by that - because maybe there exists already a way to do whatever you desire in some kind of other way. In any case, I think opening an issue on `opam-monorepo` might be the best way to go about new feature requests.

hannesm:
@TheLortex true. well, I'm not sure what is the best to do, and would leave it as is. Is this fine to be merged from your perspective? CI run with the updated mirage-skeleton is successful.

@Leonidas-from-XIV being able to specify in an opam file -- used by opam monorepo -- that a certain dependency is only to be considered by opam-monorepo. The solution as done in mirage (with opam-monorepo >= 0.3) is to use an opam variable (specify `OPAMVAR_monorepo="something"` while executing `opam monorepo`) and that allows each dependency in the opam file to conaint `"my-package" { ?monorepo }`. Sure, when I have time, I'd open a proposal on opam-monorepo. But given my experience there (no comment on an issue I opened > 6 months ago), I actually feel more productive to ask in the context (here) and gather some opinions.

hannesm:
> Would you be willing to wait for this transition to complete before removing the wrapping?

Sure. Take your time.

hannesm:
And thanks for your kind and detailed reply @NathanReb.

hannesm:
Since there was no complaint, and all suggestions have been fulfilled, merging. Thanks for your feedback.

PR #1334, opened 2022-06-15
Fix tests with opam-monorepo.0.3.2

dinosaure:
A simple `dune promote` with `opam-monorepo.0.3.2` (probably related with #1327)


hannesm:
Thanks, the OCaml-CI failure is with OCaml 5 -- where lwt fails (likely fixed soon in opam-repository and by a lwt release).

PR #1335, opened 2022-06-20
Reproducible

hannesm:
this is a draft to enable reproducible builds with mirageos 4.x

### DIFF
--- a/lib/functoria/install.ml
+++ b/lib/functoria/install.ml
@@ -43,12 +43,12 @@ let pp ppf t =
 let pp_opam ?subdir () ppf t =
   let pp_bin ppf (src, dst) =
     Fmt.pf ppf {|"cp" "%adist/%a" "%%{bin}%%/%a"|}
-      Fmt.(option ~none:(any "") (Fpath.pp ++ any "/"))
+      Fmt.(option ~none:(any "") Fpath.pp)
       subdir Fpath.pp src Fpath.pp dst
   in
   let pp_etc ppf etc =
     Fmt.pf ppf {|"cp" "%adist/%a" "%%{etc}%%"|}
-      Fmt.(option ~none:(any "") (Fpath.pp ++ any "/"))
+      Fmt.(option ~none:(any "") Fpath.pp)
       subdir Fpath.pp etc
   in
   Fmt.pf ppf "\n%a\n"
--- a/lib/functoria/install.ml
+++ b/lib/functoria/install.ml
@@ -40,11 +40,17 @@ let pp ppf t =
   Fmt.pf ppf "etc: [%s%s]" (String.concat "" etcs)
     (match etcs with [] -> "" | _ -> "\n")
 
-let pp_opam ppf t =
+let pp_opam ?subdir () ppf t =
   let pp_bin ppf (src, dst) =
-    Fmt.pf ppf {|"cp" "dist/%a" "%%{bin}%%/%a"|} Fpath.pp src Fpath.pp dst
+    Fmt.pf ppf {|"cp" "%adist/%a" "%%{bin}%%/%a"|}
+      Fmt.(option ~none:(any "") (Fpath.pp ++ any "/")) subdir
+      Fpath.pp src Fpath.pp dst
   in
-  let pp_etc ppf etc = Fmt.pf ppf {|"cp" "dist/%a" "%%{etc}%%"|} Fpath.pp etc in
+  let pp_etc ppf etc =
+    Fmt.pf ppf {|"cp" "%adist/%a" "%%{etc}%%"|}
+      Fmt.(option ~none:(any "") (Fpath.pp ++ any "/")) subdir
+      Fpath.pp etc
+  in
   Fmt.pf ppf "\n%a\n"
     (Fmt.list ~sep:(Fmt.any "\n") (fun ppf -> Fmt.pf ppf "  [ %a ]" pp_bin))
     t.bin;
--- a/lib/functoria/install.ml
+++ b/lib/functoria/install.ml
@@ -43,13 +43,13 @@ let pp ppf t =
 let pp_opam ?subdir () ppf t =
   let pp_bin ppf (src, dst) =
     Fmt.pf ppf {|"cp" "%adist/%a" "%%{bin}%%/%a"|}
-      Fmt.(option ~none:(any "") (Fpath.pp ++ any "/")) subdir
-      Fpath.pp src Fpath.pp dst
+      Fmt.(option ~none:(any "") (Fpath.pp ++ any "/"))
+      subdir Fpath.pp src Fpath.pp dst
   in
   let pp_etc ppf etc =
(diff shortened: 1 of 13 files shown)


dinosaure (approved):
I look into the PR in details (specially about the `extra-repo` variable added into #1270) and it seems good for me. The CI is green, so if it helps you to provide a reproducible for MirageOS 4, I will happy to merge it. May be a double-check from @TheLortex.

hannesm:
I'll comment here again once satisfied with the flow of build and rebuild. There may be some more changes needed for this.

TheLortex:
Thank for this work, this is very nice ! I will take some time this week to review the PR.

hannesm:
With the current version, I'm satisfied that a non-interactive non-opam driver can build unikernels using the mirage-generated opam file as follows:
- execute an `opam install --deps-only .`
- use the value of `x-mirage-extra-repo` to be added to the repositories
- execute the `x-mirage-configure` step
- execute `x-mirage-pre-build` (which execute `make lock pull`)
- gather extra duniverse sources from the `x-mirage-opam-lock-location` opam file (`x-opam-monorepo-duniverse-dirs` (this is needed for reproducing the exact same binary and keeping all urls to tarballs with checksums into a single place)) (of course the `opam monorepo pull` does this step automatically, the fetching)
- execute `build` and `install` instructions

So, with this PR, there are several fields added to the generated opam files:
- `build` only includes the `mirage build` term (but may include a `cd <subdir>` first, e.g. when building tutorial/hello from mirage-skeleton);
- `install` the same (includes the subdirectory)
- `x-mirage-opam-lock-location` a string with the path to the lockfile (tutorial/hello/mirage/hello-hvt.opam.locked)
- `x-mirage-configure` the configure that was executed (`mirage configure -t hvt --no-extra-repo`)
- `x-mirage-pre-build` the monorepo lock pull command

Wiht this PR, I'm able to conduct unikernel builds with orb -- which is great. I'm not yet able to conduct a rebuild (of the same binary), which should avoid the opam repository (repositories) and only use the tarballs from the opam switch export (which includes the `x-opam-monorepo-duniverse-dirs` for this purpose).

A review of this PR would be great, I suspect that the mirage-generated opam file is not used anywhere, so these changes should be fine.

hannesm:
An example opam file of the `tutorial/hello` unikernel is located at https://git.robur.io/hannes/test-opam/src/branch/main/packages/mirage-unikernel-hello-hvt.0.0.1/opam

hannesm:
> [TheLortex](https://github.com/TheLortex) commented [on 21 Jun](https://github.com/mirage/mirage/pull/1335#issuecomment-1161761846)
>
> Thank for this work, this is very nice ! I will take some time this week to review the PR.

Any chance you had time to look into this? Would be great to have this reviewed and merged for the next mirage release.

lib/functoria/install.ml
```diff
@@ -40,11 +40,17 @@ let pp ppf t =
   Fmt.pf ppf "etc: [%s%s]" (String.concat "" etcs)
     (match etcs with [] -> "" | _ -> "\n")
 
-let pp_opam ppf t =
+let pp_opam ?subdir () ppf t =
   let pp_bin ppf (src, dst) =
-    Fmt.pf ppf {|"cp" "dist/%a" "%%{bin}%%/%a"|} Fpath.pp src Fpath.pp dst
+    Fmt.pf ppf {|"cp" "%adist/%a" "%%{bin}%%/%a"|}
+      Fmt.(option ~none:(any "") (Fpath.pp ++ any "/"))
```
TheLortex, line 46: `Fpath.pp ++ any "/"` can output a `//` when `subdir` already ends with a `/`. 

It's the case in the example opam file that you linked: https://git.robur.io/hannes/test-opam/src/branch/main/packages/mirage-unikernel-hello-hvt.0.0.1/opam#L18

lib/functoria/install.mli
```diff
@@ -31,7 +31,7 @@ val empty : t
 val pp : t Fmt.t
 (** Print the .install rules to install [t] *)
 
-val pp_opam : t Fmt.t
+val pp_opam : ?subdir:Fpath.t -> unit -> t Fmt.t
 (** Print the opam rules to install [t] *)
```
TheLortex, line 35: Extending the comment to explain the `~subdir` parameter would be useful

TheLortex (changes requested):
The patch looks good, and indeed won't break anything as currently nothing makes use of the generated opam file. I have added two small comments.

Additionally, I think we should either (or both):
- document how to use the opam file with these extra fields. Maybe in `opam.mli`
- provide a minimal _driver_ that is able to build an unikernel from the opam definition. This would enable having a command such as `mirage-install https://git.robur.io/hannes/test-opam/raw/branch/main/packages/mirage-unikernel-hello-hvt.0.0.1/opam` or `mirage install <OPAM FILE PATH>` depending on if we want that feature in the mirage tool or as an additional binary.

What do you think ?

lib/functoria/install.ml
hannesm: done in e62652ba

lib/functoria/install.mli
hannesm: done in 7148bba9

hannesm:
Thanks for your review and comments. I addressed your minor comments in subsequent commits.

> document how to use the opam file with these extra fields. Maybe in opam.mli

Is such a documentation in opam.mli worth it? I'm more into writing a post for our website that describes these extra fields and how they can be used.

> provide a minimal driver that is able to build an unikernel from the opam definition. This would enable having a command such as mirage-install https://git.robur.io/hannes/test-opam/raw/branch/main/packages/mirage-unikernel-hello-hvt.0.0.1/opam or mirage install <OPAM FILE PATH> depending on if we want that feature in the mirage tool or as an additional binary.

I've no clue what a "minimal driver" is. From my perspective, orb (the opam reproducible builder) is the minimal application that takes such an mirage-generated opam file as input and outputs a unikernel binary (and sufficient information to reproduce the exact same binary). If you like to develop something even smaller (mirage install?), please give it a try. For me, orb is mostly unrelated and independent of mirage -- thus its dependency cone does neither include "mirage" nor "opam" being available on the host system (and thus can be used in minimal systems (jail/container).

TheLortex:
> Is such a documentation in opam.mli worth it? I'm more into writing a post for our website that describes these extra fields and how they can be used.

A blog post would be nice indeed. It's just that the link between these fields and orb or reproducible builds is not clear for the unsuspecting user that woud look into these generated files, definitely that's not critical though.

>  I've no clue what a "minimal driver" is. From my perspective, orb (the opam reproducible builder) is the minimal application that takes such an mirage-generated opam file [...]

I've looked into your MirageOS 4 work for `orb` : I meant something like that, so indeed there's no point duplicating with something else. Instead we can add a `Reproducible builds` paragraph in the README to explain the workflow and add a pointer to `orb`. It can be done in another PR.

TheLortex (approved):

hannesm:
Thanks @TheLortex, I'll merge and cut a release soon.

PR #1337, opened 2022-07-14
mirage.mli: update deprecated references

reynir:
I updated by looking at the deprecation comments in #1006 and by following the comment when installing `mirage-protocols`. There may be more dated references in the comments.


hannesm:
Thanks a lot, CI failure is unrelated.

Issue #1339, opened 2022-07-27
sunsetting stackv4 and stackv6


hannesm:
Hey,

is there anyone still using stackv4 / stackv6 (instead of the dual stackv4v6)? I'd like to remove it to reduce complexity and code size.

It was nice to have for migrating unikernels from a v4 only land to dual stack. But the dual stack allows at initialization time to have v4 / v6 disabled (even exposed as boot parameter) -- so I don't see any argument to keep the old stacks around.

Removing them in the mirage tool will make this less complex, removing the implementations from TCPIP will make it easier to move forward (revise API etc.).

What do you think?

palainp:
Hi @hannesm, I think it's a good move (both for less code to maintain, and easier API updates)!

I just searched for `org:mirage generic_stackv4` here in github (not sure if this is a good metric for an impact estimate) and it returns 159 Ocaml code results (still most of them are in mirage-www-deployment, the others are in testsuites or examples). There is only 3 matches in mirage-skeleton (in tracing, network and pgx).

If I search the whole of github, I have 300+ Ocaml matches, and around 50 of them were indexed in 2022.

hannesm:
Thanks @palainp, so:
- mirage-skeleton https://github.com/mirage/mirage-skeleton/pull/344 just uses stackv4v6
- apart from the pgx example -- anyone using pgx? it seems to have bitrotted a bit
- in this repository, #1341 is my proposal.

yomimono:
I might be able to get the pgx example working again. I'm not actively using it in a unikernel context but I have aspirations to, and I have other projects with code written against it.

yomimono:
Late reply, but this change seems good to me. :+1:

hannesm:
merged here, soon to be released

PR #1341, opened 2022-07-28
remove v4 & v6 stack, other deprecated bindings

hannesm:
see #1339 for explanation. as @palainp mentioned there, some unikernels in mirage-skeleton will need to be updated. (I'll follow up with a PR soon).


hannesm:
I'll merge this, this was agreed to by today's meeting.

Issue #1344, opened 2022-08-25
When configuring for an invalid target the error message is misleading


RyanGibb:
```
mirage configure -t hello
Fatal error: exception Failure("Invalid cached configuration. Please run configure again."
```

PR #1350, opened 2022-09-07
Improve functoria cli

TheLortex:
The two commits are fixing two separate issues:

## Context cache: don't fail if configuration is invalid

Fixes https://github.com/mirage/mirage/issues/1344

Basically, the tool can create an invalid context cache file if the command is wrong, as a parse error will still trigger re-execution of the `config.exe`. However this `config.exe` will call into `Context_cache.peek` which will then fail with `Invalid cached configuration. Please run configure again.`. The fix is to simply let `peek` return `None` in that situation. 

Another solution might be to let `config.exe` store arguments the context cache only after verifying that it is valid. That means a bit of refactoring as the tool is currently the one in charge of it.

## handle the situation when global arguments have a parse error

Currently, there's an `assert false` in `Cli.peek_args`:
```ocaml
let peek_args ?(with_setup = false) ~mname argv =
   let args =
     Subcommands.T.args { with_setup; mname; context = Term.const () }
   in
   match Cmd.eval_peek_opts ~argv args with
   | _, Ok (`Ok b) | Some b, _ -> b
   | _ -> assert false
```

This means that incorrectly using global configuration options lead to an assertion failure. (`mirage configure -f` without argument for example). This PR changes the function to return an `option` type, so that in `Tool.run_with_argv` it can detect that situation, and display the parse error instead of failing.

### DIFF
--- a/lib/functoria/cli.ml
+++ b/lib/functoria/cli.ml
@@ -203,6 +203,15 @@ type 'a args = {
   dry_run : bool;
 }
 
+let default_args =
+  {
+    context = ();
+    config_file = Fpath.v "dummy";
+    context_file = None;
+    output = None;
+    dry_run = false;
+  }
+
 type 'a configure_args = {
   args : 'a args;
   depext : bool;
@@ -463,8 +472,8 @@ let peek_args ?(with_setup = false) ~mname argv =
     Subcommands.T.args { with_setup; mname; context = Term.const () }
   in
   match Cmd.eval_peek_opts ~argv args with
-  | _, Ok (`Ok b) | Some b, _ -> b
-  | _ -> assert false
+  | _, Ok (`Ok b) | Some b, _ -> Some b
+  | _ -> None
 
 let eval ?(with_setup = true) ?help_ppf ?err_ppf ~name ~version ~configure
     ~query ~describe ~build ~clean ~help ~mname argv =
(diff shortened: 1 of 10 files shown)


samoht:
LGTM - could you add a few regression tests to make sure this keeps working?

dinosaure (approved):
Nice to delete more `assert false` 👍

TheLortex:
I have added regression tests. However I have also found a case that is not correctly handled.

I have the following logs:
```
 Parsing error in global arguments
   $ ./test.exe -o
+  test: unknown option '-o'.
+        unknown option '--dry-run'.
+  Usage: test [COMMAND] …
+  Try 'test --help' for more information.
+  [1]

 Parsing error in global arguments, with subcommand
   $ ./test.exe configure -f -o
+  test: option '-f' needs an argument
+  Usage: test configure [OPTION]…
+  Try 'test configure --help' or 'test --help' for more information.
+  [1]
```

The first case is not correctly handled as it should know about the `-o` and `--dry-run` options.  That's because the `default` term is `Term.(ret (const usage) $ setup ~with_setup`, which doesn't declare the general arguments. 

Surprisingly, declaring them using `Term.(ret (const usage) $ T.args { with_setup; mname; context = Term.const ()})` doesn't fix the issue and will display the help page instead. I'll continue investigating.

PR #1352, opened 2022-09-07
tar-mirage: raise upper bound to 3.0.0 -- the 2.0.0 series didn't break the API

hannesm:


dinosaure:
For the record, 2.1.0 introduces the GZip support, so it's mostly an extension of the API than a change.

Issue #1348, opened 2022-08-30
functoria/Key.Arg.pair doesn't pass `~sep` to `runtime_conv`


rand00:
.. this makes the CLI interface different for configure/run-stage

hannesm:
fixed in #1354

PR #1354, opened 2022-09-07
Fix the separator for list and pair combinator

### LINKED ISSUE #1349: functoria/`Key.Arg.list` doesn't support `~sep`, but `pair` does

dinosaure:
Fix #1349 (/cc @rand00)


hannesm:
does this as well fix https://github.com/mirage/mirage/issues/1348?

rand00:
yes (:

PR #1362, opened 2022-10-05
Update to Resolver_mirage changes

reynir:
The parsing of nameserver list is delayed until runtime due to ease of creation of tls authenticator with ca-certs available. Also allow specifying multiple nameservers. The `resolver` and `resolver-port` keys are melded into one key `resolver` that also supports specifying TLS nameservers. The (complicated) format should be documented.

- [x] Depends on https://github.com/mirage/ocaml-conduit/pull/415,
- [ ] The format is documented.

### DIFF
--- a/lib/functoria/install.mli
+++ b/lib/functoria/install.mli
@@ -35,7 +35,7 @@ val pp_opam : ?subdir:Fpath.t -> unit -> t Fmt.t
 (** Print the opam rules to install [t]. If [~subdir] is provided, this will be
     used as prefix (i.e. if your unikernel is in the "tutorial/hello/"
     subdirectory (which is passed as [~subdir], the install instructions will
-    use "cp tutorial/hello/dist/hello.hvt %{bin}%/hello.hvt"). *)
+    use [cp tutorial/hello/dist/hello.hvt %{bin}%/hello.hvt]). *)
 
 val dune :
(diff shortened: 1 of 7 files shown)


hannesm:
FWIW, conduit change PRed to opam-repository (using version 6.0.0 so we need to bump the bounds in mirage). https://github.com/ocaml/opam-repository/pull/22237

hannesm:
The description of the format: see https://github.com/mirage/ocaml-dns/pull/322/commits/2a44ec172f3599d647bf0006b3a357c31edc3bd2 which emits it as an error message (and earlier, https://github.com/mirleft/ocaml-x509/commit/dd16ca400009989722d22e98d76d7f13e1b71c2a describes the authenticator format). Hope that is sufficient!?

dinosaure:
Something is wrong with the generation of the OCaml code:
```
File "device-usage/http-fetch/main.ml", line 119, characters 62-64:
119 | | Ok _ as r -> Lwt.return r | Error (`Msg e) -> invalid_arg e in
                                                                    ^^
Error: Syntax error: ')' expected
File "device-usage/http-fetch/main.ml", line 105, characters 36-37:
105 | let resolver_mirage_make__14 = lazy (
                                          ^
  This '(' might be unmatched
```

reynir:
@dinosaure It should be fixed now. Thanks!

hannesm:
Thanks for fixing this up, @reynir.

PR #1363, opened 2022-10-13
Fix Makefile generation to work with beta versions of the compiler

kit-ty-kate:
I'm not sure it's the best solution but it seems to work at least (though opam-monorepo does not work with OCaml 5.0 for some reason)

A cleaner solution would be for opam-monorepo to do that instead but i don't know its code source or the purpose of this `--ocaml-version` argument so i'll let this exercise to the people who know.

### DIFF
--- a/lib/functoria/makefile.ml
+++ b/lib/functoria/makefile.ml
@@ -116,7 +116,7 @@ all::
 
 $(MIRAGE_DIR)/$(UNIKERNEL_NAME).opam.locked: $(MIRAGE_DIR)/$(UNIKERNEL_NAME).opam%a
 	@@echo " ↳ generate lockfile for monorepo dependencies"
-	@@env OPAMVAR_monorepo="opam-monorepo" $(OPAM) monorepo lock --require-cross-compile --build-only $(UNIKERNEL_NAME) -l $@@ --ocaml-version $(shell ocamlc --version)%a
+	@@env OPAMVAR_monorepo="opam-monorepo" $(OPAM) monorepo lock --require-cross-compile --build-only $(UNIKERNEL_NAME) -l $@@ --ocaml-version $(shell ocamlc --version | sed 's/~.*//' | sed 's/\+.*//')%a
 
 lock::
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
cc @Leonidas-from-XIV for opam-monorepo

Leonidas-from-XIV:
@kit-ty-kate Thanks. It doesn't work because one of the vendored dependencies does not work on OCaml 5.0, but if you use the git repo, you can build it with 5.0: https://github.com/tarides/opam-monorepo/pull/325

kit-ty-kate:
Sorry i should have been more clear: it compiles just fine, by "not working" i meant opam-monorepo is not finding a solution ever (see the ci-dev channel on slack)

kit-ty-kate:
It looks like this is indeed the expected input of the `--ocaml-version` argument so the PR is good to go. The opam-monorepo bug is being tracked in https://github.com/tarides/opam-monorepo/issues/340

dinosaure:
I am not sure I understand this PR and what the outcome of this discussion is. It seems consistent to me that `opam monorepo` should interpret `ocamlc --version` correctly. In this, I would be closer to what @kit-ty-kate said at the beginning of this PR: `opam monorepo` should understand `ocamlc --version` even if it is at version 5.

I'm not a fan of adding an `sed` in such a situation either. Should I close this PR and assume that, upstream, a fix will be offered in `opam monorepo`? Or should I consider this use of `opam monorepo` to be our problem - and reconsider the use of `sed`?

kit-ty-kate:
@Leonidas-from-XIV what do you think?

Leonidas-from-XIV:
I think that in general `opam-monorepo` should support the versions that the compiler is available in in `opam-repository` (whether this matches what `ocamlc --version` outputs in the general case is not 100% clear to me, I tested with `5.0.0~alpha1` and it did output `5.0.0~alpha1` which is a valid opam version, but I haven't tried installing any of the `opam-variants` compilers to see if they would contain `+options` as well).

kit-ty-kate:
In that case I'll close this PR and I've opened https://github.com/tarides/opam-monorepo/issues/351

PR #1365, opened 2022-10-19
Improve the documentation about `generic_dns_client` and explain the format of the `nameserver` argument

dinosaure:
Pointed out by @Engil, it is difficult to get a description of the format required to define a specific nameserver. I have added documentation to explain the format of the latter to make it easier to use. (/cc @hannesm)

### DIFF
--- a/lib/mirage/mirage.mli
+++ b/lib/mirage/mirage.mli
@@ -625,7 +625,17 @@ val generic_dns_client :
   dns_client impl
 (** [generic_dns_client stackv4v6] creates a new DNS value which is able to
     resolve domain-name from [nameservers]. It requires a network stack to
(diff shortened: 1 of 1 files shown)


reynir (approved):
Thanks!

hannesm:
once https://github.com/mirage/ocaml-dns/pull/322/commits/2a44ec172f3599d647bf0006b3a357c31edc3bd2 is merged and released, we'll have a nice error message which describes the format in detail. I'm actually not sure where the documentation makes sense -- is it at runtime (i.e. reporting a failure) or in the ocamldoc. I dislike that we repeat it all over, and e.g. here we'll remember to update it soon...

PR #1369, opened 2022-10-25
adapt to conduit 6.0.1

hannesm:


hannesm:
CI failure is the mirage-ci not knowing about mirage 4.3.1 yet (which is not yet released).

PR #1370, opened 2022-10-28
Replace 'echo -e' with 'printf' in makefiles

gridbugs:
On macos, echo does not have a -e flag, which results in the -e and ansi escape sequences being printed literally. The printf command handles ansi escape sequences by default.


dinosaure:
Thanks, I applied `ocamlformat` on your patch (you can do that with `dune build @fmt --auto`). I think it's good to merge, just waiting the CI.

PR #1376, opened 2022-11-21
Introduce an ALPN client and Mimic

dinosaure:
This PR rename the initial `git_client` used to create a witness which is able to initiate a Git connection to `mimic` and be able to re-use the same instance for an new introduced ALPN client (http/1.1 and h2).

The ALPN client is currently used by the Robur's opam-mirror unikernel and can be used in several contexts when the user wants to do an HTTP (http/1.1 or h2) requests with TLS.

The mimic device (specially the `mimic_happy_eyeballs` device) initiate an happy_eyeballs instance which can be used and re-used by several devices such as `git_client` devices or the new introduced `alpn_devices`.

This commit breaks some names and specially some unikernels like our Git example into our mirage-skeleton repository. The main change is the renaming of `git_happy_eyeballs` to `mimic_happy_eyeballs` when the second is not really related to Git stuff (and used by something else such as our new introduced `alpn_client`).

### DIFF
--- a/lib/mirage/mirage.mli
+++ b/lib/mirage/mirage.mli
@@ -995,6 +995,10 @@ val merge_git_clients : git_client impl -> git_client impl -> git_client impl
 (** [merge_git_clients a b] is a device that can connect to remote Git
     repositories using either the device [a] or the device [b]. *)
 
+val git_happy_eyeballs :
+  stackv4v6 impl -> dns_client impl -> happy_eyeballs impl -> mimic impl
+(** @deprecated You should use {!val:mimic_happy_eyeballs}. *)
+
 val git_tcp : tcpv4v6 impl -> mimic impl -> git_client impl
 (** [git_tcp tcpv4v6 dns] is a device able to connect to a remote Git repository
     using TCP/IP. *)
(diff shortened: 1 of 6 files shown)


lib/mirage/mirage.mli
```diff
@@ -910,10 +995,7 @@ val merge_git_clients : git_client impl -> git_client impl -> git_client impl
 (** [merge_git_clients a b] is a device that can connect to remote Git
     repositories using either the device [a] or the device [b]. *)
 
-val git_happy_eyeballs :
```
samoht, line 913: could you deprecate the function instead of removing it? That'll give time for our users to migrate progressively.

lib/mirage/mirage.mli
```diff
@@ -859,6 +895,51 @@ val paf_server : port:int key -> tcpv4v6 impl -> http_server impl
...
+
+      let main = foreign "Unikernel.Make" (alpn_client @-> job)
+      let stackv4v6 = generic_stackv4v6 default_network
+      let dns = generic_dns_client stack
+
+      let alpn_client =
+        let dns =
+          mimic_happy_eyeballs stackv4v6 dns (generic_happy_eyeballs stack dns)
+        in
+        alpn_client (tcpv4v6_of_stackv4v6 stackv4v6) dns
```
reynir, line 938:
```suggestion
        paf_client (tcpv4v6_of_stackv4v6 stackv4v6) dns
```

dinosaure:
Let's merge, the error is unrelated to this PR, I will make an issue about: 1) the MacOS bug 2) the deletion of `git_happy_eyeballs` for the next release

PR #1382, opened 2022-12-17
adjust solo5 bound: use 0.8.0 as upper bound

hannesm:
//cc @dinosaure I'm a bit confused why 0.8.8 should be there. From my perspective, any 0.Y may change API in 0.(Y + 1), so best to restrict to < 0.8.0 [this is done for other opam packages as well in this package].


dinosaure:
Yeah, I probably mistyping indeed.

PR #1383, opened 2022-12-22
Add `build` stanzas to packages that are supposed to be vendored

Leonidas-from-XIV:
In https://github.com/tarides/opam-monorepo/pull/355 opam-monorepo marked packages without build instructions as virtual (e.g. packages that are conf-packages, dependency-only packages, etc).

This requires that the dummy OPAM files used here also have `build` stanzas, otherwise they are considered virtual and skipped.

(The `dune` change is only for good measure for future-proofing, similar to the compiler `dune` is treated specially anyway)


TheLortex (approved):

TheLortex:
Thanks !

PR #1389, opened 2023-02-03
Add tar_kv_ro and tar_kv_rw

reynir:
In addition, `archive` is deprecated in favor of `tar_kv_ro` and `tar_kv_rw`.

### DIFF
--- a/lib/mirage/impl/mirage_impl_block.ml
+++ b/lib/mirage/impl/mirage_impl_block.ml
@@ -101,17 +101,24 @@ let generic_block ?group ?(key = Key.value @@ Key.block ?group ()) name =
     ]
     ~default:(ramdisk name)
 
-let tar_kv_ro_conf, tar_kv_rw_conf =
+let tar_kv_ro_conf =
   let packages = [ package ~min:"1.0.0" ~max:"3.0.0" "tar-mirage" ] in
   let connect _ modname = function
     | [ block ] -> Fmt.str "%s.connect %s" modname block
-    | _ -> failwith (connect_err "tar_kv" 1)
+    | _ -> failwith (connect_err "tar_kv_ro" 1)
   in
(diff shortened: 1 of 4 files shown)


hannesm:
This looks nice in general, but the Tar_mirage.Make_KV_RW functor expects a PCLOCK and a BLOCK device, so we cannot uniformly handle it.

Also, the `deprecated` attribute shouldn't be hidden in the ocamldoc, but visible to the OCaml compiler so it can emit warnings :)

lib/mirage/impl/mirage_impl_block.ml
```diff
@@ -101,17 +101,24 @@ let generic_block ?group ?(key = Key.value @@ Key.block ?group ()) name =
...
     | [ block ] -> Fmt.str "%s.connect %s" modname block
-    | _ -> failwith (connect_err "tar_kv" 1)
+    | _ -> failwith (connect_err "tar_kv_ro" 1)
   in
-  impl ~packages ~connect "Tar_mirage.Make_KV_RO" (block @-> Mirage_impl_kv.ro),
-  impl ~packages ~connect "Tar_mirage.Make_KV_RW" (block @-> Mirage_impl_kv.rw)
+  impl ~packages ~connect "Tar_mirage.Make_KV_RO" (block @-> Mirage_impl_kv.ro)
+
+let tar_kv_rw_conf =
+  let packages = [ package ~min:"1.0.0" ~max:"3.0.0" "tar-mirage" ] in
```
hannesm, line 113: being picky, the tar-mirage KV_RW functor only appeared in 2.2.0

reynir: Ah, good catch! Not picky, it's wrong with tar-mirage<2.2.0.

hannesm:
looks good, please bow to `dune build @fmt --auto`

Issue #1330, opened 2022-05-16
Generated Makefile only support GNU make


hannesm:
as reported in #1254

>    * [ ]  the generated Makefile only works with GNU make. I suggest to name it GNUmakefile so that there are no surprises (GNU make will use GNUmakefile if present, BSD make will not use it). In mirage3 we generated a Makefile that was usable with both BSD and GNU make (I don't think we should do that, it easily becomes a nightmare to write/generate portable Makefiles)

@TheLortex  asked:

> Do you know why the generated makefile doesn't work with BSD make ? If there's a simple change to make it compatible we could integrate that. Otherwise yes the renaming sounds fine.

Now I took a look again:
- (a) `--no-print-directory` is a GNU make extension
```
 depends depend::
-       @$(MAKE) --no-print-directory lock
-       @$(MAKE) --no-print-directory install-switch
-       @$(MAKE) --no-print-directory pull
+       @$(MAKE) lock
+       @$(MAKE) install-switch
+       @$(MAKE) pull
```

- (b) `$(shell ...)` is a GNU make extension (also `abspath` is GNU make only):
```
+OCAML_VERSION != ocamlc --version
+
 $(MIRAGE_DIR)/$(UNIKERNEL_NAME).opam.locked: $(MIRAGE_DIR)/$(UNIKERNEL_NAME).opam
        @$(MAKE) -s repo-add
        @echo " ↳ generate lockfile for monorepo dependencies"
-       @$(OPAM) monorepo lock --require-cross-compile --build-only $(UNIKERNEL_NAME) -l $@ --ocaml-version $(shell ocamlc --version); (ret=$$?; $(MAKE) -s repo-rm && exit $$ret)
+       @$(OPAM) monorepo lock --require-cross-compile --build-only $(UNIKERNEL_NAME) -l $@ --ocaml-version $(OCAML_VERSION); (ret=$$?; $(MAKE) -s repo-rm && exit $$ret)
...
+ABS_BUILD_DIR != realpath $(BUILD_DIR)
+
 pull:: $(MIRAGE_DIR)/$(UNIKERNEL_NAME).opam.locked
        @echo " ↳ fetch monorepo rependencies in the duniverse folder"
-       @$(OPAM) monorepo pull -l $< -r $(abspath $(BUILD_DIR))
+       @$(OPAM) monorepo pull -l $< -r $(ABS_BUILD_DIR)
```

- (c) `$<` is a GNU make extension:
```
[...]

PR #1393, opened 2023-02-07
Rename the generated Makefile to GNUmakefile

hannesm:
This solves #1330, being honest the generated file only to be used by GNU make


TheLortex (approved):

TheLortex (commented):
CI fails, but I don't understand the reason why.

```
/src/: (run (shell "opam monorepo list -l mirage.opam.locked"))
==> Using lockfile mirage.opam.locked
opam-monorepo: [ERROR] mirage.opam.locked: No such file or directory
```

hannesm:
I'll close this PR. there's still an open issue about the Makefile / GNUmakefile thingy, and for some strange reasons this one fails... so let's get back to it once there's enough energy behind it.

PR #1394, opened 2023-02-07
Pass -f <config-file> in build rule of Makefile

### LINKED ISSUE #1356: Generated Makefile is incomplete when passing -f <dir/my-config.ml>
As @TheLortex mentioned in https://github.com/mirage/mirage-crypto/pull/166#issuecomment-1252038022:

> Currently you have to use -f mirage/config.ml everywhere, which I agree is not very practical.
>
> >    Also, in the generated Makefile (at top level) the build target omits the -f mirage/config.ml -- is this intentional?
>
> That's a bug indeed.

hannesm:
Fixes #1356

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version = 0.23.0
+version = 0.24.1
 profile = conventional
 break-infix = fit-or-vertical
 parse-docstrings = true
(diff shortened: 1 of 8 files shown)


TheLortex (approved):
Thanks for the PR. 

We should have an end-to-end test that makes use of the generated makefile.

PR #1399, opened 2023-02-16
adapt to new dns-client package layout

hannesm:


hannesm:
Hmm, somehow the "Mirage CI" seems to use a not up-to-date opam-repository (that do not know the dns-client-mirage package). @TheLortex would you mind to remind me how to trigger an update thereof / when are updates happening for it?

TheLortex:
This PR introduced the CI failure that we see on `master`. Looking at the solver output:
- `dns-client-mirage.7.0.0` is not selected: `dns-client-mirage.7.0.0: Requires dns-client = 7.0.0`
- `dns-client.7.0.0` is not selected: `conduit-mirage 6.1.0 requires >= 6.4.0 & < 7.0.0`
- `conduit-mirage.6.2.0` is unreleased: https://github.com/ocaml/opam-repository/pull/23344

hannesm:
Thanks @TheLortex for your explanation and the steps :)

TheLortex:
@hannesm CI is still red because the 3.7.0 release of dune broke chamelon, as the library has a dependency cycle. A PR with a fix is pending: https://github.com/yomimono/chamelon/pull/18

hannesm:
@TheLortex ok, great. Please correct me if I'm wrong, but an upper bound for dune on released versions of chamelon doesn't work since "dune" is executed in the opam switch context, while chamelon is in the monorepo context!?

To work around this, we could add a "dune < 3.7.0" build dependency to the example unikernel until chamelon get's a release? Or should we just cut a mirage release (with the CI failing)?

PR #1400, opened 2023-02-16
adapt to mirage-block-ccm 2.0.0 changes (maclen is removed due to mirage-crypto changes)

hannesm:


hannesm:
In the Mirage-CI I can see yet again a solver error of "opam-monorepo" - which seems to have selected "mirage-crypto-rng-mirage 0.11.0" and then got disappointed that other packages conflict with that (dns-client 6.x requires tls 0.15 requires mirage-crypto-rng-mirage < 0.11.0). Any insights how such a very well-shaped opam-repository leads the "opam-monorepo" to emit solver failures?

PR #1401, opened 2023-02-17
changes for 4.3.4

hannesm:


reynir (approved):

TheLortex:
CI success depends on `conduit.6.2.0` being released https://github.com/ocaml/opam-repository/pull/23344 (see https://github.com/mirage/mirage/pull/1399#issuecomment-1438190102)

hannesm:
ok, so the final commit is a bit questionable, as @TheLortex mentioned in https://github.com/mirage/mirage/pull/1399#issuecomment-1440311118 the underlying issue is that dune 3.7 broke compilation of chamelon, and the workflow of mirage doesn't take the chamelon upper bounds into consideration -- and the dune used is indeed 3.7 which breaks the build

now, my question is: should we merge and release? or should we remove the last commit (and thus no dune upper bound)? the latter would lead to potential users of chamelon in a mirage unikernel finding strange error messages... what do you think?

palainp:
In my opinion, we should keep the last commit to ensure that users can compile without issues, merge and release plus remove it when [chamelon#18](https://github.com/yomimono/chamelon/pull/18) is merged (and release another minor version).

Issue #1398, opened 2023-02-15
Test generated makefile


hannesm:
As noted in https://github.com/mirage/mirage/pull/1394 by @TheLortex 
We should have an end-to-end test that makes use of the generated makefile.

Issue #1386, opened 2023-01-09
drop "build" subcommand


hannesm:
Please comment if you disagree / have a use for `mirage build`. From my point of view, it adds quite some complexity to the mirage utility, suffers from no customizations, and thus would need to be extended with a huge chunk of options.

It is at the moment an alias to `"dune" % "build" % "--root" % "."` (AFAIU in lib/functoria/lib.ml). The downsides are:
- I can't pass `--profile=development` (see #1374)
- Does not work well with `-f` and other sources around (see #1371)
- Trouble with local sources (#1372)

One path is to drop the `mirage build` (and remove `profile release` from the dune-workspace), and update documentation to use `dune build [--profile=release] --root <whereever> dist/unikernel.<whatever>`. Also the generated Makefile and opam file(s) should be updated to use `dune build` directly.

Another path is to add more of `dune build` options to `mirage build`, at the cost of maintenance.

What do you think (esp. @samoht @TheLortex @dinosaure)?

PR #1404, opened 2023-03-02
Deprecate build

hannesm:
//cc @TheLortex -- maybe we should add #1398 before merging. But locally this seems to work nicely.


hannesm:
I rebased and removed the build subcommand entirely in a last commit. Since there's quite some other rework going on (and we'll have some breaking changes), let's consider this for inclusion. /cc @TheLortex @samoht, fixing #1386

hannesm:
and this now uses `make build`, thus the makefile bits are tested :)

samoht:
Thanks!

PR #1407, opened 2023-03-03
Disallow empty key names

reynir:
If you try to define a key with the name `""` in config.ml an error is raised if the key becomes relevant. However, the key may not always be relevant if it depends on e.g. `is_solo5`. This change pushes the check earlier so the error doesn't depend on conditionals.

```
couldn't open output channel for key_gen file: Invalid_argument("")
```

### DIFF
--- a/lib/functoria/key.ml
+++ b/lib/functoria/key.ml
@@ -486,6 +486,8 @@ let alias name a =
   make ~setters ~arg ~name
 
 let create name arg =
+  if name = "" then
+    invalid_arg "Key.create: key name cannot be the empty string";
   let setters = [] in
   make ~setters ~arg ~name
 


reynir:
A small reproduction:

```OCaml
open Mirage

let requires_solo5 =
  let err = "This unikernel configuration requires solo5" in
  let solo5 =
    Key.Arg.conv
      ~conv:((fun _ -> `Error err),
             (fun _ppf _ -> failwith err))
      ~serialize:(fun _ppf _ -> failwith err)
      ~runtime_conv:(Fmt.str "failwith %S" err)
  in
  let requires_solo5 =
    let doc = Key.Arg.info ~doc:"This unikernel configuration requires solo5" ["requires-solo5"] in
    Key.(v (create "" Arg.(required solo5 doc)))
  in
  impl ~keys:[ requires_solo5 ] "Unit" job

let () =
  register "requires_solo5"
    [
      if_impl Key.is_solo5
        noop
        requires_solo5 ;
    ]
```

```shell
$ mirage configure -t hvt
$ # no error
```

```shell
$ mirage configure 
couldn't open output channel for key_gen file: Invalid_argument("")
$
```

hannesm:
looks good to me. thanks a lot.

PR #1409, opened 2023-03-12
fix opam-monorepo issue with Ocaml5+

palainp:
Dear developers,

This PR should solve https://github.com/tarides/opam-monorepo/issues/340 and permits to compile mirage unikernels with Ocaml5+.

The underlying issue was the `ocaml` package marked as opam-provided and thus removed from dependencies at https://github.com/tarides/opam-monorepo/blob/9cc963ce7bc91592a301feca1e2238bc17448bd8/lib/opam_solve.ml#L114 (This can be seen inside the solver: https://github.com/ocaml-opam/opam-0install-solver/issues/47). Therefore, as https://github.com/ocaml/opam-repository/blob/master/packages/base-domains/base-domains.base/opam have a dependency list which contains `ocaml` OR `ocaml-variants`, leave the solver only with `ocaml-variants` if we remove `ocaml`.

For the time being I moved ocaml to the `Monorepo` scope, this is why it now appears as monorepo in the test files, but unsure if that the best move, comments are more than welcome!

### DIFF
--- a/lib/mirage/mirage.ml
+++ b/lib/mirage/mirage.ml
@@ -376,7 +376,7 @@ let run t = %s.Main.run t ; exit 0|ocaml}
       let min = "4.3.1" and max = "4.4.0" in
       let common =
         [
-          package ~scope:`Switch ~build:true ~min:"4.08.0" "ocaml";
+          package ~scope:`Monorepo ~build:true ~min:"4.08.0" "ocaml";
           package ~scope:`Monorepo "lwt";
           package ~scope:`Monorepo ~min ~max "mirage-runtime";
           package ~scope:`Switch ~build:true ~min ~max "mirage";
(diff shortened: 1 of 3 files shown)


hannesm:
Thanks for the investigations. What about removing `ocaml` dependency from the generated opam file completely (since e.g. mirage-runtime depends on it, we will have an implicit dependency on it anyways) -- since we use ocaml-solo5 for cross-compilation (when choosing a solo5 target), there's not really a need for the ocaml dependency anyways (and we don't quite need it in the monorepo).

What do others think? I'm not too knowledgeable about the opem-monorepo and 0install-solver stuff.

palainp:
Thanks for your reply @hannesm!
To me your last argument with the cross-compilation wins all (in the future cross-compilation may also be used for other targets)!

I'll add a commit to remove `ocaml` from the dependecies, so far I managed to compile q-m-f with success.

hannesm:
great, thanks

PR #1411, opened 2023-03-21
avoid bytes dependency from functoria-runtime

hannesm:


hannesm:
I don't quite understand the failure, the alternative is (as mentioned in https://github.com/ocaml/opam-repository/pull/23542#issuecomment-1476755532) to add an explicit dependency on ocamlfind.

reynir:
The error seems to be related to https://github.com/dinosaure/paf-le-chien/pull/83

```
File "duniverse/paf-le-chien/lib/alpn.ml", line 231, characters 37-41:
231 |         H2.Client_connection.request conn request ~error_handler
                                           ^^^^
Error: This expression has type unit -> H2.Client_connection.t
       but an expression was expected of type H2.Client_connection.t
       Hint: Did you forget to provide `()' as argument?
```

hannesm:
Thanks for your investigation @reynir.

Issue #1222, opened 2021-04-15
Mirage describe help layout is wrong 


TheLortex:
When reviewing the cram tests PR  (https://github.com/mirage/mirage/pull/1220), @dinosaure noticed that the output of `mirage describe --help=plain` has some issues:
```
DESCRIPTION
       The describe command describes the configuration of a mirage
       application.

       The dot output contains the following elements:
       If vertices
           Represented as circles. Branches are dotted, and the default
           branch is in bold.       Configurables
                                        Represented as rectangles. The order
                                        of the output arrows is the order of
                                        the functor arguments.
                                                              Data
                                                              dependencies

                                                              Represented as
                                                              dashed arrows.
       App vertices
```
This is something that don't happen in the normal output (`mirage describe --help`), it should be investigated.

michaelRichards99:
I am a student from the Virtualization class at the University of Texas at Austin, and participating open source development is part of the course requirement. We are willing to fix this if y'all want. Thanks!

reynir:
Fixed in cmdliner, and tests updated in #1418.

PR #1413, opened 2023-03-28
Fix the describe help layout.

michaelRichards99:
Addresses #1222 . There seems to be an issue within the `Cmdliner` library that was causing this malformed output.


reynir:
Thank you. Unfortunately the patch seems to mangle the output when the output format is `pager`:
Screenshot from 2023-03-28 10-24-25

michaelRichards99:
I posted an issue on their Github and then saw that you have a pull request in the works :). I guess this pull request can be closed since that fix will address this issue.

reynir:
It would have been nice with a workaround until that is merged. Thanks for putting in the effort!

Issue #1415, opened 2023-03-29
Makefile: opam install the lock file


reynir:
According to this comment we need to opam install the lockfile:

https://github.com/tarides/opam-monorepo/issues/378#issuecomment-1479335882

But in the code we install the .opam file (and not the .opam.locked file):

https://github.com/mirage/mirage/blob/main/lib/functoria/makefile.ml#L138-L140

CC @TheLortex @samoht

TheLortex:
Indeed we should install the `.opam.locked` file

reynir:
Reopening this as #1462 was reverted. It seems given the current state of opam-monorepo we should not opam install the lock file but also we should(?!) it would be good to have clarified the expectations of opam-monorepo and what the workflow should look like.

hannesm:
When installing the lockfile, we ended up in #1476 which adds lots of pins to the current opam switch. An option would be to `--ignore-pin-depends` -- but in the end, I'm more interested in which concrete issue "using the lockfile" would solve.

PR #1423, opened 2023-06-04
mirage: issues with dune 3.8.0

hannesm:
see  https://github.com/ocaml/dune/issues/7875

### DIFF
--- a/mirage.opam
+++ b/mirage.opam
@@ -19,7 +19,7 @@ build: [
 
 depends: [
   "ocaml" {>= "4.08.0"}
-  "dune" {>= "2.9.0"}
+  "dune" {>= "2.9.0" & < "3.8.0"}
   "dune" {with-test & >= "3.0.0"}
   "ipaddr" {>= "5.0.0"}
   "functoria" {= version}


hannesm:
ouch, so with opam-monorepo in the chain we're not able to restrict dune, is it what the CI run is supposed to tell me? https://github.com/tarides/opam-monorepo/issues/378

hannesm:
This also leads me to the question: why is dune in the duniverse? I mean from all I can tell, the dune in the opam switch is used for compilation!? If I downgrade in the opam switch to dune 3.7.1, and then run the opam-monorepo stuff, it locks to dune 3.8.0 in the duniverse, but the compilation succeeds... so, can we remove the dune from the `duniverse`?

samoht:
> why is dune in the universe? 

This is required to by `app-info` (as it uses `build-info`) to add metadata about the libraries linking with the unikernel.

hannesm:
> > why is dune in the universe?
> 
> This is required to by `app-info` (as it uses `build-info`) to add metadata about the libraries linking with the unikernel.

But that `app-info` is optional, no? I have it not enabled anywhere. Since opam-monorepo computes a complete solution (apart from the ocaml compiler?), it's clear that dune is part of that solution (since e.g. mirage-runtime depends on it) -- but if the ocaml compiler is special, shouldn't we put "dune" as well in the special parts (i.e. `x-opam-monorepo-opam-provided` -- but is that field read for computing the dependency cone?)

reynir:
I believe the below issues are relevant. It seems the opam-monorepo lockfile is assumed to be opam install'ed by opam-monorepo, but it is not opam install'ed by mirage.
https://github.com/tarides/opam-monorepo/issues/378
https://github.com/tarides/opam-monorepo/issues/378#issuecomment-1479335882
https://github.com/mirage/mirage/issues/1415

samoht:
I'm not sure I understand your remark. The mirage tool installs the switch dependencies here: https://github.com/mirage/mirage/blob/main/lib/functoria/makefile.ml#L140

Do you mean there's a missing `make switch-depends` missing somewhere?

samoht:
Ha yes good point!!

Leonidas-from-XIV:
Yes, I would say that what @reynir is saying makes sense, it needs to install the dependencies that are not marked as `?vendor` first, since opam-monorepo is not running dune, it is the dune from the environment that is being used.

PR #1425, opened 2023-06-05
Fix tests on macOS

samoht:


hannesm:
thanks, the macOS runners still have some issues (more `-t unix` needed?). and maybe remove the ` & os != "macos"` from mirage.opam (once fixed)!?

hannesm:
thanks, merging.

Issue #1426, opened 2023-06-08
Behavior of jobs without dependencies


reynir:
In the following code we have a unikernel with a job that does not depend on any devices, but logs a message.

`config.ml`:
```OCaml
open Mirage

let hello = main "Unikernel.Hello" job
let () = register "hello-nodev" [ hello ]
```
`unikernel.ml`:
```OCaml
module Hello = struct
  (* this is imediately evaluated *)
  let start =
    Logs.info (fun f -> f "hello");
    Lwt.return_unit
end
```
This successfully compiles and runs, but on the unix target this does not print anything to the console. My suspicion is the log reporter is not set up before `Unikernel.Hello.start` is evaluated and therefore nothing is printed.

Then my idea was to suspend the computation with a dummy/null/unit device. It was not very clear, and I had to consult the source code, but I found that I could use `Mirage.noop` in `config.ml`:
```OCaml
open Mirage

let hello = main "Unikernel.Hello" (job @-> job)
let () = register "hello-nodev" [ hello $ noop ]
```
And in `unikernel.ml`:
```OCaml
module Hello (_ : sig end) = struct
  let start () =
    Logs.info (fun f -> f "hello");
    Lwt.return_unit
end 
```
I guess for the first code we can't do much about the evaluation order(??) to ensure the log reporter is registered first.

My question then is what is a job? The documentation says a "main task" (but we can have a list of them!?). Should we have a unit typ and unit impl? Can we chain jobs like this?

This was discovered with @PizieDust.

palainp:
Hi @reynir and @PizieDust for finding this :)

I'm not sure if this is a plausible path to track this issue but looking in `caml_program`, calling `camlDune__exe__Unikernel__entry`, it seems that the `start` function is inline in the first version.

I then tried to un-inline it but, as it's no longer a function, the compiler doesn't let me to do that :/.

I'm now at the point where I've added a single unit to the `start` function when no args are provided (see https://github.com/mirage/mirage/compare/main...palainp:mirage:fix-nodev). This is a breaking change as now an unikernel without functorized device will have to take a unit as argument (but I'm not sure this was used before).
@mirage/core, any thoughts on this possible fix?

hannesm:
There's #873 which looks very similar to this issue. I agree with @palainp solution to add a unit argument to `start` functions without any dependencies.

reynir:
[...]

PR #1428, opened 2023-06-19
Fail if there are jobs registered without any arguments.

### LINKED ISSUE #873: Mirage_logs/Logs doesn't work at top-level
Suppose we have the following ```config.ml```:
```ocaml
open Mirage

let main = foreign "Unikernel" job

let () =
  let packages = [] in
  register "test" ~packages [main]
```
and the following ```unikernel.ml```:
```ocaml
let start =
  Logs.info (fun f -> f "Hello world !");
  Lwt.return_unit
```
If we execute it, the hello world is not displayed. This behaviour doesn't happen if ```start``` become a function as ```Mirage_logs``` will be initialized before the call to ```Logs.info```.

I wonder if something can be done in [mirage/functoria](https://github.com/mirage/functoria). Maybe there is a way to enforce or warn if we don't have a functor/function or something.

hannesm:
This ensures that toplevel expressions in the unikernels are properly evaluated.

Fixes #873 #1426

//cc @reynir

To test with: https://github.com/mirage/mirage-skeleton/pull/359

### DIFF
--- a/lib/functoria/impl.ml
+++ b/lib/functoria/impl.ml
@@ -45,7 +45,8 @@ let abstract t = Abstract t
 let app_has_no_arguments = function
   | App { args = Cons _ ; _ }
   | Dev { args = Cons _ ; _ } -> false
-  | _ -> false
+  | App _ | Dev _ -> true
+  | If _ -> false
 
 (* Devices *)
 
--- a/lib/functoria/impl.ml
+++ b/lib/functoria/impl.ml
@@ -43,15 +43,9 @@ and 'a device = ('a, abstract) Device.t
 let abstract t = Abstract t
 
 let app_has_no_arguments = function
-  | App { args ; _ } ->
-    (match args with
-     | Cons _ -> false
-     | _ -> true)
-  | Dev { args ; _ } ->
-    (match args with
-     | Cons _ -> false
-     | _ -> true)
-  | If _ -> false
+  | App { args = Cons _ ; _ }
+  | Dev { args = Cons _ ; _ } -> false
+  | _ -> false
 
 (* Devices *)
 
--- a/lib/functoria/impl.ml
+++ b/lib/functoria/impl.ml
@@ -42,6 +42,17 @@ and 'a device = ('a, abstract) Device.t
 
 let abstract t = Abstract t
 
+let app_has_no_arguments = function
+  | App { args ; _ } ->
+    (match args with
+     | Cons _ -> false
+     | _ -> true)
+  | Dev { args ; _ } ->
+    (match args with
+     | Cons _ -> false
+     | _ -> true)
+  | If _ -> false
+
 (* Devices *)
 
 let mk_dev ~args ~deps dev = Dev { dev; args; deps }
--- a/lib/mirage/mirage.ml
+++ b/lib/mirage/mirage.ml
@@ -485,6 +485,12 @@ let ( ++ ) acc x =
 
 let register ?(argv = default_argv) ?tracing ?(reporter = default_reporter ())
     ?keys:extra_keys ?packages ?src name jobs =
+  if List.exists Functoria.Impl.app_has_no_arguments jobs then
+    invalid_arg "Your configuration includes a job without arguments. \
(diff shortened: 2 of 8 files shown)


lib/mirage/mirage.ml
```diff
@@ -485,6 +485,12 @@ let ( ++ ) acc x =
 
 let register ?(argv = default_argv) ?tracing ?(reporter = default_reporter ())
     ?keys:extra_keys ?packages ?src name jobs =
+  if List.exists Functoria.Impl.app_has_no_arguments jobs then
+    invalid_arg "Your configuration includes a job without arguments. \
+                 Please add a dependency in your config.ml: use \
+                 `let main = Mirage.main \"Unikernel.hello\" (noop @-> job) \
```
palainp, line 491: Regarding the example later, it should be `let main = Mirage.main \"Unikernel.hello\" (job @-> job) \`, no?

test/mirage/job-no-device/configure.t
```diff
@@ -0,0 +1,6 @@
+  $ export MIRAGE_DEFAULT_TARGET=unix
+
+Configure
+  $ ./config.exe configure
+  Fatal error: exception Invalid_argument("Your configuration includes a job without arguments. Please add a dependency in your config.ml: use `let main = Mirage.main \"Unikernel.hello\" (noop @-> job) register \"hello\" [ main $ noop ]` instead of `.. job .. [ main ]`.")
```
palainp, line 5: Same here, `(job @-> job)`

lib/functoria/impl.ml
```diff
@@ -42,6 +42,17 @@ and 'a device = ('a, abstract) Device.t
...
 
+let app_has_no_arguments = function
+  | App { args ; _ } ->
+    (match args with
+     | Cons _ -> false
+     | _ -> true)
+  | Dev { args ; _ } ->
+    (match args with
+     | Cons _ -> false
+     | _ -> true)
```
palainp, line 53: Here there is no distinction between `App` and `Dev`. So we can simplify things by deleting lines 46 to 49. But I'm not sure an unikernel using a device without arguments won't also fail (or is supposed to?). I surely have to look deeper into it :)

palainp (commented):
Thanks @hannesm that sounds great to me! I'll try to look deeper regarding my first comment but the PR might get merged before I'm done :)

lib/mirage/mirage.ml
hannesm: indeed, thanks for spotting this, fixed in 28e80385

lib/functoria/impl.ml
hannesm: I tried and failed due to GADT and the OCaml compiler complaining about types escaping its scope. But feel free to experiment with that and find a good solution :)

reynir, line 53:
```suggestion
let app_has_no_arguments = function
  | App { args = Cons _ ; _ } | Dev { args = Cons _ ; _ } -> false
  | App _ | Dev _ -> true

```

hannesm: thanks, done in 85886a1a

hannesm: see 85886a1a (as proposed by @reynir)

lib/functoria/impl.ml
```diff
@@ -43,15 +43,9 @@ and 'a device = ('a, abstract) Device.t
...
-     | Cons _ -> false
-     | _ -> true)
-  | Dev { args ; _ } ->
-    (match args with
-     | Cons _ -> false
-     | _ -> true)
-  | If _ -> false
+  | App { args = Cons _ ; _ }
+  | Dev { args = Cons _ ; _ } -> false
+  | _ -> false
```
palainp, line 48: Should be true :)

hannesm: true that is, I adjusted this in [b9b9ddb](https://github.com/mirage/mirage/pull/1428/commits/b9b9ddbb80b00d5762e82d8b62e339577d1667c0)

hannesm:
Requires https://github.com/mirage/mirage/pull/1428

PR #1430, opened 2023-06-19
remove tracing: this has been unused for a long time, and is unsupported with xen using solo5; also ring_size from reporter

hannesm:
please let me know if you have a use case of this functionality


samoht:
LGTM

PR #1436, opened 2023-06-21
Remove Key.Alias

samoht:
This doesn't seem to be used anymore - it was a great initial feature to have initially to migrate from previous version of mirage, but it doesn't seem to be useful anymore. So let's remove it :-)

On top of #1435


hannesm:
this looks fine to me, thanks!

PR #1437, opened 2023-06-21
Simplify the management of configure and runtime keys

samoht:
To test with https://github.com/mirage/mirage-skeleton/pull/364

This PR simplifies the management of "keys" - what mirage used to describe parameters that can be passed a configure-time (to configure the deployment target) and runtime (to configure configuration options).

This PRs does a few things to simplify the current state:
- first it remove mandatory dependency to `ipaddr` for unikernels that do not need to link with that library
- then it removes the cmdliner half-reimplementation used to describe runtime keys and allow users to use Cmdliner terms directly
- then it allows users to define runtime keys in the unikernel itself (by calling `Mirage_runtime.Key.register <cmdliner term>`)

### DIFF
--- a/lib/functoria/DSL.ml
+++ b/lib/functoria/DSL.ml
@@ -46,9 +46,9 @@ let impl ?packages ?packages_v ?install ?install_v ?keys ?extra_deps ?connect
   @@ Device.v ?packages ?packages_v ?install ?install_v ?keys ?extra_deps
        ?connect ?dune ?configure ?files module_name module_type
 
-let main ?packages ?packages_v ?keys ?extra_deps module_name ty =
+let main ?packages ?packages_v ?extra_deps module_name ty =
   let connect _ = Device.start in
-  impl ?packages ?packages_v ?keys ?extra_deps ~connect module_name ty
+  impl ?packages ?packages_v ?extra_deps ~connect module_name ty
 
-let foreign ?packages ?packages_v ?keys ?deps module_name ty =
-  main ?packages ?packages_v ?keys ?extra_deps:deps module_name ty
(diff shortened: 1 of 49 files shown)


samoht:
I'm starting to split this into smaller PRs: https://github.com/mirage/mirage/pull/1440

hannesm:
Thanks @samoht for your work. I've a hard time to review step-wise changes in this area.

Maybe it makes sense to keep the steps on this PR/branch, and merge them once we're completely happy with the result?

Since I don't understand the internals (neither this change) too well, I also trust you to say "this is a good and finished PR, let's merge, it'll be independent of hannes' request a good idea to move into that direction".

samoht:
Sounds good. Let me try to merge some of the independent cleanups first, and I'll keep iterating on the core internal changes in that PR. I'll also update the description to ease the review.

samoht:
@hannesm : what do you think of https://github.com/mirage/mirage/pull/1437/commits/bd9eebfcae5134545ee3989835000108cbd35d93 ?

Would it be a first set of changes that is going in the direction that you want?
- no more hard-requirement with ipaddr for mirage-runtime
- ability to define any keys with cmdliner at runtime
- (a simpler key_gen.ml that use Cmdliner instead of a subset of it redefined by mirage_runtime)

reynir:
Thanks for this.

I notice the default stage is `` `Configure``. In https://github.com/mirage/mirage-skeleton/pull/364/files 11 of 17 calls to `Mirage.Key.create` set ``~stage:`Run``. I think we should consider making `~stage` a required argument - it's a breaking change and users will need to consider the stage of their keys - and maybe `` `Configure`` is not an obvious default.

Issue #1422, opened 2023-06-02
question about configuration/runtime keys and "staging" thereof


hannesm:
So, I understand there's this approach that you have to define boot parameters in `config.ml`, and also configuration-time parameters.

I also understand that both kinds of keys are useful.

What I do not understand is: what is the usage of keys that are both available (can be filled) at configuration and runtime? E.g. the IP address of a network interface, there we have now 3 times when to define them:
- at development time in the key,
- at `mirage configure --ipv4=` time,
- at the startup of the unikernel.

I also understand from an academic point of view this may be an interesting property.

But from a practical point of view, I suggest to disentangle the keys that are used at configuration time (i.e. the target, and selection of features that require other packages), and those at runtime (e.g. the IP address).

Why do I propose this? Well, first of all, I went through all the unikernels I could find to figure out where exactly this "put another value at configuration time" is used and useful -- apart from the hello world example I couldn't find any. Another reason is that there is heavy machinery in functoria/functoria-runtime/mirage/mirage-runtime to deal with serialising and deserialising keys, which includes depending on ipaddr in mirage-runtime; but also all these poor unikernels that have the keys only available as base types (string / int -- but somehow Ipaddr is special enough that it got promoted as a hard dependency). Now, stuff like a Private_key.t is not there in mirage-runtime (and should not be).

So, codewise I don't quite understand functoria enough to deal with this, but my first question is: are there any real use-cases for having staged keys? The second question is: can we disentangle these keys (and (a) remove key_gen.ml, (b) the whole "defined stuff in config.ml which belongs to unikernel.ml", and (c) even the fork & modification of the cmdliner API, (d) unclutter the manpage)? What do you think?

My suspicion is that we do not even need to break the boot parameters (i.e. any existing unikernel), but we'd be able to remove a bunch of code (and the ipaddr dependency from mirage-runtime).

//cc @mirage-core @Drup

hannesm:
Thanks for your comment.

So, what would be the benefit to customize your IP address based on your target? I don't quite understand this use case. I think it is fine to "configure" with a "dune stanza", as described above, but I also think that "what can be configured" is the target, and setups for adding further packages as dependencies (i.e. enable-monitoring / tls, as in mirage-www).

What is "the equivalent of `unikernel.conf`"? Who should read and apply such a file? For e.g. solo5-hvt/spt I can see how such a thing could be done, for Xen and virtio and muen I don't see a path.

[...]

PR #1442, opened 2023-06-23
Remove key args and both

samoht:
To test with: https://github.com/mirage/mirage-skeleton/pull/364

Based on top of #1437 - experiment with removing "both" keys (just keep runtime and configure-time keys). Also remove the `~key` parameter when registering unikernel to force people to move these into the unikenerls. 

Should address most of what was requested in #1422 - but it's a bit more radical than #1437 so opening a separate PR to discuss.


hannesm:
To me, this looks like the right direction on how to make unikernel development easier to understand -- the keys are now defined in unikernel.ml, and can be accessed there directly. Should the return value from `register` (of a key) be a `lazy 'a` instead of `unit -> 'a`?

This also allows in unikernels that use other types than the base ones to specify converters for them (i.e. Ipaddr / Domain_name / X509.Key_type.t / ...). That's just great, I've been looking at that for a decade ;)

IIUC, Key_gen is still used, mainly for these keys for the network stack etc. (which are barely used in the unikernel, but the network stack uses it internally to configure itself).

Did you try to port a more complex unikernel, such as mirage-www or unipi, to this branch? That'd likely be a good test on how this pans out.

samoht:
This has been merged as #1451, #1450, #1449, #1448, #1447 and #1437 (the main one).

PR #1444, opened 2023-06-27
Upgrade to ocamlformat 0.26.0

Julow:
The aim of this commit is to gather feedback.

Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Upgrade to OCamlformat 0.26.0
+d2cf205b07b9154bec2a6b490a348457d14fec09
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version = 0.25.1
+version = 0.26.0
 profile = conventional
 break-infix = fit-or-vertical
 parse-docstrings = true
--- a/lib/functoria/install.mli
+++ b/lib/functoria/install.mli
@@ -19,7 +19,7 @@
 type t
 
 val v : ?bin:(Fpath.t * Fpath.t) list -> ?etc:Fpath.t list -> unit -> t
-(** [v ~bin:\[(src,dst),...\] ~etc ()] is the installation of [src] as [dst] as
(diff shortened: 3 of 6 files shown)


lib/functoria/key.mli
```diff
@@ -182,7 +182,7 @@ val v : 'a key -> t
 (** [v k] is the [k] with its type hidden. *)
 
 val abstract : 'a key -> t
-  [@@ocaml.deprecated "Use Functoria.Key.v."]
```
samoht, line 185: @Julow Is this expected? The annotation is supposed to be attached to the top-level item, right?

Julow: The attribute is unindented but still attached to the same item, yes.

samoht: what's the rationale for changing this?

Julow: This is changed here: https://github.com/ocaml-ppx/ocamlformat/pull/2317

It was initially implemented for the Janestreet profile and I thought less indentation is generally better so I changed the default profile too. Do you prefer the previous formatting ?

samoht: Ha ok, that matches ocp-indent behavior then? If yes, I'm fine with the changes, that's a good enough rational.

Julow: It does. The main reason was that less indentation is generally better and this is useless indentation.

samoht: Was it reverted in the new version you just pushed?

Julow: No, the attributes have been removed in cf46b243318265024c88ceb86dc47ca16c3087cc and I rebased this PR on top.

Julow:
OCamlformat 0.26.0 has been released! This PR can be merged (not squashed due to `.git-blame-ignore-revs`) or closed if it's a bit too early to upgrade.

samoht (approved):

samoht:
Thanks, I'll merge once `ocaml-ci` is green (hopefully soon)

PR #1449, opened 2023-06-29
refactor: remove some complexity about handling of required keys

samoht:
In practice none is using those required keys anyway.


samoht:
The failure in the CI should be addressed by https://github.com/mirage/mirage-skeleton/pull/371

PR #1453, opened 2023-06-30
keys: make ~stage a required argument

reynir:
There doesn't seem to be a clear preference for `` `Configure``, and since it is a breaking change it may be desirable to force users to consider the stage of the keys.

See the discussion in https://github.com/mirage/mirage/pull/1437#issuecomment-1614378015

### DIFF
--- a/lib/functoria/key.ml
+++ b/lib/functoria/key.ml
@@ -191,15 +191,15 @@ module Arg = struct
 
   let stage t = t.stage
 
-  let opt ?(stage = `Configure) conv default info =
+  let opt ~stage conv default info =
     { stage; info; kind = Opt (default, conv) }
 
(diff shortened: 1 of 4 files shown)


samoht:
In terms of breaking changes, I think I'd prefer https://github.com/mirage/mirage/pull/1455 -- ie. actually only keep our weird `Key.Arg` fork for configure-time keys.

(I have a [branch](https://github.com/mirage/mirage/compare/main...samoht:mirage:remove-cmdliner-fork?expand=1) that removes the rest of Cmdliner's fork but it doesn't work fully yet as I can't manage to make https://github.com/mirage/mirage/pull/1454 work properly for `mirage describe` -- not totally sure what is happening there, this code is still full of black magic)

reynir:
I have not looked at the PRs, but that sounds like a good way forward (sorry for the late reply; I was on vacation)

PR #1454, opened 2023-06-30
WIP: Simplify context file

samoht:
This is a (potentially breaking) user-facing change. But I do think it's necessary.

This PR makes `--config-file` always have priority over the CLI arguments. I find the current behaviour (where we merge keys on both the file and the cached file) a bit confusing. We also warn the user if something weird is happening (like passing different options on the CLI and in the config file).

The way it is handled currently is all of this was very clever but super confusing :-)

Also: Cmdliner does not distinguish between default and lack of arguments. So the tricks we were playing won't work if we remove our partial custom implementation of Cmdliner to use the upstream version. /cc @dbuenzli -- as maybe there's a way to use something like `env` to read arg values from a file instead from the environment. I did try to overwrite the `env` parameter that is passed to `Cmd.eval`, but this feels like a hack, and I don't have an easy process to map env variables and parameter names.

### DIFF
--- a/lib/functoria/context.ml
+++ b/lib/functoria/context.ml
@@ -37,6 +37,14 @@ let empty = Map.empty
 let add k v (t : t) : t = Map.add k.name (k.put v) t
 let mem k (t : t) = Map.mem k.name t
 
+let diff ~base x =
+  Map.fold
+    (fun k v acc ->
+      match Map.find_opt k x with
+      | None -> acc
+      | Some v' -> if v = v' then acc else k :: acc)
+    base []
+
 let find k (t : t) =
   if Map.mem k.name t then Some (k.get @@ Map.find k.name t) else None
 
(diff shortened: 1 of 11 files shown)


dbuenzli:
> Also: Cmdliner does not distinguish between default and lack of arguments.

You can use option values via `Arg.some'` and define your default/read from a file once you get the `None` in your lifted function. 

A default value can be documented via the `none` argument of `Arg.some'` or if there's a more complex procedure via the `absent` argument of `Arg.info`.

samoht:
> You can use option values via `Arg.some' `

Unfortunately, I don't have control over those, as the API takes `'a Cmdliner.Term.t` arguments :-/

samoht:
@Drup if you have any idea how to make `describe` work (without forcing everyone to use `Arg.some`), I would welcome your insights :-) 

In [78fda1d] (https://github.com/mirage/mirage/pull/1454/commits/78fda1ddf4a7432ca28b7445f477ff6d8c6b2566), `./config.exe describe --no-eval` is still evaluating all the keys to their default values and so do not seem to show the full graph. Any ideas?

samoht:
Edit: it seems that `describe --no-eval` only works for the graph but not for the actual text description (as I guess the info value is fully reduced anyway). So it seems to work ok on top of #1456

samoht:
I'm closing this as I don't think it's the right approach anymore.

PR #1455, opened 2023-06-30
Keep a single way to define runtime keys (using Cmdliner)

### LINKED ISSUE #1422: question about configuration/runtime keys and "staging" thereof

samoht:
To be tested with https://github.com/mirage/mirage-skeleton/pull/374

This PR moves all the definitions of runtime keys to separate runtime modules. This is a breaking change and might be annoying for users, but it also makes the separation between configure and runtime keys much cleaner (and allows removing a lot of code). 

Fix #1422


lib/mirage/impl/mirage_impl_block.ml
```diff
@@ -188,19 +189,20 @@ let docteur_unix (mode : mode) extra_deps disk branch analyze remote =
       {ocaml|let ( <.> ) f g = fun x -> f (g x) in
              let f = Rresult.R.(failwith_error_msg <.> reword_error (msgf "%%a" %s.pp_error)) in
              Lwt.map f (%s.connect ~analyze:%a %a)|ocaml}
-      modname modname Key.serialize_call (Key.v analyze) Key.serialize_call
-      (Key.v disk)
+      modname modname Runtime_key.call analyze Runtime_key.call disk_r
```
samoht, line 193: /cc @dinosaure: I am not sure what docteur does here - the disk key seems to be used at runtime to pick the name of the disk? It might makes sense on unix, but what's the semantics on solo5?

Also: I've added 2 disk parameters (one for configure-time, one for runtime) but now only the configure-time one is tweakable by the users for now. It's a good idea to separate those anyway as the location of the disk is unlikely to be similar on the machine that builds the unikernel and the machine that deploys it.

What do you think? Are you ok with two different arguments here? And what should be the default value on solo5?

lib/mirage/impl/mirage_impl_block.ml
```diff
@@ -260,17 +262,13 @@ let disk =
...
-  let doc =
-    Key.Arg.info ~doc:"Analyze at the boot time the given docteur disk."
-      [ "analyze" ]
-  in
-  Key.(create "analyze" Arg.(opt ~stage:`Run bool true doc))
+let analyze = Runtime_key.create "Mirage_runtime.analyze"
 
 let docteur ?(mode = `Fast) ?(disk = disk) ?(analyze = analyze) ?branch
     ?(extra_deps = []) remote =
+  let disk = (disk, Runtime_key.create "Mirage_runtime.disk") in
```
samoht, line 273: TODO: `Mirage_runtime.disk` doesn't exist, we need to decide what we want to do with these runtime `disk` parameters.

samoht:
All green! 🎉 

It's a significant breaking API change, so I welcome some review and questions before merging. There are potentially more cleanups to be done, but it's not possible right now to remove the thin wrapper in `Functoria.Key` around Cmdliner (or at least I haven't found how to do this).

I do not plan to break more things now, so I think the current design is a good compromise. I'll prepare a proper changelog, but the gist of it is that `keys` are not split into configure-time keys `Key.t` and runtime keys `Runtime_keyt.t` - these are now stored and represented separately. 
 
- Runtime keys are values of type `a Cmdliner.Term.t`  to be defined in some code linked to the unikernel. They store the path to this value as a string (+ some other metadata, like which package to use). They are typed, but nothing ensures that the type (or the path) is correct. MetaOCaml could help here.
- Configure-time keys are defined with a thin wrapper on top of an `a Cmdliner.Arg.t` term. The wrapper is used to build terms that can distinguish between default and absent keys.

samoht:
I've started to move some of the runtime keys to a separate package: https://github.com/mirage/cmdliner-stdlib ; I guess we can do something similar for our logs parameters. There's various tools where I'd like to re-use the same terms.

samoht:
And the log-handling code can also be moved elsewhere (to remove the `logs` dependency from `mirage-runtime`). For instance I've put it there: https://github.com/mirage/mirage-logs/pull/20

hannesm:
> @hannesm - do you want to review this patch, or should I merge it?

I am briefly back at my computer, and am happy to see the state of this PR :D I plan to test this branch with some of my unikernels, and will report back here.

> The only design decision I am unsure of is removing (read-only) configure keys from `Key_gen`. With the current state of the patch, it's pretty easy to keep those, and this might help with the `docteur` device, which uses the `disk` argument both at configure time and runtime. As I am not sure what was the intended semantics, maybe it's best to keep the previous behaviour.

I have no deep insight into docteur. I'm fine to keep Key_gen.ml around, for configure-time keys.

For this PR (and the previous ones), I still wonder whether you've tried to port e.g. mirage-www to the new boot args system. The value of experimenting with real unikernels (such as mirage-www) is that they may be a bit more subtle than our mirage-skeleton ones.

lib/mirage/mirage_key.mli
```diff
@@ -14,23 +14,19 @@
...
-  val ipv4_address : Ipaddr.V4.t converter
-  val ipv4 : Ipaddr.V4.Prefix.t converter
-  val ipv6_address : Ipaddr.V6.t converter
-  val ipv6 : Ipaddr.V6.Prefix.t converter
-  val ip_address : Ipaddr.t converter
+  val ipv4_address : Ipaddr.V4.t Arg.conv
+  val ipv4 : Ipaddr.V4.Prefix.t Arg.conv
+  val ipv6_address : Ipaddr.V6.t Arg.conv
+  val ipv6 : Ipaddr.V6.Prefix.t Arg.conv
+  val ip_address : Ipaddr.t Arg.conv
```
hannesm, line 26: I wonder, are these "configure-time keys" with ip addresses being used anywhere?

test/functoria/e2e/build.t
```diff
@@ -12,6 +12,7 @@ Build an application.
   dune
   dune.build
   dune.config
+  key.ml
```
hannesm, line 15: would you mind to elaborate what the newly generated `key.ml` is for? AFAICT it is "if you define runtime keys in config.ml", is this true? Shouldn't these keys be defined in unikernel.ml instead? My gut feeling is that the complexity increases if you generate more code, so I'd be eager to see less code being generated, and more being spelled out.

hannesm:
For an upcoming release of mirage, I'd appreciate if we can settle on this PR: merge or not merge (it also needs a rebase) -- since there's already some API changes for mirage along the way, I think it would not hurt to break the API more.

OTOH, I have some remarks / questions above, which I'd appreciate to have answers before we merge this :)

lib/mirage/mirage_key.mli
samoht: I've removed those

test/functoria/e2e/build.t
samoht: I've moved the runtime keys in app.ml

samoht:
@hannesm I've rebased and addressed your comments - the only missing thing is that I am pretty sure I have broken `docteur`, but I am not sure what is the expected behaviour. Where do the runtime keys are defined? Ideally it should be in the `docteur` package?

hannesm:
Thanks @samoht -- this looks very mature now. Would you mind to formulate a changes entry for this PR and the related ones about boot parameters? Also, from your point of view, once this PR is merged, would a release be good? Should https://github.com/mirage/cmdliner-stdlib be integrated (to reduce the code in mirage itself) -- maybe as a separate PR after this one?

samoht:
I've added a CHANGE entry. Is it clear enough? I'll look at the other changes I've made as well. And yes, making a new release makes sense :-)

CHANGES.md
```diff
@@ -1,3 +1,58 @@
...
+    ```ocaml
+    let key =
+      let doc = Key.Arg.info ~doc:"A Key." ~stage:`Run [ "key" ] in
+      Key.(create "key" Arg.(opt_all ~stage:`Run string doc))
+    ```
+    becomes:
+    ```ocaml
+    let key = Runtime_key.create "Unikernel.key"
+    ```
+    and `unikenel.ml` now contains:
```
hannesm, line 47:
```suggestion
    and `unikernel.ml` now contains:
```

CHANGES.md
```diff
@@ -1,3 +1,58 @@
...
+    ```ocaml
+    open Cmdliner
+
+    let key =
+      let doc = Arg.info ~doc:"A Key." [ "key" ] in
+      Arg.(value & opt_all string [] doc)
+    ```
```
hannesm, line 54: can we now as well define runtime keys purely in unikernel.ml (or in any library that depends on mirage-runtime)? or is there a requirement to have runtime keys of a unikernel mentioned in config.ml (if yes, why)?

samoht: Ha yes you are right I will clarify this. It's not a requirement to define the runtime key in `config.ml` - you do this only if you want to refer it in some new device initialisation code. I'll clarify.

samoht:
I've force-pushed a new version of the CHANGELOG. Does it look better?

/cc @mirage/core as well, as this is a (somehow controversial) breaking change.

lib/mirage/impl/mirage_impl_block.ml
dinosaure: I will try to check what happens next week (please, give me the time to see if we should rethink about how we handle block device, I was not very satisfied by what I did before).

samoht: no hurry - that's a big change; let me know when you are happy with it.

TBH, I'm not 100% sure about exposing the value of the configure-time key to the runtime (as a read-only value) or not. We do not do this anymore with this patch, but it wouldn't be too hard to re-add this, and maybe that'll make the life of `docteur` easier.

samoht:
Rebased to change `delay` into runtime keys.

test/mirage/describe/run.t
```diff
@@ -206,34 +148,20 @@ Describe before configure (no eval)
                 45 -> 35 [style="dashed"];
                 45 -> 33 [style="dashed"];
                 }
+<<<<<<< HEAD
+=======
+>>>>>>> f17e6c1e0 (Add Key.runtime to represent runtime keys)
+<<<<<<< HEAD
+=======
+>>>>>>> f17e6c1e0 (Add Key.runtime to represent runtime keys)
```
hannesm, line 156: these 6 lines above do not look correct.

samoht: woops

lib_runtime/mirage/mirage_runtime.ml
```diff
@@ -50,18 +56,178 @@ module Arg = struct
...
+let allocation_policy =
+  let doc =
+    "The policy used for allocating in the OCaml heap. Possible values are: \
+     $(i,next-fit), $(i,first-fit), $(i,best-fit). Best-fit is only supported \
+     since OCaml 4.10."
+  in
+  let doc =
+    Arg.info ~docs:ocaml_section ~docv:"ALLOCATION" ~doc [ "allocation-policy" ]
+  in
+  Arg.(value & opt Conv.allocation_policy `Next_fit doc)
```
hannesm, line 94:
```suggestion
  Arg.(value & opt Conv.allocation_policy `Best_fit doc)
```

hannesm: as noted in #1473

hannesm:
sorry that there's again a conflict due to my merge of #1473...

samoht:
Rebased

samoht:
Looking at this again after a while: @dinosaure did you have time to think about the `docteur` integration? Could this be done after we merge this PR?

hannesm:
from my point of view, indeed we can merge this PR without having a look at docteur. what we should do before the next mirage release is try these breaking changes on existing real unikernels (i.e. mirage-www etc.).

dinosaure:
Rebased and I updated the `docteur` part to clarify a bit what is its usage of `{runtime_,}keys`. For me, at least, on this specific part, it's good enough.

PR #1462, opened 2023-07-06
use the .opam.locked file for installing the switch dependencies

### LINKED ISSUE #1415: Makefile: opam install the lock file

hannesm:
fixes #1415


samoht:
Thanks!

PR #1463, opened 2023-08-03
ignore the format-only (and ocamlformat upgrade) revs we accumulated

hannesm:
proudly gathered by manual wading through the commit history, searching for "camlformat", "dune build @fmt", "Format". Some commits that have formatting changes and other changes are not included here (to avoid git blame being confused).


hannesm:
special thanks to @Julow for introducing this file, and making me read up the documentation about its purpose

For others who are curious, take a look at e.g. https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

hannesm:
since this only applies to github (and configured git clients), but doesn't modify the build, no need to wait for CI.

PR #1464, opened 2023-08-03
use \033 instead of \e for ansi escape sequences

hannesm:
as discussed in https://stackoverflow.com/questions/54039983/print-bold-or-colored-text-from-makefile-rule


hannesm:
if someone with a linux machine and/or macos could confirm that this works fine for them, please report here. for me, on FreeBSD this avoids printing \e[2m, and instead prints the text in gray.

hannesm:
tested on a linux VM, works great!

Issue #1381, opened 2022-12-13
ability to specify minimal mirage version in configure


hannesm:
as proposed / asked by @palainp. it would be great to be able to specify the minimal mirage version in configure that is supported. ideally, such a minimal version indication would lead to a nice error message "your mirage version is too old, please upgrade to at least x.y.z" (instead of e.g. "unknown binding <the-latest-mirage-device-exported-from-mirage>").

I don't quite understand how functoria / mirage are working these days, and whether such a feature could be implemented. as far as I can see, we can specify version bounds for compiling the unikernel, but none for the stuff used in config.ml (which may only be mirage itself anyways).

this was suggested in a recent mirage meeting, preserving it here as issue so it won't be forgotten.

hannesm:
related is https://github.com/mirage/mirage-skeleton/pull/338 and @TheLortex mentioned this in the most recent MirageOS call as well.

hannesm:
done in #1466

PR #1466, opened 2023-08-06
Allow a mirage version in config.ml

hannesm:
Please take a look at the second commit:

check in mirage configure for a version lower bound in config.ml
    
If the first line of config.ml contains:
(* mirage >= a.b.c *) -- an OCaml comment
    
Then, a.b.c.d is checked to be smaller or equal the current version. This allows
within configure to error out early with a reasonable error message if a too old
mirage is attempted to be used with a unikernel.
    
Addresses #1381

I'm pretty sure this can be "optimized", but tbh I waded through the "action.ml" and was unable to figure out the purpose and value (instead of directly calling Bos). I hope someone can shed some light into why it is there.

Of course I'm happy for review and feedback on this PR.

//cc @samoht @TheLortex

~~EDIT: it is completely unclear to me why the mirage-CI fails with the "job without dependencies" exception.~~ (thanks to @samoht I removed the mirage-skeleton PR ref)

### DIFF
--- a/lib/functoria/tool.ml
+++ b/lib/functoria/tool.ml
@@ -32,76 +32,86 @@ end
 
 let check_version ~name ~version data =
   let ( let* ) = Result.bind in
-  let first_str = "(* " ^ name ^ " " in
-  let fl = String.length first_str in
-  if fl < String.length data && String.equal (String.sub data 0 fl) first_str
-  then
-    let* lower_version, upper_version =
-      let vs =
-        String.split_on_char ' ' (String.sub data fl (String.length data - fl))
+  let extract_version v =
+    try Ok (Scanf.sscanf v "%u.%u.%u" (fun ma mi pa -> (ma, mi, pa))) with
+    | Scanf.Scan_failure _ | End_of_file -> (
+        try Ok (Scanf.sscanf v "%u.%u" (fun ma mi -> (ma, mi, 0))) with
+        | Scanf.Scan_failure _ | End_of_file -> (
+            try Ok (Scanf.sscanf v "%u" (fun ma -> (ma, 0, 0)))
+            with Scanf.Scan_failure _ | Failure _ | End_of_file ->
+              Error ("couldn't extract version (%u) from " ^ v))
+        | Failure f ->
+            Error ("couldn't extract version (%u.%u) from " ^ v ^ ": " ^ f))
+    | Failure f ->
+        Error ("couldn't extract version (%u.%u.%u) from " ^ v ^ ": " ^ f)
+  in
+  if String.equal version ("%%" ^ "VERSION%%") then (
+    Log.info (fun m ->
+        m "Skipping version check, since our_version is not watermarked");
+    Ok ())
+  else
+    let* version' = extract_version version in
+    let first_str = "(* " ^ name ^ " " in
+    let fl = String.length first_str in
+    if fl < String.length data && String.equal (String.sub data 0 fl) first_str
+    then
+      let* lower_version, upper_version =
+        let vs =
+          String.split_on_char ' '
+            (String.sub data fl (String.length data - fl))
+        in
+        let rec go lower upper = function
+          | "&" :: tl -> go lower upper tl
+          | ">=" :: v :: tl ->
+              if lower = None then go (Some v) upper tl
+              else Error "Bad comment, multiple >= constraints"
+          | "<" :: v :: tl ->
+              if upper = None then go lower (Some v) tl
+              else Error "Bad comment, multiple < constraints"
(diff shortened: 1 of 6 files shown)


hannesm:
This is IMHO a quality of life improvement for the future of the mirage utility, similarly to when it outputs code that initializes functors and emits code calling initialization functions, where we embed upper and lower bounds of the opam packages (since a while).

I just added support to be able to specify: `(* mirage >= 1.2.3 && < 3.4.5 *)` in the config.ml. Please let me know what you think about that /cc @mirage/core

palainp:
I totally approve this feature! Thanks!
The proposed solution seems fine to me :), but I'm unable to test in a short term :(

samoht:
> I waded through the "action.ml" and was unable to figure out the purpose and value

`Action` is useful to mock up Bos action and thus unit test all of this logic without doing any filesystem effects.

hannesm:
> > I waded through the "action.ml" and was unable to figure out the purpose and value
> 
> `Action` is useful to mock up Bos action and thus unit test all of this logic without doing any filesystem effects.

Great to hear, thanks for the insight.

samoht:
Also: Mirage CI is pinning `mirage-skeleton` to #338 as it is mentioned in the PR body. We probably need to make the "friend PR" link a bit weaker to avoid this in the future.

lib/functoria/tool.ml
```diff
@@ -197,17 +197,120 @@ module Make (P : S) = struct
     rm_gen_files ()
 
   (* App builder configuration *)
-  let with_alias ~save_args args ~depext:_ ~extra_repo:_ ?ppf ?err_ppf argv =
+  let configure ({ args; _ } : _ Cli.configure_args) ?ppf ?err_ppf argv =
+    let* () =
+      match
+        Bos.(
+          OS.Cmd.out_string
```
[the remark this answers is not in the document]
hannesm: done as you suggested in [dbe2e00](https://github.com/mirage/mirage/pull/1466/commits/dbe2e00aad4e9075ae03dc9b7aa5d623fbf07e66)

hannesm:
> Also: Mirage CI is pinning `mirage-skeleton` to #338 as it is mentioned in the PR body. We probably need to make the "friend PR" link a bit weaker to avoid this in the future.

ah, ok. I removed the mention of that PR from the initial comment now.

hannesm:
testing is a bit tricky in a non-substituted mirage git repo. I did manual testing by adapting the hardcoded strings.

lib/functoria/tool.ml
```diff
@@ -197,17 +197,104 @@ module Make (P : S) = struct
...
+      if not is_file then
+        Action.errorf "configuration file %a missing" Fpath.pp file
+      else Action.ok ()
+    in
+    let* () =
+      let* data =
+        let cmd = Bos.Cmd.(v "head" % "-1" % p file) in
+        Action.run_cmd_out ~err:`Null cmd
+      in
+      let pkg = "%%NAME%%" in
```
samoht, line 213: You can use `P.name` here.

lib/functoria/tool.ml
```diff
@@ -197,17 +197,104 @@ module Make (P : S) = struct
...
+      let first_str = "(* " ^ pkg in
```
samoht, line 214: Do you mind extracting this function (and exposing it somewhere) in the API and adding a few unit tests? I understand testing is annoying because of the `VERSION_NUM` substitution, but I can see becoming it hard to maintain without at least a few simple tests.

samoht: (we should probably have a way to inject `VERSION_NUM` in our tests, but that is more complicated that just unit-testing this function and can definitely come in a separate PR if their is a need for better tests later)

lib/functoria/tool.ml
hannesm: thanks, I also used P.version below.

lib/functoria/tool.ml
hannesm: that's all now done in [9b667fe](https://github.com/mirage/mirage/pull/1466/commits/9b667fe62f1ae54fc1b0aa6ea32c886f39f049a4) and [cfbed95](https://github.com/mirage/mirage/pull/1466/commits/cfbed95d98af62d1f6f542da51d12c60b32bea60)

Your feedback would be valuable (or a merge :)

hannesm:
CI issue is some cohttp change marking stuff deprecated, which for some reason is an error -- leading to static-website-tls not compiling nicely.

samoht:
LGTM

samoht (approved):

PR #1469, opened 2023-09-21
minor fixes

hannesm:
The first commit adds a dependency -- otherwise for the common `unix` target, a `make build` doesn't work anymore (don't know how to build hello). The second is purely stylistic for the formatter (and was broken in f33f94100).


samoht:
LGTM

Issue #1374, opened 2022-11-11
build system and warnings


hannesm:
With mirage 4, I fail to find a `warn-error` option (this has been removed in 76db9815542270b154eef8db8f33e80da9388276 by @samoht). Now the question is, since `(profile release)` is unconditionally emitted into `dune-workspace`; and there's ` (flags (:standard -warn-error -A))` unconditionally put into `dune.config`, how can I see warnings and errors?

I recently manually removed both profile and the flags in qubes firewall, and that exposed some unused opens and unused functions, which I never saw a warning about before (see https://github.com/mirage/qubes-mirage-firewall/pull/161). Any suggestion how to observe warnings would be highly welcome.

samoht:
I hadn't realised that we were building with `-warn-error -A` unconditionally. I suggest that we just remove this line. Would that help?

I'm also happy to re-introduce a flag to not build in release mode during development.

hannesm:
> I hadn't realised that we were building with `-warn-error -A` unconditionally

this is there since 87a661e73

> I suggest that we just remove this line. Would that help?

I guess it would. But I've no clue about the side effects and why it was set up as it is now.

> I'm also happy to re-introduce a flag to not build in release mode during development.

I again have not much knowledge of the tradeoffs, I only remember that "not using release mode for JS_of_OCaml leads to huge binaries". So the default (mirage build / dune build) should likely be `release` (in contrast to what `dune` uses as default), but I've no idea whether dune with a default profile set to release via dune-workspace respects the `--profile=dev`?

samoht:
> I've no idea whether dune with a default profile set to release via dune-workspace respects the --profile=dev

yes it does - I suspect that Lucas disabled the warnings to not fail for warnings in vendored dependencies. This has been fixed a while ago (as warnings in `duniverse/` are not an error anymore) and thus could be reverted safely (just need to double check with @TheLortex that it was its initial intention).

And it's indeed better to use the `release` mode for generated unikernels, for instance, to make sure cross-module inlining is enabled.

hannesm:
> > I suggest that we just remove this line. Would that help?
>
> I guess it would. But I've no clue about the side effects and why it was set up as it is now.

Unfortunately it doesn't help too much (done in #1379), since it's only in dune.config and affecting the compilation of config.exe

The real issue is the `profile release` as default. Since I barely know anything what dune does with it under the hood, I'm still puzzled what we should do.

[...]

PR #1470, opened 2023-09-21
Move profile release to makefile

hannesm:
see #1374 for background. now the build system is much more hackable again.


samoht:
LGTM

Issue #1468, opened 2023-09-19
Provide a boot parameter to delay startup by some seconds


hannesm:
Tunrs out, some backends (e.g. muen) need some more time to setup the virtual network card. A configurable delay as boot parameter would be excellent in such a use case.

palainp:
Thanks for the PR, but I wonder, as muen uses solo5 as tender (but maybe at some point this PR will be used for other targets, and my comment will be irrelevant :) ), wouldn't it be possible (hmm maybe it's harder or adds to much complexity :( ) to tackle this issue inside solo5 (e.g. `solo5/bindings/muen/start.c`)?

hannesm:
The question is: what do we want to achieve? Of course, we can implement a lot in a specific target. But this makes the whole system differently behaving depending on the target.

In addition, implementing this in solo5 would mean that the boot arguments need to be passed to solo5 and decoded there.

So, I'm in favour of target-independent solutions -- and while there already happens a lot of stuff before the delay is conducted (such as memory setup, OCaml runtime startup, ...) -- the underlying issue (waiting for the network device to be ready) is handled by the PR provided. :)

palainp:
That is perfectly fine for me, thank you!

PR #1472, opened 2023-09-21
support for an initial delay of unikernels

hannesm:
This provides a boot parameter specifying an amount of seconds to wait before starting up. This solves #1468.

### DIFF
--- a/lib/mirage/mirage.ml
+++ b/lib/mirage/mirage.ml
@@ -295,6 +295,26 @@ let git_http ?authenticator ?headers ?(pclock = default_posix_clock) tcpv4v6 ctx
     =
   Mirage_impl_git.git_http ?authenticator headers $ pclock $ tcpv4v6 $ ctx
 
+let delay = job
+
+let delay_startup =
+  let delay_key = Key.delay in
+  let keys = [ Key.v delay_key ] in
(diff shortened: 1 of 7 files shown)


lib/mirage/mirage.ml
```diff
@@ -295,6 +295,22 @@ let git_http ?authenticator ?headers ?(pclock = default_posix_clock) tcpv4v6 ctx
...
+
+let delay_startup =
+  let delay_key = Key.delay in
+  let keys = [ Key.v delay_key ] in
+  let packages = [ package ~max:"1.0.0" "duration" ] in
+  let connect i _ _ =
+    let modname = match Mirage_impl_misc.get_target i with
+      | `Unix | `MacOSX -> "Unix_os.Time"
+      | `Xen | `Qubes -> "Xen_os.Time"
+      | `Virtio | `Hvt | `Spt | `Muen | `Genode -> "Solo5_os.Time"
```
hannesm, line 308: this duplicates some code, but using a functor imposes that we'd need a module to apply the Time onto, and I don't think it is worth it. This way, it is a rather slick implementation (of course, we now have the information about the Time module name duplicated in mirage_impl_time and here -- happy to hear suggestions that don't require us to have a functor around).

samoht:
LGTM

PR #1473, opened 2023-09-25
default to best-fit allocation policy

hannesm:
OCaml since 4.13 uses best fit as default, but MirageOS never updated...  Thus now, if --allocation-policy is not specified, it got overwritten with next-fit (which is inferior).


hannesm:
//cc @samoht not sure what your `cmdliner-stdlib` codebase does...

reynir (approved):

samoht:
It uses the default of the current OCaml version: https://github.com/mirage/cmdliner-stdlib/blob/main/lib/cmdliner_stdlib.ml#L59 -- so similar change than yours for a recent-enough version of OCaml.

PR #1475, opened 2023-10-19
dune.config file misconfigured

memst:
This PR is intended to fix an issue I've faced when setting up Mirage.

Say we have set up a simple MirageOS project. I will be using [mirage-skeleton hello](https://github.com/mirage/mirage-skeleton/tree/main/tutorial/hello) for this example. To configure and build the project, I run:
```bash
$ mirage configure -t unix
$ make
```

The process generates several files in the root directory, including:
- `dune`
- `dune.build`
- `dune.config`

From what I understand, it should be possible to change the contents of `dune` to use `dune.config` instead of `dune.build`. This is mentioned on the [Mirage 4 release page](https://mirage.io/docs/mirage-4):
> The mirage command-line interface hasn't fundamentally changed, but when a project is configured, the following additional files are generated:
> 
> - **dune.build**: the dune rules to build the unikernel.
> - **dune.config**: the dune rules to build the unikernel's configuration.
> - **dune**: switch between dune.build and dune.config depending on the context.

However, if I try to activate `dune.config` and run `$ dune build`, I get the following error:
```
$cat dune
(include dune.config)

$ cat dune.config 
(data_only_dirs duniverse)

;; Generated by mirage.v4.4.0


(executable
 (name config)
 (modules config)
 (libraries mirage))

$ dune build
File "dist/dune", line 3, characters 0-137:
3 | (rule
4 |  (mode (promote (until-clean)))
5 |  (target hello)
6 |  (enabled_if (= %{context_name} "default"))
7 |  (action
8 |   (copy ../hello %{target}))
9 | )
Error: No rule found for hello
```

The culprit is a rule in `dist/dune` that should only be active when `dune.build` is running, but it is never disabled and therefore interferes with `dune.config`. `dune.config` is already trying to solve a similar issue on its first line by ignoring other `dune` files that are active in other parts of the repository:
```
$ cat dune.config 
(data_only_dirs duniverse)
<...>
```

I suggest adding another exclusion to this rule to solve the issue:
```
(data_only_dirs duniverse dist)
```


hannesm:
Thanks for your detailed description, I find this very convincing and would <3 to merge it.

Since I do not know much of the "generate files" part of mirage, I'd kindly ask the experts in this area @TheLortex @samoht (maybe @dinosaure) to express their opinion.

samoht:
This fix looks sensible, but to understand what you are trying to do a bit better: when you say "if I try to activate dune.config" just mean `echo (include dune.config) > dune`? Any reasons you are trying to do this? :-)

memst:
> when you say "if I try to activate dune.config" just mean echo (include dune.config) > dune?

Yes, that's correct. I want to make dune build the `config.ml` again so that `ocamllsp` provides correct syntax highlighting for it. Otherwise I get the following error:

Screenshot 2023-10-19 at 14 04 26

I'm not sure if there's a more canonical way to achieve the same goal. Let me know if there is!

samoht:
Ha ok. I agree it would be nice to have a way to compile both config.ml and unikernel.ml with the same dune configuration. But your PR is already an improvement, so merging.

Issue #1476, opened 2023-10-24
`opam-monorepo` wants to pin many projects when we want to `make depends`


dinosaure:
I tried to test #1455 but my experimentation did not go further due to a weird behavior of the _upstream_ version of `mirage`. It seems that for the `docteur` project (but I suspect it is the same for projects into our `mirage-skeleton` repository), `mirage` produces an OPAM file with many pin-depends. It's an addition of a hundred or so libraries as a pin, which of course pollutes the current switch. I'm not sure how this was introduced and I don't understand why it's not observable in our CI. So I produced a recording of my terminal to show that a "basic" use of `mirage` is actually a problem.

You can run it with:
```shell-session
$ unzip log.zip
$ cd log
$ scriptreplay --timing=mirage.timing mirage.log
```

[log.zip](https://github.com/mirage/mirage/files/13120246/log.zip)

I will try to look into it more deeply tomorrow, but, if we agree about my "basic" usage, we must fix this issue before #1455.

dinosaure:
The issue exist with `opam-monorepo.0.3.6`, the regression seems related to the `mirage` tool so.

dinosaure:
Ok, I can confirm that if I revert #1462, the behavior disappears. I will look reasons behind this regression.

samoht:
I'll have a look at the issue. I suspect the`opam.locked` file explicitely pins all the package to a specific version (and use `pin-depends` to make sure the location/hashes are independent of any changes in `pam-repository`).

PR #1477, opened 2023-10-24
Revert "use the .opam.locked file for installing the switch dependencies"

hannesm:
Reverts mirage/mirage#1462

The issue is outlined in #1476 -- with the code as is, the `opam install --deps-only $<` pin-depends everything (even those packages only meant for the `duniverse`. Let's revert that commit, and suffer from #1415


hannesm:
An alternative would be to embed yet another option into the makefile, namely `--ignore-pin-depends` -- to avoid the pinning of a million packages. BUT this would break in non-obvious ways a package installation if at some point we'd have a pin-depend in that opam file that is not `?vendor` / `?monorepo`'d.

reynir (approved):

dinosaure:
I've convinced myself that this is the best solution for the time being.

PR #1481, opened 2023-11-10
Use Cmdliner.Arg.doc_alts_enum

reynir:
This should make it easier to maintain doc strings.

Downsides are:
- phrasings are changed slightly
- the text is styled differently: instead of italic (or underline in my system) the values are rendered bold.


hannesm (approved):
thanks, this looks good to me :)

PR #1482, opened 2023-11-10
mirage_runtime: restore converters for log_threshold and allocation_policy

hannesm:
Previously, they were part of the Arg submodule. Now they're in their own Conv module. The reason is that mirage-monitoring uses this value for dynamically adjusting the log level.

### DIFF
--- a/lib_runtime/mirage/mirage_runtime.mli
+++ b/lib_runtime/mirage/mirage_runtime.mli
@@ -30,6 +30,17 @@ val set_level : default:Logs.level option -> log_threshold list -> unit
     sources appearing in [l] be used. *)
 
 val logs : log_threshold list Term.t
+(** [logs] is a command-liner term for setting the log_threshold. *)
+
+(** {2 Mirage command-line converters} *)
+module Conv : sig
+  val log_threshold : log_threshold Cmdliner.Arg.conv
+  (** [log_threshold] converts log reporter threshold. *)
+
+  val allocation_policy :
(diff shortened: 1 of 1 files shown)


hannesm:
I'm open for suggestions, the code in place (mirage-monitoring) is:

```OCaml
let adjust_log_level s = (* s being a string such as "tcp:error,whateer:debug,something:quiet,*:warning" *)
  let ts =
    List.map
      (fun s -> (fst Mirage_runtime.Arg.log_threshold) s) (* with this PR, it'll be Mirage_runtime.Conv.log_threshold *)
      (String.split_on_char ',' s)
  in
  let* oks =
    List.fold_left (fun acc t ->
        let* acc = acc in
        match t with
        | `Ok l -> Ok (l :: acc)
        | `Error msg -> Error msg)
      (Ok []) ts
  in
  Mirage_runtime.set_level ~default:(Logs.level ()) oks;
  Ok `Empty
```

reynir (approved):
I'm not sure it's useful to expose `allocation_policy`. As far as I can see at runtime it is not useful - it is only used in code generation:

https://github.com/mirage/mirage/blob/4939ca6a058e8fa78201d103ba3673222f504cb2/lib/mirage/mirage.ml#L420-L426

It would however be *very* nice to have the `log_threshold` converter back!

hannesm:
ok, let's merge this, and figure later out if we would like to remove the allocation_policy converter. (there's as well the possibility to use mirage-logs.cli and cmdliner-stdlib -- but I'm afraid they're changing the default values and introduce more dependency complexity, so let's wait at least another release with it).

PR #1484, opened 2023-11-10
fix log_threshold documentation

hannesm:
previously, the rendered `--help` included `@` characters


hannesm:
I found the section

```
ENVIRONMENT
       These environment variables affect the execution of retreat:

       MIRAGE_LOGS
           See option --logs.
```

always very confusing -- let's remove it, since most MirageOS targets won't do anything with environment variables. The last commit removes the MIRAGE_LOGS environment variable support.

reynir (approved):

PR #1485, opened 2023-11-10
Document in the --help output also the exit codes that will be present.

hannesm:


hannesm:
I removed the default cmdliner exit codes, and manually crafted the list, since we use a different set of exit codes anyways (doing a match on `eval_value` with explicit `exit` calls.

PR #1486, opened 2023-11-12
Remove some functions, also remove dependency on astring (replace code to use Stdlib String module)

hannesm:
best reviewed commit-by-commit. I'm pretty happy that Mirage_impl_misc does not carry these functions anymore (which were needed in mirage3 days to construct the pkg-config based library linking stuff).


lib/mirage/impl/mirage_impl_kv.ml
```diff
@@ -11,8 +10,12 @@ let crunch dirname =
     | '0' .. '9' | 'a' .. 'z' | 'A' .. 'Z' -> true
     | _ -> false
   in
-  let modname = String.filter is_valid dirname in
-  let name = "Static_" ^ String.Ascii.lowercase modname in
+  let modname =
+    let b = Bytes.make (String.length dirname) '_' in
+    String.iteri (fun idx c -> if is_valid c then Bytes.set b idx c) dirname;
+    Bytes.unsafe_to_string b
+  in
```
reynir, line 17: This differs but that's okay since we are only using this for making up names for the output. We can as well use `String.map`:
```suggestion
  let modname = String.map (fun c -> if is_valid c then c else '_') dirname in
```
I would furthermore suggest inlining `modname` as below it is being shadowed, and this is actually **not** the final module name. The `modname` as defined here *could* start with an underscore, and that would not be a legal OCaml module name.

lib/mirage/impl/mirage_impl_block.ml
```diff
@@ -326,10 +326,13 @@ let ccm_block ?nonce_len key =
     | [ block ] ->
         Fmt.str
           {ocaml|let key = %a in
-                 let key = match Astring.String.cut ~sep:"0x" key with
```
reynir, line 329: Can we then remove `astring` from `packages` above?

lib/mirage/impl/mirage_impl_kv.ml
hannesm: that's a great suggestion, done in 295fe8cf

lib/mirage/impl/mirage_impl_block.ml
hannesm: great suggestion, done in 9e3d3cd0

reynir (approved):

PR #1487, opened 2023-11-12
If Functoria_key.key is called too late, raise an error

hannesm:


reynir (commented):
I wonder if there might be an issue with this TODO:
https://github.com/mirage/mirage/blob/main/lib/functoria/lib.ml#L492

hannesm:
No clue whether this is a good change or not. Since I don't understand the mirage/functoria code too well, I'll just close and leave it to the experts to move on.

Issue #1483, opened 2023-11-10
new release -- testing out the current main branch


hannesm:
Dear all @mirage/core ,

I tried the main branch (well, plus #1480) out with the (rather small) dns-secondary unikernel -- please have a look at the diff (at https://github.com/robur-coop/dns-secondary/pull/4) -- but also reproduced below:

```OCaml
diff --git a/config.ml b/config.ml
index 9d16630..41e8970 100644
--- a/config.ml
+++ b/config.ml
@@ -2,10 +2,6 @@
 
 open Mirage
 
-let keys =
-  let doc = Key.Arg.info ~doc:"nsupdate keys (name:type:value,...)" ["keys"] in
-  Key.(create "keys" Arg.(opt (list string) [] doc))
-
 let dns_handler =
   let packages =
     [
@@ -13,19 +9,17 @@ let dns_handler =
       package ~min:"5.0.0" ~sublibs:["mirage"] "dns-server";
       package "dns-tsig";
     ]
-  and keys = [ Key.v keys ]
   in
   foreign
-    ~keys
     ~packages
     "Unikernel.Main" (random @-> pclock @-> mclock @-> time @-> stackv4v6 @-> job)
 
 let enable_monitoring =
-  let doc = Key.Arg.info
+  let doc = Cmdliner.Arg.info
       ~doc:"Enable monitoring (only available for solo5 targets)"
       [ "enable-monitoring" ]
   in
-  Key.(create "enable-monitoring" Arg.(flag ~stage:`Configure doc))
+  Key.(create "enable-monitoring" Arg.(flag doc))
 
 let stack = generic_stackv4v6 default_network
 
@@ -35,47 +29,39 @@ let management_stack =
     (generic_stackv4v6 ~group:"management" (netif ~group:"management" "management"))
     stack
 
-let name =
-  let doc = Key.Arg.info ~doc:"Name of the unikernel" [ "name" ] in
-  Key.(v (create "name" Arg.(opt string "a.ns.robur.coop" doc)))
+let (name : string Runtime_key.key) = Runtime_key.create "Unikernel.K.name"
 
 let monitoring =
-  let monitor =
-    let doc = Key.Arg.info ~doc:"monitor host IP" ["monitor"] in
-    Key.(v (create "monitor" Arg.(opt (some ip_address) None doc)))
-  in
+  let monitor = Runtime_key.create "Unikernel.K.monitor" in
   let connect _ modname = function
     | [ _ ; _ ; stack ] ->
[...]

PR #1488, opened 2023-11-12
move runtime keys back (as strings) to mirage_runtime_key

hannesm:
this is mainly meant for discussion


reynir:
What is the motivation?

hannesm:
Basically to move the definition closer to its usage, and don't spread everything over the different modules and packages... Maybe worth to read https://github.com/mirage/mirage/issues/1483#issuecomment-1807259100 and subsequent remarks

hannesm:
I lack motivation here, closing.

PR #1491, opened 2023-11-23
lower bound to 4.4.1

hannesm:


lib/mirage/mirage_key.ml
```diff
@@ -205,7 +205,7 @@ let space_overhead =
   let doc =
     "The percentage of live data of wasted memory, due to GC does not \
      immediately collect unreachable blocks. The major GC speed is computed \
-     from this parameter, it will work more if smaller. Default: 80."
+     from this parameter, it will work more if smaller. Default: 120."
```
reynir, line 208: Maybe `Printf.sprintf "[...]. Default: %d" Gc.space_over_head` so we don't have to maintain this number? Maybe this is for another PR...

reynir (approved):

lib/mirage/mirage_key.ml
hannesm: The issue is that the output of `mirage configure` emits OCaml code, which could then be stored somewhere, and at another point in time (with a different OCaml compiler & runtime) being compiled... Thus, in order to fix this properly, we should delay the evaluation (of the documentation and default value) until the code will be compiled.

Since there has been quite some work on keys in the main branch, maybe it is not worth to deeply look into this in the 4.4 branch... though I've no clue what to do with the main branch.

PR #1492, opened 2024-01-05
Provide actual network interface name (regression in #1455)

samoht:
I've re-open #1480 as the underlying branch was deleted.


hannesm:
Thanks for your insight in https://github.com/mirage/mirage/pull/1480#issuecomment-1878863094 about the changes to the run.t file. Since CI passes, let's merge this :)

PR #1500, opened 2024-03-05
Add line directives to OCaml files that are copied around

samoht:
This will help to get better locations in error messages.

This is on-top of #1499 and extracted from #1493

### DIFF
--- a/test/functoria/e2e/configure.t
+++ b/test/functoria/e2e/configure.t
@@ -27,6 +27,7 @@ is passed:
                               "dry_run" = false }
   config.exe: [INFO] Generating: main.ml (main file)
   config.exe: [INFO] Generating: key_gen.ml (keys)
+  config.exe: [INFO] Generating: key_gen.ml (keys)
   config.exe: [INFO] Generating: dune.build (dune.build)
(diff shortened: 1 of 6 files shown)


test/functoria/e2e/configure.t
```diff
@@ -27,6 +27,7 @@ is passed:
                               "dry_run" = false }
   config.exe: [INFO] Generating: main.ml (main file)
   config.exe: [INFO] Generating: key_gen.ml (keys)
+  config.exe: [INFO] Generating: key_gen.ml (keys)
```
hannesm, line 30: how comes this is now repeated? is that intentional?

samoht: yea, this is an unfortunate side-effect of computation the generated files (to detect if they need a `copy_file` or a `copy_file#` directive depending on their extension). However, while the action is done twice (thus generating two messages in the logs), actually writing the file is only performed once. I'm not sure how to solve that log duplication (and if it's worth fixing it).

hannesm: I'm fine with your explanation, and don't think (from your explanation) that it is worth fixing.

PR #1502, opened 2024-03-05
4.4 for main

hannesm:
//cc @samoht I went through the 4.4 branch and manually reconstructed the patches relevant for the main branch.


samoht:
Many thanks, that's really useful!

PR #1504, opened 2024-03-06
Add line locations to device connect functions

samoht:
This is kept optional for the `main` function at the moment, not to break every unikernel out there. We can make this mandatory if needed in the future.

Extracted from #1493 

To be tested with https://github.com/mirage/mirage-skeleton/pull/385

It would be nice to test what happens when users get the wrong number of functor parameters- where it breaks and whether the location of the error message is correct. i.e. check that we haven't actually made things worse in the main case (even if it is now much better when the device code got their connect function wrong -- but this is not supposed to happen when users use the existing devices!)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -60,6 +60,11 @@
        they are not available `key_gen.ml` anymore. As a consequence,
        `Key_gen.target` has been removed.
 
+- BREAKING: the `connect` functions passed to create devices (with `impl`)  now
+  take values of type `code` instead of `string`. Value of type `code` are
+  created using a new `code` function, that takes `~pos:__POS__` as parameter.
+  This is used to generate better locations in the generated code, leading to
+  better error messages (#1504, @samoht)
 - BUGFIX: fix `mirage describe` output (#1446 @samoht), add test (#1458 @samoht)
 - Remove ipaddr from runtime (#1437 @samoht, #1465 @hannesm)
 - BREAKING: Mirage.register (and Functoria.register) no longer have `?packages`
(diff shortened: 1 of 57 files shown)


samoht:
@mirage/core any opinion on those error messages? The one the still refers to the generated `main.ml` is fixable by threading more `__LOC__` in many more places. I'm not sure that it's worth breaking the API. The current patch "only" breaks config.ml that define a new device. For instance, mirage-skeleton needs just this patch to compile:

```ocaml
diff --git a/applications/git/config.ml b/applications/git/config.ml
index dc2c338..dd59005 100644
--- a/applications/git/config.ml
+++ b/applications/git/config.ml
@@ -17,13 +17,13 @@ let git_impl path =
   let connect _ modname _ =
     match path with
     | None ->
-        Fmt.str
+        code ~pos:__POS__
           {ocaml|%s.v (Fpath.v ".") >>= function
                  | Ok v -> Lwt.return v
                  | Error err -> Fmt.failwith "%%a" %s.pp_error err|ocaml}
           modname modname
     | Some key ->
-        Fmt.str
+        code ~pos:__POS__
           {ocaml|( match Option.map Fpath.of_string %a with
                  | Some (Ok path) -> %s.v path
                  | Some (Error (`Msg err)) -> failwith err
diff --git a/applications/http/config.ml b/applications/http/config.ml
index d6607e6..0c5b002 100644
--- a/applications/http/config.ml
+++ b/applications/http/config.ml
@@ -21,7 +21,7 @@ let minipaf =
 let conn =
   let connect _ modname = function
     | [ _pclock; _tcpv4v6; ctx ] ->
-        Fmt.str {ocaml|%s.connect %s|ocaml} modname ctx
+        code ~pos:__POS__ {ocaml|%s.connect %s|ocaml} modname ctx
     | _ -> assert false
   in
   impl ~connect "Connect.Make" (pclock @-> tcpv4v6 @-> mimic @-> conn)
```

PR #1515, opened 2024-04-08
Make config.ml and unikernel.ml co-buildable

samoht:
This helps in dev mode when to get merlin working on all files. This makes `mirage` installed in the monorepo.


samoht:
This is a follow-up of https://github.com/mirage/mirage/pull/1475 (somehow extracted from #1495). The idea is always to be able to build both `config.ml` and `unikernel.ml` simultaneously so that tools like `merlin` can work on both seamlessly.

The drawback of this PR is that `mirage` becomes a `duniverse` dependency - that might make pinning dev versions of `mirage` a bit more convoluted (but possible). Maybe I should just document how to do this somewhere.

hannesm:
While I appreciate to have merlin support for editing config.ml, I'm sceptical of this change for multiple reasons:
- how do we ensure that the mirage installed in the switch is the same as in "duniverse" (and what happens if the versions diverge?)
- how is mirage now cross-compiled within the duniverse for the solo5 target (is it? does that work properly? is the mirage binary from the duniverse ever used?)
- another issue is about capturing the software that went into the unikernel binary -- here mirage is then listed twice -- once as part of the duniverse, once in the opam switch -- I don't really like to have them mixed

I agree that the main feature - being able to use merlin on config.ml - would be great to have. but another drawback of this PR is that only once mirage configure and get the duniverse directory have been done, this is enabled. IMHO we should take a step back at the drawing board and figure out a different solution for that problem which doesn't intermix the unikernel and config dependencies.

of course, it may be that I misunderstand the PR and its consequences - but I'd appreciate to have a broader discussion of such a heavy change. thanks.

samoht:
I understand the concerns and I'm happy to revert the PR to give more time to discussions. 

To answer the questions first to see that this is actually not so different from what we have now:

- what happens if the mirage binary is different from the switch and duniverse: this is a good question and we should probably do something to force the version installed in the duniverse to be the same as the one in the switch.
- there is no cross-compilation of the mirage tool: this is installed (and compiled) in the "default" switch which is setup for unix/macos.
- I don't think the mirage dependency is listed as a switch dependency anymore? For instance https://github.com/mirage/mirage/pull/1515/files#diff-f802a02971d4e74dffc7f9e1448d8b82998b9b94fd897a1141279a05f5554cfdR48 doesn't contain `mirage`. But maybe you mean the `mirage` tool needs to be installed in both the `.opam` and `.opam.locked` ? To be honest I'm still not sure to understand how that `opam` file is used apart from generating the lock file.

Let me revert and fold these changes/discussions back in #1495

hannesm:
I changed my mind on this PR, if we get the developer experience that merlin works once mirage has its duniverse, let's go for it. IMHO improving developer experience is a great goal, and if we've some dangling dependencies (that do not end up into the unikernel binary), that's fine with me.

Thanks for your work on this :) Please rebase and re-open a PR based on this one :)

samoht:
See https://github.com/mirage/mirage/pull/1582

Issue #1513, opened 2024-03-27
question about retiring mirage-random


hannesm:
Dear Madam or Sir,

I encountered the "mirage-random" opam package, and its usage. I am curious what you think about it?

My feeling is that the interface can just be removed from the dependency cone, and unikernels can use `Mirage_crypto_rng` directly. I don't see much value in the abstraction (the mirage_impl_random only provide mirage-crypto-rng anyways).

I even think that "functorizing over the RNG" is maybe not what we desire to do -- unless there's actual use of a different RNG (or plans to do so)? I guess that "RNG initialization" is still something the mirage utility should do (and here I've no clue whether there's another path than functorizing stuff over the RNG).

From my perspective, this originates from when we introduced nocrypto and entropy harvesting, but with the current state of mirage-crypto-rng (being independent of gmp/zarith), there's no need to abstract/functorize over it (when I repeat for myself that functors are great if and only if you have the requirement to have the same thing using two different modules in the same application). But from where I stand, there's really only ever a single RNG being used, and to reach the CSPRNG properties, our fortuna is the current choice (obviously, happy to get something different if there's new research in that area).

WDYT @mirage/core?

hannesm:
Maybe there's a wider question, and we should address this in the next major release. There are certain things a unikernel is functorized over that I'm rather sad about:
- pclock (for a posix timestamp) (`Mirage_clock.PCLOCK`)
- mclock (for a monotonic clock) (`Mirage_clock.MCLOCK`)
- time (for "sleep_ns") (`Mirage_time.S`)
- random (for generate) (`Mirage_random.S`)

These all don't have an internal state that you act on (something like TCP.t for creating connections) - but usually you use `(R : Mirage_random.S)` to use `R.generate` in your unikernel.

Now, what are alternatives? Do we actually care which modules use these functionality? Could the dune virtual library help us, and we simply use `Mirage_clock.now ()` and get a `Ptime.t` (and mirage takes care which platform we're on?)? Would that work in a testing environment (see TCP tests) as well (i.e. provide a testing clock that runs faster?)? I'm still unsure about this whole testing, I personally think that unit tests (of the pure core) and end-to-end tests are worth, but these intermediate level tests lead to embedding lots of code into the Lwt monad and testing it -- which I'm not too convinced of.

[...]

PR #1521, opened 2024-04-23
Mirage logs no clock

hannesm:
see discussion in #1513


hannesm:
This has been an interesting experiment, but for upstreaming there needs to be some further cleaning that will be done in separate PRs.

PR #1522, opened 2024-04-23
functoria: Quote Runtime_arg.call

Julow:
Fixes codes that use an arg call as an argument to a function:

    code ~pos:__POS__ "f ~arg:%a"
      Runtime_arg.call my_arg

### DIFF
--- a/lib/functoria/runtime_arg.ml
+++ b/lib/functoria/runtime_arg.ml
@@ -66,5 +66,5 @@ let serialize ~runtime_modname fmt (Any k) =
   Format.fprintf fmt "let %s__key = %s.register @@@@\n%a  @[<v2>%s@]\n;;\n"
     (ocaml_name k.name) runtime_modname pp_pos k.pos k.code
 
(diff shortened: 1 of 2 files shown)


samoht:
Thanks! Do you have a use-case for this? Devices are not supposed to call function (the initial goal was that only functoria would use this)

samoht:
With the new API the argument name is passed as the last parameters of the call to connect. You can then just use it directly, you don't need to serialize it yourself.

We should probably remove that function from the public API to avoid people doing this :-)

Julow:
This is indeed not needed in https://github.com/robur-coop/caldav/pull/36, thanks!

`Runtime_arg.call` is still needed within mirage, so I think quoting doesn't hurt.

samoht:
Could you please rebase this PR? While I don't think the change is strictly necessary, it makes the generated code more robust so I'm fine with merging it.

samoht (approved):

PR #1523, opened 2024-04-25
doc: Update inline examples to Runtime_arg

Julow:
Update the code blocks in `mirage.mli` that use the old Key API.


hannesm:
thanks

PR #1525, opened 2024-04-26
add the options to specify "host" (and "socket") for the host system network stack

hannesm:
the native ocaml stack can be used by passing "OCaml" or "direct".

over time, the confusion of "direct" has escalated. it is kept for backwards compatibility

//cc @pitag-ha


reynir (approved):
Thanks! I can never for the life of me get it right what `direct` and `socket` is. I believe in either case a socket is used, and I am not sure what is more direct (and in what sense?!) so the confusion is total for me.

PR #1526, opened 2024-05-06
mirage-runtime: provide sleeper storage

hannesm:
The purpose of this storage is to be able to move the `Time.sleep_ns` functionality to a different place than mirage-solo5/mirage-xen. Since this is a very different change than other things in #1521, I'd appreciate this to be reviewed separately.

As mentioned, this is a prerequisite to solve #1513 and has been carved out of #1521.

### DIFF
--- a/lib_runtime/mirage_runtime.ml
+++ b/lib_runtime/mirage_runtime.ml
@@ -250,6 +250,16 @@ let at_exit f = add f exit_hooks
 let at_leave_iter f = add f leave_iter_hooks
 let at_enter_iter f = add f enter_iter_hooks
 
+type sleep = { time : int64; mutable canceled : bool; thread : unit Lwt.u }
+
+let new_sleepers = ref []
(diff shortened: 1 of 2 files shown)


samoht:
Just to understand the design constraints better: why does this change need to be here and not in `mirage_time.ml` ?

hannesm:
We need in mirage-solo5/mirage-xen access to this sleeper storage, but the mirage-time package should be able to implement the sleep_ns (by pushing stuff into this queue).

Now, we could push the storage into mirage-time, but what would this mean for a unix backend (where the sleeper is implemented differently using Lwt_unix)? We could put it into mirage-solo5 / mirage-xen, but then there can't be a mirage-time.solo5 implementiation for both solo5 and xen (but there would need to be a disjoint mirage-time.xen).

What does it hurt to have it in mirage-runtime, next to the at_exit stuff etc.?

hannesm:
ok, so I agree we can move it to mirage-time. I can't see blockers for doing that.

PR #1527, opened 2024-05-06
fix generated opam file - the build rule is now `make "-Csubdir" "build"

hannesm:
Previously, `sh -exc 'cd subdir && mirage build'` was used in the generated opam file as build command. Since #1404 got merged and released, the mirage build subcommand does no longer exist.

In the opam file, now "make build" is used instead.

### DIFF
--- a/test/mirage/action/test.ml
+++ b/test/mirage/action/test.ml
@@ -14,7 +14,8 @@ let print_banner s =
   print_newline ()
 
 let info context =
-  Info.v ~packages:[] ~keys:[] ~runtime_args:[] ~build_cmd:"mirage build"
+  Info.v ~packages:[] ~keys:[] ~runtime_args:[]
+    ~build_cmd:(fun _ -> "make build")
     ~context ~src:`None ~project_name:"test" "NAME"
 
 let test target =
--- a/test/mirage/action/test.ml
+++ b/test/mirage/action/test.ml
@@ -15,7 +15,10 @@ let print_banner s =
 
 let info context =
   Info.v ~packages:[] ~keys:[] ~runtime_args:[]
-    ~build_cmd:(fun _ -> "make build")
+    ~build_cmd:(fun sub ->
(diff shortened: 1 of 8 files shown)


test/mirage/action/test.ml
```diff
@@ -14,7 +14,8 @@ let print_banner s =
   print_newline ()
 
 let info context =
-  Info.v ~packages:[] ~keys:[] ~runtime_args:[] ~build_cmd:"mirage build"
+  Info.v ~packages:[] ~keys:[] ~runtime_args:[]
+    ~build_cmd:(fun _ -> "make build")
```
samoht, line 18: given that we discard the path argument here, do we really need to change the `build` above to be a `Fpath.t -> string` function?

hannesm: Yes, we do. This is the `test/mirage` directory. In the "normal" generated opam file, the subdirectory is not discarded.

In e.g. the `caldav` repository (https://github.com/robur-coop/caldav), if you `cd mirage && mirage configure -t hvt`, the resulting opam file will now include: `build: [ make "-Cmirage" "build" ]`.

I added aff5e9f4 which uses the argument provided also in the test.

hannesm:
But of course, in the opam file generation, we can avoid the `sub` being passed, and use the `t.subdir` (as done before for `sh -exc 'cd t.subdir && mirage build'`). To me, it is not clear how much logic to encode in which of these layers (Lib.Make.get_cmds vs Opam.pp). Any opinion?

hannesm:
Well, it really is a question where to hardcode these `make` steps. At the moment, we have the makefile targets in `makefile.ml`, and they are as well in `lib.ml` (the `get_cmds`).

Now, we could enrich the `makefile.ml` API to expose the pre_build and build targets, and also figure out where to put the configure_cmd and lock_location, and then use the makefile API from the Opam module. Feels like busywork.

PR #1529, opened 2024-05-07
Mirage time defunctorised

hannesm:
The context is #1513, and #1521. This is a first shot at the mirage-time package, to use dune variants. It builds on top of #1526 and is best read with https://github.com/mirage/mirage-skeleton/pull/394.

I wrote some more text to our mailing list: https://lists.xenproject.org/archives/html/mirageos-devel/2024-05/msg00002.html

This is marked as draft, since I wouldn't want to merge this as-is, but would appreciate feedback first -- maybe the code in mirage can be further simplified: an `impl` that doesn't have a `connect` should maybe not exist in `main.ml` -- i.e. no need to generate anything, since there's nothing to initialize -- we can just use it in `unikernel.ml` -- but my functoria/mirage knowledge is not sufficient to write up such a thing. The default value for `connect` is eventually `return ()` -- while I'd like to have `connect` being an option (but as mentioned, I'm not sure what else needs to change and what will break).

Another remark is about the "no argument check" -- I think this brief hack introduced by myself is not the right thing. We should ensure that `main` (and other jobs) have at least one argument -- but if we pass in one argument (of `()`) for each "device" that doesn't need it, we'll add a lot of things for no good reason (and have again complexity, and it being hard to understand).


lib/functoria/impl.ml
```diff
@@ -42,16 +42,6 @@ and 'a device = ('a, abstract) Device.t
...
-let rec app_has_no_arguments = function
-  | App { args = Cons _; _ } | Dev { args = Cons _; _ } -> false
-  | Dev { dev; _ } when String.equal (Device.module_name dev) "Unit" ->
-      (* special hack for Job.noop *)
-      false
-  | App _ | Dev _ -> true
-  | If { cond = _; branches; default } ->
-      app_has_no_arguments default
-      || List.exists (fun (_, branch) -> app_has_no_arguments branch) branches
-
```
samoht, line 54: Why is this needed? Did the reason why this was added changed with that PR?

hannesm: From my comment above:

Another remark is about the "no argument check" -- I think this brief hack introduced by myself is not the right thing. We should ensure that main (and other jobs) have at least one argument -- but if we pass in one argument (of ()) for each "device" that doesn't need it, we'll add a lot of things for no good reason (and have again complexity, and it being hard to understand).

--> I think this "app_has_no_argument" check was a good first iteration, but what we actually want to check is that "start takes at least an argument" (to have it executed after a log reporter) -- but already with extra_deps and runtime_args we get arguments. So, indeed this removal is unrelated and should not be merged as is. It would be great to have a better check (esp. since e.g. `noop` is already excluded from the check).

samoht:
I like the direction of that one too. A few comments:

- I would move the sleeper machinery in a `Mirage_time.Runtime` module (or something similar, but in the `mirage-time` package) so this is independent of `mirage` (if possible)
- As we need to pass the `extra_dep` parameter around, I still think it'd be helpful to make this type abstract and require `Mirage_time.sleep_ns` to take that parameter. Call it a capability or something else, but I think there is still value in having a way to know when a function needs access to the time device (and if it does, this is reflected in its type)
- What about the error case: what happens if someone uses `Mirage_time.sleep_ns` but forgets to add the `extra_deps` parameter? I guess it'll try to link with `mirage-time.unix`, which will fail on the solo5 target. Do we have a good way to deal with the error? Maybe passing the `Mirage_time.t` parameter around is a good way to help with this? Or something else?
- Related to mirage-time specifically (but also quite minor): it would be great to see if we could use variants to mock the time implementation and write unit tests for the package.

hannesm:
Thanks for your feedback.

I will do (a) move the sleeper space to mirage-time.

With (b) - passing a value of an abstract type around, I'm undecided: what is the value thereof? What is a real use case (or requirement)? At least since mirage-time 1.0 (2016), the interface is `val sleep_ns : int64 -> unit Lwt.t`... In mirage-clock, since 2.0 (2018) the API is `val now_d_ps : unit -> int * int64` (and `val elapsed_ns : unit -> int64`).

For other "devices" (such as a network stack, a network interface, block device, file system, ...) that have data attached to it, I for sure see the value. For time/clock/random, I don't.

(c) indeed, the error is a bit mystic about "unix library hidden".

(d) sure, we can have in mirage-time a ".mock" implementation, but I'm not sure what should it do? directly return? return earlier? If you've a suggestion or an actual test library, please let me know.

hannesm:
When putting the sleeper storage into mirage-time, I ran into the dune issue again. That's fantastic since now there's a pretty small reproduction case https://github.com/hannesm/dune-variant (reported https://github.com/ocaml/dune/issues/10460). Let's hope the dune developers will have some time and be able to fix the root cause of the issue.

lib/functoria/impl.ml
hannesm: superseeded by #1544

hannesm:
I came back here, and various issues have solved over the time (i.e. the mentioned "no argument check" has been fixed in a separate PR).

I think my question is now: how should we proceed? I'm very much in favour to (a) use less functors and (b) generate less code. The PR as presented here is incomplete in that sense, I'd prefer to just remove the `lib/devices/time.ml{i}` entirely.

To achieve this, I'd like a lightweight way to depend on "mirage-time" -- and since there doesn't need any code being generated, I guess any user of mirage-time can just depend on it. Now, the `mirage` utility only needs to select the right variant -- and here there's trouble:
- we can't really use `match_impl` etc. since an `impl` always pushes some code into `main.ml`
- we could unconditionally emit `mirage-time.solo5` (or `mirage-time.unix`) dependency in all `dune.build` (I guess this speaks against minimality)
- the same as for mirage-time can be applied to mirage-clock

Now, would there be any extra binary size when we'd depend on mirage-time and mirage-clock unconditionally? I'd guess no. I also think the final link step is smart enough to get rid of the symbols.

----

I think we need to discuss what the pain points of mirage are nowadays and where we want to go to - or whether we want to go anywhere at all. I got some feedback in the MirageOS workshop at FunOCaml, and have a goal: get rid of config.ml. I know this sounds pretty hard to achieve, but I think we can get there - step-by-step. And as I tried to elaborate before, the less code we generate, the easier it gets.

But of course, there may be other paths and opinions on how mirage should move, I'm interested in your opinions.

palainp:
Dear @hannesm,
Thank for your proposition. I agree that an unconditional addition of mirage_time and mirage_clock will probably not lead, most of the time, to a binary size increment, as I think, most of the time, unikernels in the wild already use them. In the meantime I've checked that assumption against mirage-skeleton, and the current state of mirage-skeleton contains 15 unikernels that does not use `default_time` [1] and more or less the same for `*_clock`.
Anyway I still think a lot of unikernels will use that so it won't change the total size, and even if it's not used, according to `objdump` with unix target, the `Time` code seems to be small: 96B for code and 192B for data (I haven't checked for clocks).
TLDR: I'm in the 'always add that' team :)

[1]:
```bash
$ grep -L default_time {applications,device-usage,tutorial}/*/config.ml | cut -d/ -f2 | sort -u | wc -l
15
```

hannesm:
superseeded by #1579

PR #1533, opened 2024-05-09
Consolidate mirage bootvar

hannesm:
I found an even simpler example than "mirage-time" in respect to "how to use dune variants": boot parameters.

We used to have mirage-bootvar-xen, mirage-bootvar-unix, mirage-bootvar-solo5, and the parse-argv opam packages that were all involved.

Now, I created (retaining history) the single repository https://github.com/mirage/mirage-bootvar. This uses dune variants, the default is the unix implementation.

What is interesting about bootvar is that no one bothered to functorize it ;) Also, it shows how a unikernel could still decide to use "no_argv", bypassing any dune variant restrictions (I'm not sure whether this is desirable, but it is possible -- this also sheds some light that we could use configure-time keys to select some other implementation).

Anyways, this is much simpler than #1529 and you may enjoy looking into it.

EDIT: CI fails since the mirage-bootvar is not released, I'm happy to do that if no one objects about the approach taken.

### DIFF
--- a/lib/devices/argv.ml
+++ b/lib/devices/argv.ml
@@ -2,24 +2,22 @@ open Functoria.DSL
 
 let ty = Functoria.argv
 
-let argv_unix =
-  let packages = [ package ~min:"0.1.0" ~max:"0.2.0" "mirage-bootvar-unix" ] in
-  let connect _ _ _ = code ~pos:__POS__ "Bootvar.argv ()" in
-  impl ~packages ~connect "Bootvar" ty
-
-let argv_solo5 =
-  let packages = [ package ~min:"0.6.0" ~max:"0.7.0" "mirage-bootvar-solo5" ] in
-  let connect _ _ _ = code ~pos:__POS__ "Bootvar.argv ()" in
-  impl ~packages ~connect "Bootvar" ty
-
 let no_argv =
   let connect _ _ _ = code ~pos:__POS__ "return [|\"\"|]" in
   impl ~connect "Mirage_runtime" ty
 
-let argv_xen =
-  let packages = [ package ~min:"0.8.0" ~max:"0.9.0" "mirage-bootvar-xen" ] in
-  let connect _ _ _ = code ~pos:__POS__ "Bootvar.argv ()" in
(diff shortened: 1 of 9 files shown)


reynir (approved):
To me that looks great!

hannesm:
To go a bit deeper on my motivation: I find this makes it much easier for developers to discover the code MirageOS uses. It can easily be compared what the difference between the different targets is, and API changes can be done very locally without having to think about the "release train" and in which order to release what.

Furthermore, the complexity on tooling (let's say opam) is reduced by the reduction of opam packages that are required in the universe.

Also, the maintenance of mirage is relaxed - we can replace (archive) 4 git repositories with all sorts of (partially outdated) metadata by a single one.

I guess a question is whether we should "buy in" into dune variants -- and my point of view is that it radically simplifies our tooling, plus enhances the developer experience (not shown here, since there isn't any functor removed) -- and one of the (at least for me) most useful feature of IDEs (or editor support), the "jump to definition", is finally on the horizon to work just neatly.

palainp:
Thank you, this PR looks good to me!

samoht:
This change seems very uncontroversial and so looks great to me - could you please update the test outputs to please the CI?

hannesm:
This needs to wait for https://github.com/ocaml/opam-repository/pull/25887 being merged to have the CI pass.

hannesm:
The commit [a0dc4ac](https://github.com/mirage/mirage/pull/1533/commits/a0dc4ac5512e7dc1de9be7c5404d5961e10e2aac) fixes an actual bug that was discovered due to how mirage-bootvar is put into place. I'll mention this explicitly in the CHANGES.

hannesm:
I moved forth and back... so our `Package.v  ~sublibs:["x"] "foo"` does add a dependency onto `foo.x`.
The documentation said `[ "foo" ; "foo.x" ]` would be the dependency.

Reality shows there are some packages that don't provide `foo` -- so for the mirage-bootvar case, I specified two packages: `"mirage-bootvar"` and `"mirage-bootvar" ~sublibs:["unix"|"solo5"|"xen"]`. This should result in CI compiling :)

PR #1543, opened 2024-05-20
Reverse the dependency between happy-eyeballs and dns-client

dinosaure:
use with https://github.com/mirage/mirage-skeleton/pull/398

### DIFF
--- a/lib/devices/happy_eyeballs.ml
+++ b/lib/devices/happy_eyeballs.ml
@@ -8,7 +8,8 @@ type happy_eyeballs = Happy_eyeballs
 
 let happy_eyeballs = typ Happy_eyeballs
 
-let generic_happy_eyeballs timer_interval =
+let generic_happy_eyeballs aaaa_timeout connect_delay connect_timeout
+    resolve_timeout resolve_retries timer_interval =
   let packages =
     [ package "happy-eyeballs-mirage" ~min:"1.1.0" ~max:"1.2.0" ]
   in
@@ -16,10 +17,18 @@ let generic_happy_eyeballs timer_interval =
   let err () = connect_err "generic_happy_eyeballs" 3 ~max:4 in
   let connect _info modname = function
     | _time :: _mclock :: stack :: rest ->
+        let aaaa_timeout, rest = pop ~err aaaa_timeout rest in
+        let connect_delay, rest = pop ~err connect_delay rest in
+        let connect_timeout, rest = pop ~err connect_timeout rest in
+        let resolve_timeout, rest = pop ~err resolve_timeout rest in
+        let resolve_retries, rest = pop ~err resolve_retries rest in
         let timer_interval, rest = pop ~err timer_interval rest in
         let () = match rest with [] -> () | _ -> err () in
-        code ~pos:__POS__ {ocaml|Lwt.return (%s.create %a %s)|ocaml} modname
-          (pp_opt "timer_interval") timer_interval stack
+        code ~pos:__POS__ {ocaml|Lwt.return (%s.create %a%a%a%a%a%a %s)|ocaml}
+          modname (pp_opt "aaaa_timeout") aaaa_timeout (pp_opt "connect_delay")
+          connect_delay (pp_opt "connect_timeout") connect_timeout
+          (pp_opt "resolve_timeout") resolve_timeout (pp_opt "resolve_retries")
+          resolve_retries (pp_opt "timer_interval") timer_interval stack
     | _ -> err ()
   in
   impl ~runtime_args ~packages ~connect "Happy_eyeballs_mirage.Make"
--- a/lib/devices/happy_eyeballs.ml
+++ b/lib/devices/happy_eyeballs.ml
@@ -13,7 +13,16 @@ let generic_happy_eyeballs aaaa_timeout connect_delay connect_timeout
   let packages =
     [ package "happy-eyeballs-mirage" ~min:"1.1.0" ~max:"1.2.0" ]
   in
-  let runtime_args = runtime_args_opt [ timer_interval ] in
+  let runtime_args =
+    runtime_args_opt
+      [
+        aaaa_timeout;
+        connect_delay;
+        resolve_timeout;
+        resolve_retries;
+        timer_interval;
+      ]
+  in
   let err () = connect_err "generic_happy_eyeballs" 3 ~max:4 in
   let connect _info modname = function
     | _time :: _mclock :: stack :: rest ->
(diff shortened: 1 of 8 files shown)


lib/devices/happy_eyeballs.ml
```diff
@@ -8,36 +8,18 @@ type happy_eyeballs = Happy_eyeballs
...
-  let err () = connect_err "generic_happy_eyeballs" 3 ~max:8 in
+  let runtime_args = runtime_args_opt [ timer_interval ] in
+  let err () = connect_err "generic_happy_eyeballs" 3 ~max:4 in
   let connect _info modname = function
     | _time :: _mclock :: stack :: rest ->
-        let aaaa_timeout, rest = pop ~err aaaa_timeout rest in
-        let connect_delay, rest = pop ~err connect_delay rest in
-        let connect_timeout, rest = pop ~err connect_timeout rest in
-        let resolve_timeout, rest = pop ~err resolve_timeout rest in
-        let resolve_retries, rest = pop ~err resolve_retries rest in
```
hannesm, line 34: why removal of these arguments?

hannesm:
now that dns and happy-eyeballs is released, is this ready @dinosaure ?

dinosaure:
I just fixed the device according to the last release of `dns.8.0.0`. A double-check will be appreciate. I took the example available here: https://github.com/mirage/ocaml-dns/pull/346#issuecomment-2120612738

hannesm:
thanks, I rebased onto main and force-pushed. let's see whether CI is happy

hannesm:
The mirage-skeleton CI fails at:
```
File "applications/http/config.ml", line 39, characters 35-38:
39 |     mimic_happy_eyeballs stackv4v6 dns (generic_happy_eyeballs stackv4v6 dns)
                                        ^^^
Error: This expression has type happy_eyeballs impl -> dns_client impl
       but an expression was expected of type dns_client impl
```

hannesm:
I opened https://github.com/mirage/mirage-skeleton/pull/398 which should fix the issue observed. I also pushed a commit here which fixes the code in the documentation.

hannesm:
CI failure is now waiting for conduit being released, I pushed into mirage-dev the release https://github.com/mirage/mirage-dev/commit/8fd6ad85421ba4a75099c1d7634c1d33d7f7bff3

hannesm:
Ok, I pushed here and to https://github.com/mirage/mirage-skeleton/pull/398 some commits, CI is happy now.

What is remarkable, and we should be aware of this, is that now since mimic_happy_eyeballs no longer takes a dns_client, we need to pass the dns_client to the unikernel -- otherwise we will have an happy-eyeballs instance that is not able to resolve hostnames -- @dinosaure please have a look at the changes, esp. the git or http unikernel: https://github.com/mirage/mirage-skeleton/pull/398/files#diff-34715b4b5bca7e52e32c853d9b9334ae4355d8b8be7056010c9128a70eac2bd9 https://github.com/mirage/mirage-skeleton/pull/398/files#diff-843e3232dba77394d550c0ed3e00efb387d9ad6d7e04c51f781a41f4611f9f34

From my point of view, this is fine. Another approach could be to re-add the Dns_client to mimic-happy-eyeballs, and then we need fewer changes and will always have a mimic-happy-eyeballs which is able to resolve DNS hostnames. WDYT? This would require some changes to mimic-happy-eyeballs and another release thereof.

lib/devices/mimic.ml
```diff
@@ -9,11 +8,11 @@ type mimic = Mimic
 let mimic = typ Mimic
 
 let mimic_happy_eyeballs =
-  let packages = [ package "mimic-happy-eyeballs" ~min:"0.0.5" ] in
+  let packages = [ package "mimic-happy-eyeballs" ~min:"0.0.8" ] in
```
hannesm, line 11: just as a reminder to my future self -- here earlier mirage version do not restrict the mimic-happy-eyeballs upper bound, thus 0.0.8 could be used... but given that mimic-happy-eyeballs 0.0.8 requires dns-client 8.0.0 (and earlier mimic-happy-eyeballs require dns-client < 8), this is fine.

in general, I prefer if we put upper bounds on all the packages to allow future development of these packages. It is not clear to me what the versioning scheme of mimic-happy-eyeballs is - thus whenever its `Make` functor changes (or `connect`), will the major version be increased? or the minor? at the moment there are only patch-level releases - but for mirage it'd be very nice to have such a versioning scheme where we can then put adequate upper bounds in the tool.

hannesm:
> From my point of view, this is fine. Another approach could be to re-add the Dns_client to mimic-happy-eyeballs, and then we need fewer changes and will always have a mimic-happy-eyeballs which is able to resolve DNS hostnames. WDYT? This would require some changes to mimic-happy-eyeballs and another release thereof.

My current favourite would be to revert the mimic-happy-eyeballs change (and re-introduce a dns-client dependency there), this would mean fewer changes to unikernels and would bring us that a mimic-happy-eyeballs is always able to resolve hostnames -- a nice property :)

dinosaure:
I'm ok with your change, I also re-introduce dns into mimic. Would you like another release and should we cut a release of mirage now?

hannesm:
> I'm ok with your change, I also re-introduce dns into mimic. Would you like another release and should we cut a release of mirage now?

I'd be in favour of another mimic release, and then cutting the mirage release :)

hannesm:
the CI results are looking great -- with mimic 0.0.9 being in https://github.com/mirage/mirage-dev all of mirage-skeleton compiles. So I'll cut a release once mimic is merged into opam-repository.

lib/devices/happy_eyeballs.ml
```diff
@@ -12,34 +11,25 @@ let happy_eyeballs = typ Happy_eyeballs
...
-      [
-        aaaa_timeout;
-        connect_delay;
-        resolve_timeout;
-        resolve_retries;
-        timer_interval;
-      ]
-  in
-  let err () = connect_err "generic_happy_eyeballs" 4 ~max:9 in
+  let runtime_args = runtime_args_opt [ timer_interval ] in
```
hannesm, line 16: I'm unsure about this change -- why is the runtime_args list now much smaller?

hannesm:
while looking at the code, this is good to merge. but the usage of `runtime_args_opt` makes these arguments tedious -- since a unikernel needs to define their own keys and provide them (in dns, git, happy_eyeballs, ip). should be revisited before a release.

hannesm:
I merged this. I still think we may want to define the boot parameters inside mirage (instead of in each unikernel).

PR #1544, opened 2024-05-27
job with no arguments, check a bit more thoroughly

hannesm:
This adds a test case (initially failing) where a device with some runtime arguments should not fail the `configure` stage.

The reasoning is: we really need to check that execution is delayed (i.e. it is not `let start = ...`, but more `let start _ = ...`).

Previously, we only checked for `args` -- which are the functor applications. Now, we also check for extra_deps - which are passed as arguments - and furthermore we check whether runtime_args is empty. And only in this case, where all of the three things are empty, and in addition we're not at the `noop` job, we raise an error.

Addresses @samoht comment on https://github.com/mirage/mirage/pull/1529#discussion_r1604834773

### DIFF
--- a/lib/functoria/impl.ml
+++ b/lib/functoria/impl.ml
@@ -42,12 +42,15 @@ and 'a device = ('a, abstract) Device.t
 
 let abstract t = Abstract t
 
-let rec app_has_no_arguments = function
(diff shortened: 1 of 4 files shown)


hannesm:
I'll merge this, it is an improvement: existing test cases pass, and even more (which have been added in this PR).

PR #1548, opened 2024-06-06
Upgrade the tool to use tcpip.8.1.0 and mirage-qubes.0.10.0

dinosaure:
/cc @hannesm, I'm not sure that it's enough for your recent change about tcpip. At least, this change is required for `qubes-miragevpn`.

### DIFF
--- a/lib/devices/ip.ml
+++ b/lib/devices/ip.ml
@@ -85,7 +85,7 @@ type ipv6_config = {
 (** Types for IP manual configuration. *)
 
 let ipv4_qubes_conf =
-  let packages = [ package ~min:"0.9.0" ~max:"0.10.0" "mirage-qubes-ipv4" ] in
+  let packages = [ package ~min:"0.10.0" ~max:"0.11.0" "mirage-qubes-ipv4" ] in
   let connect _ modname = function
     | [ db; _random; _mclock; etif; arp ] ->
         code ~pos:__POS__ "%s.connect@[@ %s@ %s@ %s@]" modname db etif arp
--- a/lib/devices/ip.ml
+++ b/lib/devices/ip.ml
@@ -85,7 +85,7 @@ type ipv6_config = {
 (** Types for IP manual configuration. *)
 
 let ipv4_qubes_conf =
-  let packages = [ package ~min:"0.10.0" ~max:"0.11.0" "mirage-qubes-ipv4" ] in
+  let packages = [ package ~min:"0.9.0" ~max:"0.11.0" "mirage-qubes-ipv4" ] in
   let connect _ modname = function
     | [ db; _random; _mclock; etif; arp ] ->
         code ~pos:__POS__ "%s.connect@[@ %s@ %s@ %s@]" modname db etif arp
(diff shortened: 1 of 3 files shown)


lib/devices/ip.ml
```diff
@@ -85,7 +85,7 @@ type ipv6_config = {
 (** Types for IP manual configuration. *)
 
 let ipv4_qubes_conf =
-  let packages = [ package ~min:"0.9.0" ~max:"0.10.0" "mirage-qubes-ipv4" ] in
+  let packages = [ package ~min:"0.10.0" ~max:"0.11.0" "mirage-qubes-ipv4" ] in
```
hannesm, line 88:
```suggestion
  let packages = [ package ~min:"0.9.0" ~max:"0.11.0" "mirage-qubes-ipv4" ] in
```

The 0.9 release is fine as well, we don't need to update the lower bound, only allow the upper bound.

lib/devices/qubesdb.ml
```diff
@@ -5,7 +5,7 @@ open Misc
 type qubesdb = QUBES_DB
 
 let qubesdb = typ QUBES_DB
-let pkg = package ~min:"0.9.0" ~max:"0.10.0" "mirage-qubes"
+let pkg = package ~min:"0.10.0" ~max:"0.11.0" "mirage-qubes"
```
hannesm, line 8:
```suggestion
let pkg = package ~min:"0.9.0" ~max:"0.11.0" "mirage-qubes"
```

see above

lib/mirage.ml
```diff
@@ -164,7 +164,7 @@ let socket_stackv4v6 = Stack.socket_stackv4v6
 let tcpv4v6_of_stackv4v6 v =
   let impl =
     let packages =
-      [ package "tcpip" ~sublibs:[ "stack-direct" ] ~min:"7.1.0" ]
+      [ package "tcpip" ~sublibs:[ "stack-direct" ] ~min:"8.1.0" ]
```
hannesm, line 167: why is this needed?

hannesm:
In general, this looks fine.

But I feel like there's a lack of why there are bounds in mirage?

The reason is that sometimes the API changes, esp. in respect to `connect` -- which is hardcoded (argument order and amount of arguments) by mirage -- the tool. This is why at some point we pushed for having lower and upper bounds: we can release updates to opam-repository, and a mirage version won't pick it up (e.g. the dns 8.0.0 changes to dns-client). This is good.

Now, when to update bounds: we can relax (allow newer versions) if there's a release that keeps the `connect` API (or however your device initialization is called). Sometimes I first look at the mirage bounds and then select on a version number for a release (i.e. the mirage-qubes release I'd have named 0.9.1, since then there wouldn't be any need for changes in the mirage tool).

When the API changes, such as in dns 8.0.0, we need to adapt the code in mirage, and also raise the lower bound. In general, I try to bind to p >= major && p < major + 1. But releases of mirage libraries and picking version numbers has been a bit random (read: no semantic versioning) -- so some bounds are tighter.

hannesm:
I added my last comment into the RELEASE.md in #1549, hope that makes it clear.

dinosaure:
LGTM.

PR #1555, opened 2024-07-12
Re-introduce deps which can be useful to track dependencies which need to be built but not used to apply a functor

dinosaure:
This PR reintroduce the `deps` argument into the `main` to be able to track a dependency which needs to be built by `dune` but we don't want to generate a module/functor from it. It's actually used by `pasteur` to track some JS files generated by `js_of_ocaml` and packed with `docteur` (to make a disk image).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+### Unreleased
+
+- BREAKING re-introduce `?deps` argument into `Mirage.main` function to be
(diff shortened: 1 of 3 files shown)


samoht:
Sound good - could you add an entry to the CHANGELOG too?

hannesm (approved):
LGTM

hannesm:
curious: should there be a release with this change? AFAICT this would unblock your pasteur unikernel.

PR #1557, opened 2024-08-06
allow tar 3.0.0 release series (no API change in tar-mirage)

hannesm:


reynir (commented):
This looks good to me. CI failure seems to be due to opam-monorepo.0.3.2 vendoring an older version of the opam libraries. They seem to have stopped doing that at some point as the duniverse directory is gone in the main branch. From the CHANGES file I don't find any useful information to help decide on a lower bound for opam-monorepo.

PR #1558, opened 2024-08-12
allow mirage-qubes 0.11 series

hannesm:
for https://github.com/ocaml/opam-repository/pull/26350

### DIFF
--- a/lib/devices/ip.ml
+++ b/lib/devices/ip.ml
@@ -85,7 +85,7 @@ type ipv6_config = {
 (** Types for IP manual configuration. *)
 
 let ipv4_qubes_conf =
-  let packages = [ package ~min:"0.9.0" ~max:"0.11.0" "mirage-qubes-ipv4" ] in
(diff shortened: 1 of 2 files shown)


palainp:
Thank you @hannesm I missed that mirage needs that. I would have done better to set 0.10.1 :(
So LGTM!

hannesm:
hmm, it is a bit inconvenient... maybe we should bump to a 1.0 and apply semantic versioning (and then say in mirage to use >= 1 & < 2) -- and then we'll have the freedom and convenience of what everyone thinks about versioning :)

hannesm:
and at the same time, I think our qubes unikernels don't use the target qubes -> so we actually don't need to worry much ;)

the remaining question is whether the qubes target is worth to have and keep, if all our qubes unikernels don't use it (since we need to control tightly the network configuration etc.).

palainp:
I would prefer to keep it for unikernels that don't need downstream links (like a dns-resolver I have) :)

PR #1564, opened 2024-10-01
opam src url: path includes leading slash

### LINKED ISSUE #1563: opam file: An extra `/` in the beginning of the path of http(s) URIs
This introduces an extra `/` in the beginning of the path of http(s) URIs. I'll see if I can come up with a fix.

_Originally posted by @reynir in https://github.com/mirage/mirage/issues/1268#issuecomment-2385578051_

reynir:
It seems `ssh://` urls are relative when they otherwise look absolute, see e.g. https://github.com/ocaml/opam/issues/3382. I guess an absolute path is `ssh://user@host//absolute/path` i.e. a double leading slash.

Closes #1563.

Draft PR while I look into how we can test this.

### DIFF
--- a/lib/functoria/opam.ml
+++ b/lib/functoria/opam.ml
@@ -61,6 +61,7 @@ module Endpoint = struct
         Astring.String.cut ~sep:":" (String.sub str consumed (len - consumed))
       with
       | Some ("", path) ->
+          let path = "/" ^ path in
           let local =
             List.map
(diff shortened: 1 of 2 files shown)


reynir:
The test, which took quite a few times to get working with the different versions of git, shows the expected output, and it fails before the latest commit. Unfortunately, CI isn't run on the first commit.

hannesm:
thanks!

PR #1566, opened 2024-10-01
functoria/version check: in case our_version is not parseable, emit a…

hannesm:
… warning but continue operation

//cc @reynir


reynir:
Hrm, still some corner cases! First it seemed to work:
```
reynir@solsort:~/workspace/qubes-miragevpn$ mirage configure -t xen
reynir@solsort:~/workspace/qubes-miragevpn$ mirage --version
9b6fcef
```
No errors, but no warning! But look at that version string. I had removed the `(* mirage >= 4.7.0 & < 4.8.0 *)` line in config.ml, so I added it back and got this:
```
reynir@solsort:~/workspace/qubes-miragevpn$ mirage configure -t xen
Version mismatch: required is mirage < 4.8.0, but 9b6fcef is installed. Please downgrade your installation (opam update; opam install 'mirage<4.8.0')
```
So since that commit checksum happened to start with a digit it got parsed as a major version 9! Which is in the year 2038 and Mirage survived the Epochalypse.

hannesm:
thanks @reynir, what about 61083323 ? :)

reynir:
```
reynir@solsort:~/workspace/qubes-miragevpn$ mirage configure -t xen
mirage: [WARNING] Skipping version check, since our_version ("17e5015") fails to parse: only digits and . allowed in version
```
:partying_face: 

(this might have unintended consequences if we ever want to release `mirage.9.0.0~alpha1` or so I guess)

reynir (approved):

hannesm:
the consequences are leading to a warning message, and should not be present with released mirage tools -- and tbh if you're using a pinned version, it's your choice and the version checking code isn't useful....

TL;DR the version check code is meant as a UX improvement for people using mirage in a released version. :)

PR #1568, opened 2024-10-02
config.ml: Make deprecation warnings non-errors

### LINKED ISSUE #1553: Warnings in config.ml
Hey,

when there's a deprecation warning in `config.ml` (such as using `etif` with mirage 4.6.0), this results in the following during mirage configure:

```
318.213s:+ mirage configure -t xen --context ./mirage/context --no-extra-repo
318.583s:File "config.ml", line 7, characters 15-19:
318.583s:7 | let ethernet = etif default_network
318.583s:                   ^^^^
318.583s:Error (alert deprecated): Mirage.etif
318.583s:Deprecated. Use [ethif] instead.
```

So, any warning is an error. It would be great if this would just be a warning, and there'd be a "strict" (or `-warn-error`) mode for `mirage configure`). I don't understand these parts of the code well. Maybe it is simply to put `--profile release` at the right spot.

Obviously, the question rises what should be the default? I guess for unreleased mirage versions, the development profile, for released versions the release profile.

reynir:
Also add a test that shows the deprecation warning as a non-error through Mirage.etif.

Unfortunately, the warning seems to be printed twice for some reason.

Fixes #1553


reynir:
Besides lint-fmt failing debian-12-5.2_arm32 fails because the warning is only printed **once** as the only configuration.

hannesm:
I guess once for bytecode compilation, once for native.. and 5.x doesn't do native code for 32 bit anymore.

I'm not yet convinced of this PR.... what about all the other errors/warnings - why is deprecation special? shouldn't there similar to dune be a release and a development profile?

hannesm:
> Unfortunately, the warning seems to be printed twice for some reason.

I can see two paths forward: disable tests on 32 bit platforms and ocaml 5 (or mark mirage as available only on 64 bit platforms -- I think it has been some time since someone tried it on 32bit). The alternative is to use only bytecode (or only native code). I somehow have a preference to only compile to native code (can be done via some dune runes) - but we can leave it as is as well.

hannesm:
I like this PR! :)

reynir:
I tried to play around with `(modes ...)` in the dune file. It seems it would be ideal if we could use `best`, but playing around I couldn't figure out how it works.

https://dune.readthedocs.io/en/stable/reference/dune/executable.html#linking-modes

reynir:
Rebased on main

hannesm:
nice PR, but I'd appreciate if we could get our ocaml-ci to have a green checkmark... as mentioned, we can disable testing on 32 bit -- but here is some 4.14 vs 5.1 change, where newlines and spaces got introduced... I'm unsure how the expect tests are working, but maybe there can be different expect files given for different compiler versions?

reynir:
After some thought I redirected stderr to /dev/null in the test. After all, we are not too interested in the exact warnings as such, just that the warnings are not treated as errors. It would be nice if we could keep the stderr output **iff** the exit code is non-zero. For example if `etif` is removed in the future it would be easier to update the test with stderr present.

hannesm (approved):
LGTM, thanks a lot

PR #1574, opened 2024-10-07
provide Mirage_runtime.register_arg, deprecate Mirage_runtime.register

hannesm:
The name `Mirage_runtime.register` is slightly too generic in my opinion.

### DIFF
--- a/lib_runtime/functoria/functoria_runtime.mli
+++ b/lib_runtime/functoria/functoria_runtime.mli
@@ -37,14 +37,16 @@ end
 
 val register : 'a Cmdliner.Term.t -> unit -> 'a
 [@@ocaml.deprecated "Use register_arg instead."]
-(** [register t] registers the Cmdliner term [k] as a runtime key and return a
-    callback [f] that evaluates to [t]s' value passed on the command-line.
+(** [register t] registers the Cmdliner term [k] as a runtime argument and
+    return a callback [f] that evaluates to [t]s' value passed on the
+    command-line.
 
     [f] will raise [Invalid_argument] if called before cmdliner's evaluation. *)
 
 val register_arg : 'a Cmdliner.Term.t -> unit -> 'a
-(** [register_arg t] registers the Cmdliner term [k] as a runtime key and return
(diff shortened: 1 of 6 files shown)


lib_runtime/functoria/functoria_runtime.mli
```diff
@@ -36,11 +36,18 @@ module Arg : sig
...
 (** [register t] registers the Cmdliner term [k] as a runtime key and return a
     callback [f] that evaluates to [t]s' value passed on the command-line.
 
     [f] will raise [Invalid_argument] if called before cmdliner's evaluation. *)
 
+val register_arg : 'a Cmdliner.Term.t -> unit -> 'a
+(** [register_arg t] registers the Cmdliner term [k] as a runtime key and return
+    a callback [f] that evaluates to [t]s' value passed on the command-line.
+
+    [f] will raise [Invalid_argument] if called before cmdliner's evaluation. *)
```
reynir, line 49: In the doc string we refer to it as (runtime) key, but we call the function `register_arg`. Should we maybe rename it to `register_key` or update the documentation?

reynir:
While reviewing I notice that `Functoria_runtime.Arg` seems to be entirely unused outside of its parent module (and seems useless, too). Maybe we should just remove it from the .mli?

lib_runtime/functoria/functoria_runtime.mli
hannesm: I guess we should decide on terminology ;)

I see in the changes entry for 4.5.0 the "configuration key" and "arguments at runtime".

Now, the module is called Key and Runtime_arg. I guess we should stick to that, then.

I find it very subtle, is it `key`, `parameter`, `argument`? Please also note that the docstring was only copied from the `register` function. But I'll adapt to "argument" to stay consistent.

hannesm:
> While reviewing I notice that `Functoria_runtime.Arg` seems to be entirely unused outside of its parent module (and seems useless, too). Maybe we should just remove it from the .mli?

sounds like a good idea -- maybe best for another PR?

reynir (approved):
Looks good to me! Thanks

PR #1577, opened 2024-10-07
keys for DNS/HE/GIT devices

hannesm:
works with https://github.com/mirage/mirage-skeleton/pull/403

### DIFF
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -30,7 +30,7 @@ let generic_dns_client ?group ?timeout ?nameservers ?cache_size () =
         cache_size;
       ] ->
         code ~pos:__POS__
-          {ocaml|%s.connect @[?nameserver:%s ?timeout:%s ?cache_size:%s@ (%s, %s)@]|ocaml}
+          {ocaml|%s.connect @[?nameservers:%s ?timeout:%s ?cache_size:%s@ (%s, %s)@]|ocaml}
           modname nameservers timeout cache_size stackv4v6 happy_eyeballs
     | _ -> connect_err "generic_dns_client" 8
   in
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -11,28 +11,28 @@ type dns_client = Dns_client
 
 let dns_client = typ Dns_client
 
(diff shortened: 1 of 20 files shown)


hannesm:
To explain the reasoning behind this: some devices have optional arguments -- let's take happy-eyeballs as example, where you can configure the timeout for establishing a connection.

Previously, our interface allowed some unikernel developer to optionally pass a key (of type int64 option) for that -- meaning they could decide whether the resulting unikernel should have `--he-connect-timeout` or not.

Now, we generate this key always -- so all unikernels using happy eyeballs will be configurable, automaticatlly \o/. In addition, if you have two network stacks and a happy eyeballs device on each, you'll be able to have two sets of keys -- `--service-connect-timeout` and `--management-connect-timeout`.

In the end, unikernels are more predictable and developers need to do less. They can still provide their own custom value for the connect_timeout (via the optional argument), but don't need to build keys for doing so. Below the line, our unikernels using git no longer need to expose all the different arguments (ssh_key/ssh_password/ssh_authenticator/http_authenticator/http_headers).

This PR adds some code to mirage, and I'm sure one day we'll be able to condense the code again. But for the time being, the runtime arguments used by mirage devices are declared in that way - and are only of basic types (to limit the dependency cone of mirage-runtime).

Reviews welcome. I'm aware this needs some adjustments in mirage-skeleton that I'll work on soon.

hannesm:
So, CI is happy. Feedback would be highly welcome :)

reynir:
I like the motivation behind this. I briefly looked over the code, but not in depth. I think this is nice.

dinosaure:
Seems good to me, I never used `group` argument but the rational behind seems correct as you described :+1:. I don't have too much time to look on it however.

hannesm:
> Seems good to me, I never used `group` argument but the rational behind seems correct as you described 👍. I don't have too much time to look on it however.

To get a better understanding, the `group` is nice and useful for e.g. network stacks -- considering we have two (in a VPN unikernel, or when doing monitoring) -- here we have the arguments `--ipv4=` and `--management-ipv4=` -- where the first doesn't use `group`, the second `group:"management"`.

PR #1578, opened 2024-10-09
Mirage.register: document ?src argument

reynir:


hannesm:
thanks for your PR. could you run a `dune bu @fmt --auto`? thanks!

hannesm:
thanks

PR #1579, opened 2024-10-09
Defunctorise TIME, PCLOCK, MCLOCK, RANDOM -- allow defaults to be passed to Mirage.register

hannesm:
Now, the `register` function takes timer, clock, time, random devices: in case you want to have something custom, go for it.

Also, take the opportunity for renamings:
- TIMER used to be TIME (the mirage-timer package will have the sleep_ns function)
- TIME used to be MCLOCK (the mirage-time package will carry the monotonic wall-clock interface)
- CLOCK used to be PCLOCK (the mirage-clock package will have the posix clock interface)

Strip these functors away from the other devices.

Please note, this is my most recent proposal of #1513: we will use dune variants in timer, clock, time. A lot of MirageOS devices will need some updates (removing functor arguments).

This PR is meant as discussion base for how the new world could look like :)

~~If there's some functoria pro, I'd really like to get some hint on how to avoid "Obj.magic" in Mirage.register. Thanks for reading :)~~

A great starting point for feedback is https://github.com/mirage/mirage/pull/1579/files#diff-79c838533b44a0761156ff031b0018fcc5f0e3837463a67484468a19c073756e

(TODO: due to version constraints I plan to rename mirage-crypto-rng-mirage (not yet sure what the new name should be, maybe mirage-crypto-rng-mirage2))


hannesm:
this is outdated, let's continue in #1599

PR #1580, opened 2024-10-10
OCaml runtime arguments are now in cmdliner-stdlib

hannesm:
A special entry function, `configure_ocaml_runtime`, is defined in Mirage_runtime and run by mirage.ml.

This reduces the code generated in main.ml drastically.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -397,100 +397,17 @@ end
 include Lib.Make (Project)
 module Tool = Tool.Make (Project)
 
-let backtrace =
-  let runtime_args = Runtime_arg.[ v backtrace ] in
-  let connect _ _ = function
-    | [ backtrace ] ->
-        code ~pos:__POS__ "return (Printexc.record_backtrace %s)" backtrace
-    | _ -> Misc.connect_err "backtrace" 1
-  in
-  impl ~runtime_args ~connect "Printexc" job
-
-let randomize_hashtables =
-  let runtime_args = Runtime_arg.[ v randomize_hashtables ] in
-  let connect _ _ = function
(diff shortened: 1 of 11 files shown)


samoht:
This somehow seems redundant with using https://github.com/mirage/cmdliner-stdlib? I don't have a strong opinion on where this should live..

hannesm:
Thanks for pointing to that repository, I was not aware of its existence. I don't mind either where the code should live, if cmdliner-stdlib is being used elsewhere, let's use that.

I somehow prefer the cmdliner-stdlib, the main reason is that the defaults to the arguments are taken from `Gc.get` and are not hardcoded. I'll revise this PR to use cmdliner-stdlib.

hannesm:
updated, and with https://github.com/mirage/cmdliner-stdlib/pull/2 we're in the same shape. WDYT?

lib/mirage.ml
```diff
@@ -397,101 +397,24 @@ end
...
     ~runtime_modname:"Mirage_runtime" argv
 
+let ocaml_runtime =
+  let packages = [ package ~min:"1.0.0" ~max:"2.0.0" "cmdliner-stdlib" ] in
+  let runtime_args =
+    [
+      Runtime_arg.v
+        (Runtime_arg.create ~pos:__POS__
+           "Cmdliner_stdlib.setup ~backtrace:(Some true) \
+            ~randomize_hashtables:(Some true) ()");
```
hannesm, line 413: the reason for these two defaults is since MirageOS explicitly sets those to `true` -- and in this PR I didn't want to change behaviour.

samoht:
> +163 −645
LGTM!

PR #1582, opened 2024-10-10
Make config.ml and unikernel.ml co-buildable 

samoht:
Revive #1515


lib/mirage.ml
```diff
@@ -380,7 +380,7 @@ let run t = %s.Main.run t ; exit 0|ocaml}
         [
           package ~scope:`Monorepo "lwt";
           package ~scope:`Monorepo ~min ~max "mirage-runtime";
-          package ~scope:`Switch ~build:true ~min ~max "mirage";
+          package ~scope:`Monorepo ~build:true ~min ~max "mirage";
```
hannesm, line 383: I think we need both scopes.

hannesm, line 383:
```suggestion
          package ~scope:`Monorepo ~build:true ~min ~max "mirage";
          package ~scope:`Switch ~build:true ~min ~max "mirage";
```

samoht: I'm not sure what's the semantic of adding two scopes here..

hannesm: I'm also not sure what the semantic of adding two scopes is.

The issue is, for opam file generation - we generate the "switch" dependencies (and use them for the reproducible builds, https://github.com/robur-coop/orb). Now, if we move "mirage" to be a "monorepo" dependency, orb will no longer install "mirage" in the switch, and the build will fail.

hannesm: Is there anything done inside functoria that uses these scopes? Or is it solely for tagging "monorepo" for opam-monorepo invocation (via makefile, where it defines the monorepo "scope")?

PR #1584, opened 2024-10-11
Add a call to Printexc.record_backtrace

reynir:
Runtime argument getters may be called before `Printexc.record_backtrace` is called - therefore we call it just before raising in order to get a nice backtrace.

For context see https://github.com/mirage/mirage-skeleton/pull/405#discussion_r1797009524

### DIFF
--- a/lib_runtime/functoria/functoria_runtime.ml
+++ b/lib_runtime/functoria/functoria_runtime.ml
@@ -25,6 +25,8 @@ module Arg = struct
   let get t =
     match t.value with
     | None ->
+        (* When used incorrectly this may raise before we otherwise turn on back traces *)
+        Printexc.record_backtrace true;
         invalid_arg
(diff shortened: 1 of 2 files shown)


hannesm:
so, I suggest to move the code to the toplevel in Mirage_runtime. :)

reynir:
It would be interesting if you could somehow make sure that the binary is run in a way like `OCAMLRUNPARAM=b`, but I don't know how to achieve that in general.

hannesm:
I manually verified this does what I expected:

```
Fatal error: exception Invalid_argument("Called too early. Please delay this call to after the start function of the unikernel.")
Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
Called from Dune__exe__Unikernel.hello in file "unikernel.ml", line 6, characters 2-76
Solo5: solo5_exit(2) called
```

PR #1585, opened 2024-10-11
add parens around the return value of register/register_arg

hannesm:
and disable ocamlformat here locally

cc @reynir


hannesm:
@reynir I added a comment, I think this is good to go now!?

PR #1586, opened 2024-10-14
call Printexc.record_backtrace early -- will be later overwritten by --backtrace (if provided)

hannesm:
inspired by #1584 //cc @reynir

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -333,6 +333,7 @@ module Project = struct
       {ocaml|open Lwt.Infix
 type 'a io = 'a Lwt.t
 let return = Lwt.return
+Printexc.record_backtrace true
 let run t = %s.Main.run t ; exit 0|ocaml}
       (os_of_target info)
 
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -332,8 +332,8 @@ module Project = struct
     Fmt.str
       {ocaml|open Lwt.Infix
 type 'a io = 'a Lwt.t
-let return = Lwt.return
-Printexc.record_backtrace true
+let return = Lwt.return;;
+Printexc.record_backtrace true;;
 let run t = %s.Main.run t ; exit 0|ocaml}
       (os_of_target info)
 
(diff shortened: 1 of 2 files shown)


reynir:
I think your force push broke the output -- the semi colon is needed I think -- maybe even ;;

hannesm:
Hmm hmm, but with a semicolon it broke as well. Thanks for your comment, I'll figure out what is needed.

hannesm:
see 80693f21

reynir:
I'm afraid this might still call `Printexc.record_backtrace` too late. I have been working on writing a test, but it is very challenging because of opam-monorepo. I may resort to a manual test...

hannesm:
> I'm afraid this might still call `Printexc.record_backtrace` too late. I have been working on writing a test, but it is very challenging because of opam-monorepo. I may resort to a manual test...

Really? I thought calling it at the top of the module as a top-level exprssion will be early enough...

reynir:
My thinking is this:

* A call to `Mirage_runtime.register_arg` in unikernel.ml could be at the top level (if used "incorrectly").
* So the evaluation order depends on what order the modules are linked
* My thinking is main.ml depends on unikernel.ml, so unikernel.ml is linked earlier and thus evaluated earlier.

I am creating a new switch in order to pin mirage[-runtime] to this branch in order to do a manual test. I will report back.

hannesm:
Right, you're absolutely right. So, the place is either your PR or in mirage-runtime at top-level (uncondtionally)!?

reynir:
So indeed:
```
$ ./dist/early-arg 
Fatal error: exception Invalid_argument("Called too early. Please delay this call to after the start function of the unikernel.")
```
(it is annoying I can't figure out how to add this as a cram test /o\)

> So, the place is either your PR or in mirage-runtime at top-level (uncondtionally)!?

Right, I think so. I guess it's maybe fine to do at the top-level unconditionally?? As the unikernel can/will set/unset record_backtrace using the runtime argument. I guess Mirage_runtime is always linked before both unikernel.ml and main.ml.

hannesm:
closing, let's use #1584 -- it should be somewhere in the Mirage_runtime module.

PR #1589, opened 2024-10-30
Upgrade to OCamlformat 0.27.0

Julow:
The aim of this preview is to gather feedback.

Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md


test/functoria/test_cli.ml
```diff
@@ -32,25 +32,25 @@ let test_configure () =
...
-           args =
-             {
-               context = (true, false);
-               output = None;
-               config_file = Fpath.v "config.ml";
-               context_file = None;
-               dry_run = false;
-             };
-         }))
+       (Cli.Configure
```
samoht, line 35: having 7 spaces here is a bit weird - why is this needed?

lib/functoria/cli.mli
```diff
@@ -140,4 +140,5 @@ type 'a result =
     also carry global command-line parameters. *)
 
 val peek : ?with_setup:bool -> mname:string -> string array -> unit result
-(** [peek] is the same as {!val:eval} but without failing on unknown arguments. *)
+(** [peek] is the same as {!val:eval} but without failing on unknown arguments.
+*)
```
samoht, line 144: is this due to a bug in accounting for 80-chars wide lines?

lib/functoria/impl.ml
```diff
@@ -185,8 +185,7 @@ let rec equal : type t1 t2. t1 t -> t2 t -> (t1, t2) Typeid.witness =
 
 and equal_abstract (Abstract x) (Abstract y) = Typeid.to_bool @@ equal x y
 
-and equal_tl :
-    type t1 t2 v1 v2.
+and equal_tl : type t1 t2 v1 v2.
```
samoht, line 188: Why is this needed?

samoht (commented):
The changes in the doc strings seem to be related to bug fixes so I guess they are fine? - I'm not sure I like the new way it's printed though -- ie. why does `*)` isn't aligned with the opening parenthesis as it's used to be for non-empty lines. 

I'm not sure to understand why the other changes are needed. They don't seem to fix any previous consistency issue.

test/functoria/test_cli.ml
Julow: This is to make the indentation identical for polymorphic and normal variant arguments:

```ocaml
    (Ok
       (...))
    (`Ok
       (...))
```

lib/functoria/cli.mli
Julow: Yes, the `*)` was previously breaking the margin.
By aligned with the opening parenthesis, do you mean this ?
```
(** ...
 *)
```
This strange formatting only appears when the comment is just long enough to hit the margin but do not start to wrap.
In other cases, the `*)` is placed at the end of the last paragraph:
```
(** Every other comments
    are like this *)
```

lib/functoria/impl.ml
Julow: It's only a cosmetic change. These can make the type annotation harder to read when it all fits on one line and the `and equal_tl :` line is always short when the type is long enough.

Julow:
Thanks for the review!

lib/functoria/cli.mli
Julow: The formatting of regular comments is slightly different, if the `*)` exceed the margin, the last word of the paragraph wraps with it:
```ocaml
(* *[peek] is the same as {!val:eval} but without failing on unknown
   arguments. *)
```
Do you think that's better than breaking only the `*)` ?

Julow:
OCamlformat 0.27.0 as been released :tada: This is ready to be merged.

Julow:
Thanks for your feedback. I'll close this PR as it's no longer necessary.

PR #1591, opened 2024-11-05
revise syslog, add monitoring

hannesm:
this removes the type syslog_config, and introduces boot parameters for your syslog configuration

also, mirage-monitoring is integrated into the possible mirage devices

----

The goal is to reduce the amount of code in `config.ml` across the unikernels that we use. Now, the approach to provide boot parameters for configuration instead of hardcoded values is now applied to syslog as well.

Bits to be upstreamed are:

```OCaml
let enable_monitoring =
  let doc = Key.Arg.info
      ~doc:"Enable monitoring (syslog, metrics to influx, log level, statmemprof tracing)"
      [ "enable-monitoring" ]
  in
  Key.(create "enable-monitoring" Arg.(flag doc))

let management_stack =
  if_impl
    (Key.value enable_monitoring)
    (generic_stackv4v6 ~group:"management" (netif ~group:"management" "management"))
    stack

let optional_monitoring time pclock stack =
  if_impl (Key.value enable_monitoring)
    (monitoring $ time $ pclock $ stack)
    noop

let optional_syslog pclock stack =
  if_impl (Key.value enable_monitoring)
    (syslog $ pclock $ stack)
    noop

let () =
  register ...
    [
      optional_syslog default_posix_clock management_stack ;
      optional_monitoring default_time default_posix_clock management_stack ;
      ...
    ]
```

But this requires some more thoughts. I guess a `--enable-monitoring` configuration time parameter would be sensible. The question remains whether we should hardcode syslog_udp, or provide as well a configuration parameter (or embed UDP and TCP support into the unikernel anyways). For syslog-tls, I don't really have a usecase (I use separate network devices for service and management, as seen above).

WDYT?

### DIFF
--- a/lib/devices/runtime_arg.ml
+++ b/lib/devices/runtime_arg.ml
@@ -179,17 +179,25 @@ let syslog ?group ?docs default =
   runtime_network_key ~pos:__POS__ "syslog %a%a%a" pp_group group pp_docs docs
     (pp_option pp_ipaddr) default
 
-let monitor ?group ?docs default =
-  runtime_network_key ~pos:__POS__ "monitor %a%a%a" pp_group group pp_docs docs
-    (pp_option pp_ipaddr) default
-
 let syslog_port ?group ?docs default =
   runtime_network_key ~pos:__POS__ "syslog_port %a%a%a" pp_group group pp_docs
     docs (pp_option Fmt.int) default
 
-let syslog_hostname ?group ?docs default =
-  runtime_network_key ~pos:__POS__ "syslog_hostname %a%a%S" pp_group group
-    pp_docs docs default
+let syslog_truncate ?group ?docs default =
+  runtime_network_key ~pos:__POS__ "syslog_truncate %a%a%a" pp_group group
+    pp_docs docs (pp_option Fmt.int) default
+
+let syslog_keyname ?group ?docs default =
+  runtime_network_key ~pos:__POS__ "syslog_keyname %a%a%a" pp_group group
+    pp_docs docs (pp_option Fmt.string) default
+
+let monitor_hostname ?group ?docs () =
+  runtime_network_key ~pos:__POS__ "monitor_hostname %a%a ()" pp_group group
+    pp_docs docs
+
+let monitor ?group ?docs default =
+  runtime_network_key ~pos:__POS__ "monitor %a%a%a" pp_group group pp_docs docs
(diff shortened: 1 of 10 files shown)


hannesm:
Having commented above on what else could be upstreamed, this PR is self-contained enough and useful if merged. Don't hesitate to review and merge it.

hannesm:
I'm struggling with the CI error - and why suddenly we see:
```
$ mirage configure -t unix 2>/dev/null
+  [1]
```

Full output:

```
File "test/mirage/warn-etif/run.t", line 1, characters 0-0:
/usr/bin/git --no-pager diff --no-index --color=always -u _build/.sandbox/d7b77c87dedadde08613b3d70ed0a673/default/test/mirage/warn-etif/run.t _build/.sandbox/d7b77c87dedadde08613b3d70ed0a673/default/test/mirage/warn-etif/run.t.corrected
diff --git a/_build/.sandbox/d7b77c87dedadde08613b3d70ed0a673/default/test/mirage/warn-etif/run.t b/_build/.sandbox/d7b77c87dedadde08613b3d70ed0a673/default/test/mirage/warn-etif/run.t.corrected
index ed7b586..56ba5d8 100644
--- a/_build/.sandbox/d7b77c87dedadde08613b3d70ed0a673/default/test/mirage/warn-etif/run.t
+++ b/_build/.sandbox/d7b77c87dedadde08613b3d70ed0a673/default/test/mirage/warn-etif/run.t.corrected
@@ -1,6 +1,7 @@
Configure the project for Unix:


$ mirage configure -t unix 2>/dev/null
+  [1]


The command succeeded and created a number of files. It produced a load of
warnings, but matching on the warnings given different compiler versions with
@@ -9,13 +10,9 @@ ever slightly different output is not worth it.
$ ls -a
.
..
-  Makefile
_build
config.ml
-  dist
dune
dune-project
-  dune-workspace
-  dune.build
dune.config
mirage
```

Any ideas?

hannesm:
Found the issue of my last comment: the test refered to the `syslog_config` that this PR removes. Pushed a commit that fixes the test (since stderr is redirected to /dev/null, both errors and warnings weren't observable).

palainp (approved):
Thank you @hannesm , I really approve the goal of reducing `config.ml` files :)
I currently don't use management interfaces for my unikernels, but this PR LGTM!

Probably due to recent #1593 the tests line numbers are wrong now :(

PR #1592, opened 2024-11-05
Emit output what to do next. This fixes #921

### LINKED ISSUE #921: output "what to do next"
in our today's meeting, @samoht mentioned that `mirage` doesn't tell you what to do next, which would be nice:
- `mirage configure` --> `now build your unikernel using mirage build (or make/make depend)`
- `mirage build` --> `now run your unikernel using ./solo5-vt ... `

hannesm:
It gives some hints to the user what they should do next.

The only thing slightly worrying is that the cram test cases have the log message two times - any hints why this is the case?

----

We could after the `build` improve depending on the target, if we would have access to the target. But I have a tough time to figure out how to pass this information through.

So, this is ready for review :D

### DIFF
--- a/lib/devices/target.ml
+++ b/lib/devices/target.ml
@@ -398,15 +398,19 @@ let dune i =
   let (module Target) = choose target in
   Target.dune i
 
+let output_message = ref true
+
 let configure i =
   let open Action.Infix in
   let target = Info.get i Key.target in
   let (module Target) = choose target in
   Target.configure i >|= fun () ->
-  Logs.app (fun m ->
-      m
-        "Successfully configured the unikernel. Now run 'make' (or more \
-         fine-grained steps: 'make all', 'make depends', or 'make lock').")
+  if !output_message then (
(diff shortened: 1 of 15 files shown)


palainp (approved):
Dear @hannesm , thank you for that PR it should be helpful for newcomers :)
I tried (but failed) to find why we call configure two times.

The first one seems to comes from `let* () = Engine.configure i jobs in` in lib.ml that calls `let* () = Device.configure dev info in` in engine.ml which calls configure that prints the message.

I however have not found yet the second caller. Maybe it can be possible to add an optional do_not_print_msg defaulted to false, and call with true the first time?

hannesm:
rebased, fixed, and opened #1594 for the future. NB: getting a callstack is pretty easy:

```OCaml
  let bt = Printexc.get_callstack 20 in
  Logs.info (fun m -> m "backtrace: %s" (Printexc.raw_backtrace_to_string bt));
```

PR #1593, opened 2024-11-15
add DHCP device

palainp:
Dear developers,

While working with the mirage-nat example, I tried to configure the public interface with DHCP, but it failed because `ipv4_of_dhcp` is not currently exported. This PR aims to add it.

`dune runtest` works fine, ~~except for the message `Error (warning 32 [unused-value-declaration]): unused value ipv4_of_dhcp.` but I don't know where I should use it :( ~~ EDIT: It was a redundant declaration, I also fixed line numbers in the tests with the additional line.

I'd be happy to change anything that isn't good enough :)

Best,

### DIFF
--- a/lib/devices/ip.mli
+++ b/lib/devices/ip.mli
@@ -29,15 +29,6 @@ type ipv6_config = {
   gateway : Ipaddr.V6.t option;
 }
 
-val ipv4_of_dhcp :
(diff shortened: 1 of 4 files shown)


hannesm:
looks fine, thanks!

PR #1598, opened 2025-01-09
Update README.md

### LINKED ISSUE #1597: [Question] Official website URL on Github
Hi, would like to confirm on the official site that is attached on Github, when I try to click the URL it gives me this warning from browser

image

Yet on the README.md, the official site is pointed to https://mirage.io

https://github.com/mirage/mirage/blob/3b7e649f29a9b803465b6b0dba074e2fe29f1d22/README.md?plain=1#L37-L39

I tried to ignore the warning and it redirects me to the same website as https://mirage.io, should we just update the link to https://mirage.io instead?

image

samoht:
Fix #1597

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <div align="center">
-  <a href="https://mirage.io">
+  <a href="https://mirageos.org">
     <img src="./logo.svg" alt="MirageOS logo"/>
   </a>
   <br />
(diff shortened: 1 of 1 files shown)


hannesm:
So, there's a deployment issue.. I can see the bad certificate as well.

```
Websites prove their identity via certificates. Firefox does not trust this site because it uses a certificate that is not valid for mirageos.org. The certificate is only valid for the following names: mirage.io, next.mirage.io
 
Error code: SSL_ERROR_BAD_CERT_DOMAIN
 
View Certificate
```

I think @Firobe deployed the OCaml 5 and some diversion -- it seems one of the unikernels has a bad certificate -- meaning not valid for both mirage.io and mirageos.org.

Firobe:
I turned off the diversion for now in case it has anything to do with it, I'll try to look at it further in the next days (it may be the case that a change in `mirage-redeploy` broke this, a revert might help)

PR #1605, opened 2025-05-14
Fix Mirage_runtime documentation headers

reynir:
It seems headers require a space after the comment, e.g.:

    (** {2 Header} *)

    let _ = ...

instead of:

    (** {2 Header} *)
    let _ = ...

The last two functions also didn't belong to the "exit codes" header, so add a more meaningful header for those.


reynir:
Cf. https://ocaml.org/p/mirage-runtime/latest/doc/Mirage_runtime/index.html where `Blocks` and `Startup delay` are sub headings under `Log threshold` and the `Command-line converters` and `Manpage sections` headers are missing.

hannesm:
looks fine, thanks

PR #1607, opened 2025-07-02
Add support for the Unikraft backend targets

shym:
This PR is part of a set of related PRs to add support for Unikraft backends in the following projects:

- [mirage],
- [mirage-mtime](https://github.com/mirage/mirage-mtime),
- [mirage-ptime](https://github.com/mirage/mirage-ptime),
- [mirage-sleep](https://github.com/mirage/mirage-sleep).

[mirage]: https://github.com/mirage/mirage

Those PRs are joint work with @fabbing and @Firobe.
They are best reviewed together, in particular to be able to test them.

## Overview of this PR

This PR adds support for the Unikraft backend in the `mirage` tool itself, now that OCaml/Unikraft (`ocaml-unikraft*`) and Mirage/Unikraft (`mirage-{,net-,block-}unikraft`) packages are available in the opam repository.
The main change it brings, besides adding Unikraft targets at every location where all targets are listed, is the `Unikraft` target module, which is highly similar to the `Solo5` one (without the added complexity of the manifest). It might be helpful to check out the difference between the two modules to review this part[^1].

[^1]: For instance, using `bash` syntax, by `git diff <(awk '/^module Solo5/,/^end/' lib/devices/target.ml) <(awk '/^module Unikraft/,/^end/' lib/devices/target.ml)`...

## How to test the set of related PRs

First of all, these PRs build on previous work recently released on the opam repository, in particular:

- the [OCaml/Unikraft packages](https://github.com/ocaml/opam-repository/pull/27856),
- the [`mirage-{,block-,net-}unikraft` packages](https://github.com/ocaml/opam-repository/pull/28089),
- and the latest version of opam-monorepo, [`0.4.3`](https://github.com/ocaml/opam-repository/pull/28101), which fixes an issue triggered by the `ocaml-unikraft` package.

So to test these PRs, you will need to `opam update && opam upgrade opam-monorepo` before building any unikernel.

We propose an opam overlay repository to make it easy to test the 4 PRs together, so you can start by adding it:

```
$ opam repository add unikraft-overlays https://github.com/Firobe/mirage-unikraft-overlays.git
```

(Note that this overlay repository is being used currently in the [CI tests for `mirage/ocaml-unikraft`] to check that it can build and run the network skeleton unikernel.)

[CI tests for `mirage/ocaml-unikraft`]: https://github.com/mirage/ocaml-unikraft/blob/c1e92469af94981ac64af3d91d94b9168338040f/.github/workflows/build.yml#L115

With this overlay repository set up and `opam-monorepo.0.4.3` installed, you’re ready to build your unikernel.
You can install the OCaml/Unikraft cross compiler and backend of your choice first, if you’d rather not let the opam solver pick the first one that fits. Note that both QEMU and Firecracker backends can be installed simultaneously, and so do the two (`arm64` and `x86_64`) OCaml cross compilers, but there can be only one _default_ architecture (namely the architecture that will targeted when using the `unikraft` toolchain, in the ocamlfind meaning of “toolchain”).

```
$ opam install ocaml-unikraft-default-x86_64 ocaml-unikraft-backend-qemu
  # or ...-arm64 ...-firecracker or ...
```

You’ll obviously also need to install this PR’s version of `mirage`:

```
$ opam install mirage
```

Once `mirage` is installed, you can build your unikernel as usual:

```
$ mirage configure -t unikraft-qemu # or unikraft-firecracker
$ make
```

Then, the way to run it will obviously depends on the VMM you’ve chosen. `ocaml-unikraft` contains a couple of examples to help you get started:

-   the [examples] can be tested in `verbose` mode to see how the VMM is launched
    ```
    $ (cd examples/all/ && UNIKRAFTBACKEND=qemu dune runtest --verbose)
     # or UNIKRAFTBACKEND=firecracker ...
    ```
    but those examples are simple enough that they don’t need to set up a real network stack (in particular, the Firecracker configuration that `generator.ml` creates contains no network interface),

-   the test of the [network skeleton unikernel in CI] starts by setting up the required network configuration for the Linux kernel and ends by [launching QEMU] with quite a few options, in particular to set this other half of the network configuration up… Also note that KVM is not used in the CI but you’ll probably want to use it for speed if the host and target architectures are identical.

[examples]: https://github.com/mirage/ocaml-unikraft/tree/c1e92469af94981ac64af3d91d94b9168338040f/examples/all
[network skeleton unikernel in CI]: https://github.com/mirage/ocaml-unikraft/blob/c1e92469af94981ac64af3d91d94b9168338040f/.github/workflows/build.yml?plain=1#L159
[launching QEMU]: https://github.com/mirage/ocaml-unikraft/blob/c1e92469af94981ac64af3d91d94b9168338040f/.github/workflows/build.yml?plain=1#L178


hannesm:
thanks

Issue #1372, opened 2022-11-09
Local dune libraries are not accessible within a mirage project


kit-ty-kate:
Using mirage-skeleton's `hello` project as an example, adding a simple local dune library such as:
```
$ mkdir lib
$ echo "(library (name my_local_lib))" > lib/dune
```
then trying to declare it in mirage:
```diff
-let main = main ~packages:[ package "duration" ] "Unikernel.Hello" (time @-> job)
+let main = main ~packages:[ package "duration"; package "my_local_lib" ] "Unikernel.Hello" (time @-> job)
```
Will make opam-monorepo fail during `make depends` as `my_local_lib` is not an opam project available in opam-repository (and even if there was that would cause a conflict between the vendored `duniverse/my_local_lib` and `my_local_lib`)

Could there be an additional parameter to `Mirage.package` to declare the package as a local library?

kit-ty-kate:
If anybody else is looking for a solution, the following hack works ok for me: https://github.com/kit-ty-kate/exn.st/blob/3894fcea9c08bb61a4d771e5046fd250f31b4201/Dockerfile#L5

hannesm:
As stated on IRC, I think the way to achieve this is to run `mirage configure` in the top-level with `-f my_unikernel/config.ml`, and thereafter `dune build --root . my_unikernel/dist` which will take the local lib into account.

That's at least from my memory/experience, you may want to take a look at https://github.com/mirage/mirage-crypto/blob/main/.test-mirage.sh (executed by .e.g .cirrus.yml) that attempts to compile these unikernels as a CI job with the HEAD of mirage-crypto...

kit-ty-kate:
I think it isn’t great if the expected toolchain (`mirage configure && make depends && mirage build`) does not work and we have to go out of our way to make something that should have a simple solution, happen.

It would be nice to have an extra `?local_libs:string list` argument to `Mirage.main`, to be able to do that out of the box.

hannesm:
agreed, related to #1371

Issue #1359, opened 2022-10-03
Be able to use PPX on modules used by the `unikernel.ml`


dinosaure:
Currently, even if MirageOS 4 handles correctly PPX, they are only applied on `duniverse/` projects. The idea is to put the ability to say that some of our files used by the `unikernel.ml` need PPX. That mostly mean that the end-user should be able, via the `config.ml`, to generate `(preprocess (pps _))`.

TheLortex:
Two solutions:
- short term: in config.ml allow devices to extend the generated dune file with arbitrary content (or something specific for ppxs)
- long term: let the user write `unikernel.ml`'s dune file as a library providing the unikernel functor. This way, user has full control

kit-ty-kate:
I like the long term solution a lot. That would also fix https://github.com/mirage/mirage/issues/1372

PR #1609, opened 2025-08-02
Add a ?local_libs parameter to allows local dune libs

omegametabroccolo:
This PR allows using local dune libraries in a Mirage unikernel, as suggested in #1372.

This also (somewhat) solves #1359: the sources that need to be preprocessed can be put in a local library with its own `dune` file.


hannesm:
thanks a lot. I'll merge and soon cut a release.

PR #1611, opened 2025-09-03
add a name argument to each unikernel

hannesm:
This is a new feature: every unikernel gets the possibility to have a name! By default it is set to "no-name".

This is useful for monitoring services, etc. -- and it is the name that you're calling your unikernel. In a data center operations, of course this name could be a UUID or whatever.

Any feedback on this is welcome //cc @reynir @Firobe @PizieDust @dinosaure

### DIFF
--- a/lib_runtime/mirage_runtime.ml
+++ b/lib_runtime/mirage_runtime.ml
@@ -121,20 +121,29 @@ let delay =
 
 (** {3 Name} *)
 
-let _name = ref "no-name"
-
 let name_k =
   let doc =
     Arg.info ~docs:Cmdliner.Manpage.s_common_options
-      ~doc:"Name of the unikernel" [ "name" ]
+      ~doc:
+        "Runtime name of the unikernel. Accessible with `Mirage_runtime.name` \
+         (), used for example by syslog"
+      ~absent:
+        "defaults to the configuration-time name (first argument to \
+         `Mirage.register`)"
+      [ "name" ]
   in
-  Arg.(value & opt string !_name doc)
+  Arg.(value & opt (some' string) None doc)
 
-let set_name s = _name := s
+let _name : string option ref = ref None
+let set_name s = _name := Some s
 
 let name =
   let r = Functoria_runtime.register_arg name_k in
-  fun () -> match r () with "no-name" -> !_name | x -> x
+  fun () ->
+    match (r (), !_name) with
+    | Some x, _ -> x
+    | None, Some x -> x
+    | None, None -> "no-name"
 
 (* Hooks *)
 
--- a/lib_runtime/mirage_runtime.mli
+++ b/lib_runtime/mirage_runtime.mli
@@ -79,14 +79,13 @@ val delay : int Term.t
 
 (** {2 Name} *)
 
-val set_name : string -> unit
-(** Set the default name. *)
-
-val name_k : string Term.t
+val name_k : string option Term.t
 (** The name key. *)
 
 val name : unit -> string
-(** The name of the unikernel. *)
+(** The current name of the unikernel. This is expected to be the same during
+    the lifetime of an unikernel (but there's no guarantee since it can be
+    modified). *)
 
 (** {2 Registering scheduler hooks} *)
 
@@ -150,3 +149,6 @@ val register : 'a Cmdliner.Term.t -> (unit -> 'a)
    superfluous. *)
 [@@ocamlformat "disable"]
 [@@ocaml.deprecated "Use Mirage_runtime.register_arg instead."]
+
+val set_name : string -> unit
+(** Set the name of the unikernel, called at load time for the default name. *)


lib_runtime/mirage_runtime.ml
```diff
@@ -119,6 +119,17 @@ let delay =
...
+(** {3 Name} *)
+
+let name_k =
+  let doc =
+    Arg.info ~docs:Cmdliner.Manpage.s_common_options
+      ~doc:"Name of the unikernel" [ "name" ]
+  in
+  Arg.(value & opt string "no-name" doc)
+
+let name = Functoria_runtime.register_arg name_k
```
reynir, line 131: What stands out to me is that `name` is the only runtime argument we register here in `Mirage_runtime`.

lib/devices/runtime_arg.ml
```diff
@@ -32,6 +32,8 @@ let runtime_network_key ~pos fmt =
 
 let delay = runtime_arg ~pos:__POS__ "delay"
 
+let name = runtime_arg ~pos:__POS__ "name_k"
```
reynir, line 35: I think if you use this you potentially register `name_k` twice and I don't know what happens if you do that.

reynir (commented):
I'm in favor of the idea, but the implementation raises some questions for me.

hannesm:
So, names:
- with this PR, a unikernel can use `Mirage_runtime.name ()` to retrieve its name \o/
- also, a potential syslog thingy can look as shown below
- now, the "register" (Mirage.register, in your config.ml) already carries a string -- but this is compile-time known the name for your project/unikernel (i.e. I have tens of unikernels all with the name "unipi" and "secondary-dns" -- i.e. identical image, but executed multiple times for different things) -- this is mainly used for the output filename, opam file, etc. I'm also fine to remove it entirely.
- albatross indeed has a name "albatross-client create <name> <options> <image>" -- and this <name> is often (for syslog & monitoring) passed another time to the unikernel (so is is "albatross-client create foo foo.hvt --arg='--name=foo'") -- it is nice to have a single name both on the outside and on the inside of a unikernel for metrics. what I plan to do is in albatross to pass the <name> by default as `--arg='--name=<name>'` (unless a `--name` is already provided), and if the unikernel fails with 64 (argument_error), restart without the `--name` (for backwards compatibility -- you may have a unikernel which doesn't have a `--name` argument yet).

syslog:
```OCaml
let syslog =
  let syslog = Runtime_arg.(v (syslog None)) in
  let connect _ modname = function
    | [ stack ; name ; syslog ] ->
      code ~pos:__POS__
        "Lwt.return (match %s with\
         | None -> Logs.warn (fun m -> m \"no syslog specified, dumping on stdout\")\
         | Some ip -> Logs.set_reporter (%s.create %s ip ~hostname:%s ()))"
        syslog modname stack name
    | _ -> assert false
  in
  impl
    ~packages:[ package ~sublibs:["mirage"] ~min:"0.5.0" "logs-syslog" ]
    ~runtime_args:[ Runtime_arg.(v name) ; syslog ]
    ~connect "Logs_syslog_mirage.Udp"
    (stackv4v6 @-> job)
```

lib_runtime/mirage_runtime.ml
hannesm: exactly -- also the delay_key, ocaml_runtime etc. are used by the generated code, while the name is not. there's no need to put more stuff into the generated main.ml.

hannesm:
> Is there an argument for defaulting this new name to no-name instead of the known static one?

That is a great idea. But how could this be done?

hannesm:
To explain why I don't quite understand how to do it:
- we could have the Runtime_arg.name_k being `string -> string runtime_arg`, and then in mirage.ml in the module Project(.create) fill that with `Info.name i` and have it as a runtime_args of the Project.create.
- but then, any other user of that argument in a config.ml, how would they refer to that argument? Or how would they receive the compile-time name? (see what the `syslog` code does above)

lib_runtime/mirage_runtime.ml
```diff
@@ -119,6 +119,23 @@ let delay =
...
+    Arg.info ~docs:Cmdliner.Manpage.s_common_options
+      ~doc:"Name of the unikernel" [ "name" ]
+  in
+  Arg.(value & opt string !_name doc)
+
+let set_name s = _name := s
+
+let name =
+  let r = Functoria_runtime.register_arg name_k in
+  fun () -> match r () with "no-name" -> !_name | x -> x
```
Firobe, line 137: Currently the man page would show `--name=VAL (default=no-name)`, as `_name` hasn't been set before key registration. I suggest something like that to both improve doc and make the behaviour a bit more explicit in the code (and maybe error instead of returning "no-name"?)

```ocaml
let _name = ref None

let name_k =
  let doc =
    Arg.info ~docs:Cmdliner.Manpage.s_common_options
      ~doc:"Runtime name of the unikernel. Used by syslog and accessible \
            with Mirage_runtime.name ()"
      ~absent:"defaults to the configuration-time name"
      [ "name" ]
  in
  Arg.(value & opt (some' string) None doc)

let set_name s = _name := Some s

let name =
  let r = Functoria_runtime.register_arg name_k in
  fun () -> match (r (), !_name) with
    | Some x, _ -> x
    | None, Some x -> x
    | None, None -> "no-name"
```

This would also need a change of `name_k`'s type to `string option Term.t` in the mli

lib_runtime/mirage_runtime.mli
```diff
@@ -77,6 +77,17 @@ val delay : int Term.t
     Defaults to 0. Useful for tenders and environments that take some time to
     bring devices up. *)
 
+(** {2 Name} *)
+
+val set_name : string -> unit
+(** Set the default name. *)
```
reynir, line 83: Should this be behind `(**/**)`? I don't mind either way, but if not we could change the doc for `val name` to say "The *current* name of the unikernel". It would be nice to know if one can safely assume the name will not change during the lifetime of the unikernel.

hannesm:
Thanks @Firobe and @reynir, I pushed [8182915](https://github.com/mirage/mirage/pull/1611/commits/8182915f0e8071efdeede116615335ba90858dcd) addressing your review comments.

PR #1612, opened 2025-09-07
Remove some stack exports

hannesm:
The main objective is to reduce our code. And do not expose things that have turned up to be not used at all.

This removes some exports of some "stacks", which I haven't found any use of. If there's somewhere a use of these things, please let me know and we can re-add them. The direct_stackv4v6 is used in unikernels that use uTCP.

Also, the types ipv4_config and ipv6_config have been removed. I couldn't find any use of them.

### DIFF
--- a/lib/devices/ip.ml
+++ b/lib/devices/ip.ml
@@ -18,12 +18,6 @@ let ipv4 : ipv4 typ = ip
 let ipv6 : ipv6 typ = ip
 let ipv4v6 : ipv4v6 typ = ip
 
-type ipv4_config = {
-  network : Ipaddr.V4.Prefix.t;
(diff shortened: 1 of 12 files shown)


ansiwen:
We use `ipv4_config` in [our code](https://github.com/Nitrokey/nethsm/blob/main/src/s_keyfender/config.ml#L8) (actually written by yourself over 4 years ago, [see here](https://github.com/Nitrokey/nethsm/blame/10971d026047b27326709c36791b1f28e194733b/src/s_keyfender/config.ml#L8C39-L8C39)).

But I'm happy to refactor the code, if there is a better way nowadays to achieve the same.

PR #1613, opened 2025-09-25
Refactor: qualify open

reynir:
This PR removes a bunch of `open _` top level phrases in `lib/devices/`. The motivation is that these `open _` top level phrases makes it more challenging to find the definition of symbols - especially if you are unfamiliar with the code base. The refactoring was guided by Merlin in Vim via `:MerlinRefactorOpenQualify`.

I mostly kept the `open _` top level phrases that refer to functoria modules. The exception being in one case where `open Functoria` shadowed another `Key` module. I also dropped `Astring` in one module as we did not use any functions that aren't in the stdlib's `String` module.

I may look into defining `Syntax` sub-modules in functoria so we reduce what we bring into scope all over the place.

### DIFF
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -16,7 +16,6 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-open Functoria.DSL
 open Devices
 module Type = Functoria.Type
 module Impl = Functoria.Impl
--- a/lib/mirage.ml
+++ b/lib/mirage.ml
@@ -16,13 +16,13 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-open Functoria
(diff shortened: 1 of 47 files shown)


lib/mirage.ml
```diff
@@ -16,13 +16,13 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-open Functoria
+open Functoria.DSL
```
hannesm, line 19:
```suggestion
```

according to OCaml-CI, this is unused

reynir: Oh right, we do `include Functoria.DSL` a bit further down.

hannesm:
looks nice, thanks for this epic removal of `open`. To make the CI happy, you need to run `dune bu @fmt --auto` and `dune runtest --auto`.

hannesm:
looks good, CI is in favour. merging.

PR #1604, opened 2025-04-03
Handle cmdliner deprecations (see #1602)

hannesm:
Please keep in mind that I've no idea what functoria's cmdliner magic is about -- but this code compiles.

PR #1614, opened 2025-10-03
upgrade to cmdliner 2.0

hannesm:
in contrast to #1604, here the tests are running. Turns out, some testcases used no longer available cmdliner functionality, and the "expect tests" used shortforms of the arguments (i.e. `package` instead of `packages`). Furthermore, functoria embedded a `--context` argument to argv, while it should be `--context-file`.

I understand that capturing cmdliner output is something @dbuenzli is against, and I agree with him -- but I've no other idea for these "e2e" tests. So I'll let them be there.

### DIFF
--- a/lib/devices/key.ml
+++ b/lib/devices/key.ml
@@ -49,10 +49,10 @@ let (target_conv : mode Cmdliner.Arg.conv), target_doc_alts =
       ("unikraft-qemu", `QEMU);
     ]
   in
-  let parser, printer = Cmdliner.Arg.enum enum in
-  ((parser, printer), Cmdliner.Arg.doc_alts_enum enum)
+  let conv = Cmdliner.Arg.enum enum in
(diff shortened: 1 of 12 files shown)


hannesm:
Please keep in mind that this is the least invasive change I could come up with. I'm sure we can improve the code going full in with cmdliner 2.0.0. But I'm not familiar enough with functoria to understand which parts to obsolete / move.

hannesm:
> I have no clue why the expect tests in CI have quoted arguments, and not quoted ones on my machine.

good news, at least on macOS the CI behaves like my local machine. Any hints for the quoting / not quoting would be very welcome! Reminds me of the fact that in the opam-repository we restrict to OCaml < 5 for with-test due to difference in the expect output. As a workaround, I can offer to put a `| tr -d '` at the offending lines.

PR #1627, opened 2025-11-24
Cram tests less reliant on exact compiler errors

reynir:


hannesm:
I cherry-picked this into #1629 to see whether we get a green CI there. Thanks a lot!

hannesm:
But that fails as well, so I reverted the cherry-pick.

hannesm:
I've no clue why:
```
--- a/_build/default/test/functoria/e2e/errors.t
+++ b/_build/default/test/functoria/e2e/errors.t.corrected
@@ -21,7 +21,7 @@ Then, not enough:
$ ./test.exe configure -f errors/in_functor_not_enough.ml
test.exe: [WARNING] Skipping version check, since our_version ("1.0~test") fails to parse: only digits and . allowed in version
$ dune build 2>&1 | grep '^Error:'
-  Error: The module Unikernel_make__4 is a functor, it cannot have any components
+  [1]
$ ./test.exe clean -f errors/in_functor_not_enough.ml


Also check that we have proper errors when the config file is missing:
```

reynir:
yea, sorry I should have closed this PR. It never worked /o\

PR #1628, opened 2025-12-03
Dhcp ipv4 and leases

reynir:
This depends on https://github.com/mirage/charrua/pull/142.

This PR makes available the DHCP lease from `Mirage.ipv4_of_dhcp` to configure other devices. As an example `Mirage.generic_dns_client` is extended so it can get DNS servers from DHCP. Another candidate could be syslog that may use DHCP log server as its log destination.

The code currently compiles, but I am almost certain it's not 100% correct. I am opening a draft PR as I think it is of interest to get early feedback on the overall design. I think the interface is how I want it; what needs work is the code generation that I'm almost certain isn't correct.

The ideas have been tested out in this PR: https://github.com/robur-coop/traceroute/pull/7

The new `Mirage.dhcp_requests` type contains mutable state. This is somewhat straying away from the current style. I think however it has merit. The challenge is that between a device `x` and a `dhcp_ipv4` there is a sort of mutual dependency: `x` wants a lease from `dhcp_ipv4` while `dhcp_ipv4` is required to know what DHCP option `x` and any other dependents are interested in. I think with mutable state we have a somewhat nice way of propagating back from `x` to `dhcp_ipv4` what dhcp options we are interested in.

### DIFF
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -31,17 +31,13 @@ let generic_dns_client ?dhcp ?group ?timeout ?nameservers ?cache_size () =
             ] ->
               (* The nameservers argument has to be turned into a string. *)
               code ~pos:__POS__
-                "let nameservers =@[@ \
-                   match %s, %s with@ | None, None -> None@ \
-                   | Some ns, _ -> Some ns@ \
-                   | None, Some lease ->@[@ \
-                     let nameservers = Dhcp_wire.collect_name_servers lease in@.\
-                     Some (List.concat_map (fun ip -> \
-                            [Fmt.str \"udp:%%a\" Ipaddr.V4.pp ip; \
-                             Fmt.str \"tcp:%%a\" Ipaddr.V4.pp ip]) nameservers)\
-                 @]@]@.\
-                 in@.\
-                 %s.connect @[?nameservers ?timeout:%s ?cache_size:%s@ (%s, %s)@]"
+                "let nameservers =@[@ match %s, %s with@ | None, None -> None@ \
+                 | Some ns, _ -> Some ns@ | None, Some lease ->@[@ let \
+                 nameservers = Dhcp_wire.collect_name_servers lease in@.Some \
+                 (List.concat_map (fun ip -> [Fmt.str \"udp:%%a\" Ipaddr.V4.pp \
+                 ip; Fmt.str \"tcp:%%a\" Ipaddr.V4.pp ip]) \
+                 nameservers)@]@]@.in@.%s.connect @[?nameservers ?timeout:%s \
+                 ?cache_size:%s@ (%s, %s)@]"
                 nameservers lease modname timeout cache_size stackv4v6
                 happy_eyeballs
           | _ -> Misc.connect_err "generic_dns_client" 6
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -42,7 +42,7 @@ let generic_dns_client ?dhcp ?group ?timeout ?nameservers ?cache_size () =
                 happy_eyeballs
           | _ -> Misc.connect_err "generic_dns_client" 6
         in
-        ( package ~min:"2.0.0" ~max:"3.0.0" "charrua" :: packages,
+        ( package ~min:"2.0.0" ~max:"4.0.0" "charrua" :: packages,
           Some [ dep lease ],
           connect )
   in
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -29,11 +29,17 @@ let generic_dns_client ?dhcp ?group ?timeout ?nameservers ?cache_size () =
           | [
               stackv4v6; happy_eyeballs; lease; nameservers; timeout; cache_size;
             ] ->
+              (* The nameservers argument has to be turned into a string. *)
               code ~pos:__POS__
                 "let nameservers =@[@ \
-                 match %s, %s with@ | None, None -> None@ \
-                 | Some ns, _ -> Some ns@ \
-                 | None, Some lease ->@ Some (Dhcp_wire.collect_name_servers lease)@ @]\
+                   match %s, %s with@ | None, None -> None@ \
+                   | Some ns, _ -> Some ns@ \
+                   | None, Some lease ->@[@ \
+                     let nameservers = Dhcp_wire.collect_name_servers lease in@.\
+                     Some (List.concat_map (fun ip -> \
+                            [Fmt.str \"udp:%%a\" Ipaddr.V4.pp ip; \
+                             Fmt.str \"tcp:%%a\" Ipaddr.V4.pp ip]) nameservers)\
+                 @]@]@.\
                  in@.\
                  %s.connect @[?nameservers ?timeout:%s ?cache_size:%s@ (%s, %s)@]"
                 nameservers lease modname timeout cache_size stackv4v6
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -4,18 +4,40 @@ type dns_client = Dns_client
 
 let dns_client = typ Dns_client
 
-let generic_dns_client ?group ?timeout ?nameservers ?cache_size () =
+let generic_dns_client ?dhcp ?group ?timeout ?nameservers ?cache_size () =
+  Option.iter
+    (fun (requests, _) -> Ip.Dhcp_requests.add requests 6 (* DNS_SERVERS *))
+    dhcp;
   let packages = [ package "dns-client-mirage" ~min:"10.0.0" ~max:"11.0.0" ] in
   let nameservers = Runtime_arg.dns_servers ?group nameservers
   and timeout = Runtime_arg.dns_timeout ?group timeout
   and cache_size = Runtime_arg.dns_cache_size ?group cache_size in
   let runtime_args = Runtime_arg.[ v nameservers; v timeout; v cache_size ] in
-  let connect _info modname = function
-    | [ stackv4v6; happy_eyeballs; nameservers; timeout; cache_size ] ->
-        code ~pos:__POS__
-          {ocaml|%s.connect @[?nameservers:%s ?timeout:%s ?cache_size:%s@ (%s, %s)@]|ocaml}
-          modname nameservers timeout cache_size stackv4v6 happy_eyeballs
-    | _ -> Misc.connect_err "generic_dns_client" 5
+  let packages, extra_deps, connect =
+    match dhcp with
+    | None ->
+        let connect _info modname = function
+          | [ stackv4v6; happy_eyeballs; nameservers; timeout; cache_size ] ->
+              code ~pos:__POS__
+                {ocaml|%s.connect @[?nameservers:%s ?timeout:%s ?cache_size:%s@ (%s, %s)@]|ocaml}
+                modname nameservers timeout cache_size stackv4v6 happy_eyeballs
+          | _ -> Misc.connect_err "generic_dns_client" 5
+        in
+        (packages, None, connect)
+    | Some (_, lease) ->
+        let connect _info modname = function
+          | [
+              stackv4v6; happy_eyeballs; lease; nameservers; timeout; cache_size;
+            ] ->
+              code ~pos:__POS__
+                {ocaml|let nameservers = match %s, %s with | None, None -> None | Some ns, _ -> Some ns | None, Some lease -> Some (Dhcp_wire.collect_name_servers lease) in %s.connect @[?nameservers ?timeout:%s ?cache_size:%s@ (%s, %s)@]|ocaml}
+                nameservers lease modname timeout cache_size stackv4v6
+                happy_eyeballs
+          | _ -> Misc.connect_err "generic_dns_client" 6
+        in
+        ( package ~min:"2.0.0" ~max:"3.0.0" "charrua" :: packages,
+          Some [ dep lease ],
+          connect )
   in
-  impl ~runtime_args ~packages ~connect "Dns_client_mirage.Make"
+  impl ~runtime_args ~packages ?extra_deps ~connect "Dns_client_mirage.Make"
     (Stack.stackv4v6 @-> Happy_eyeballs.happy_eyeballs @-> dns_client)
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -33,7 +33,7 @@ let generic_dns_client ?dhcp ?group ?timeout ?nameservers ?cache_size () =
               code ~pos:__POS__
                 "let nameservers =@[@ match %s, %s with@ | None, None -> None@ \
                  | Some ns, _ -> Some ns@ | None, Some lease ->@[@ let \
-                 nameservers = Dhcp_wire.collect_name_servers lease in@.Some \
+                 nameservers = Dhcp_wire.collect_dns_servers lease in@.Some \
                  (List.concat_map (fun ip -> [Fmt.str \"udp:%%a\" Ipaddr.V4.pp \
                  ip; Fmt.str \"tcp:%%a\" Ipaddr.V4.pp ip]) \
                  nameservers)@]@]@.in@.%s.connect @[?nameservers ?timeout:%s \
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -30,7 +30,12 @@ let generic_dns_client ?dhcp ?group ?timeout ?nameservers ?cache_size () =
               stackv4v6; happy_eyeballs; lease; nameservers; timeout; cache_size;
             ] ->
               code ~pos:__POS__
-                {ocaml|let nameservers = match %s, %s with | None, None -> None | Some ns, _ -> Some ns | None, Some lease -> Some (Dhcp_wire.collect_name_servers lease) in %s.connect @[?nameservers ?timeout:%s ?cache_size:%s@ (%s, %s)@]|ocaml}
+                "let nameservers =@[@ \
+                 match %s, %s with@ | None, None -> None@ \
+                 | Some ns, _ -> Some ns@ \
+                 | None, Some lease ->@ Some (Dhcp_wire.collect_name_servers lease)@ @]\
+                 in@.\
+                 %s.connect @[?nameservers ?timeout:%s ?cache_size:%s@ (%s, %s)@]"
                 nameservers lease modname timeout cache_size stackv4v6
                 happy_eyeballs
           | _ -> Misc.connect_err "generic_dns_client" 6
--- a/lib/devices/dns.ml
+++ b/lib/devices/dns.ml
@@ -33,9 +33,10 @@ let generic_dns_client ?dhcp ?group ?timeout ?nameservers ?cache_size () =
               code ~pos:__POS__
                 "let nameservers =@[@ match %s, %s with@ | None, None -> None@ \
                  | Some ns, _ -> Some ns@ | None, Some lease ->@[@ let \
-                 nameservers = Dhcp_wire.collect_dns_servers lease in@.Some \
-                 (List.concat_map (fun ip -> [Fmt.str \"udp:%%a\" Ipaddr.V4.pp \
-                 ip; Fmt.str \"tcp:%%a\" Ipaddr.V4.pp ip]) \
+                 nameservers = Dhcp_wire.collect_dns_servers lease in@.if \
+                 nameservers = [] then None@.else Some (List.concat_map (fun \
+                 ip -> [Fmt.str \"udp:%%a\" Ipaddr.V4.pp ip; Fmt.str \
+                 \"tcp:%%a\" Ipaddr.V4.pp ip]) \
                  nameservers)@]@]@.in@.%s.connect @[?nameservers ?timeout:%s \
                  ?cache_size:%s@ (%s, %s)@]"
                 nameservers lease modname timeout cache_size stackv4v6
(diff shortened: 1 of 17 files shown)


reynir:
I did some fixes and now I can compile the following unikernel which gets its ip address **and** its dns servers from DHCP and does a HTTP request. What I don't like so much so far is I had to build the TCP/IP stack manually. It would be great to use `generic_stackv4v6` and I may look into modifying that.

<details>
<summary>Unikernel code and run output</summary>

Config.ml:
```OCaml
open Mirage

let client =
  main "Unikernel.Client" (alpn_client @-> job)

let () =
  let e = ethif default_network in
  let a = arp e in
  let dhcp_requests, dhcp_ipv4 = ipv4_of_dhcp default_network e a in
  let stack =
    direct_stackv4v6 (dhcp_proj_net $ dhcp_ipv4) e a (dhcp_proj_ipv4 $ dhcp_ipv4)
      (create_ipv6 (dhcp_proj_net $ dhcp_ipv4) e)
  in
  let he = generic_happy_eyeballs stack in
  let dns = generic_dns_client ~dhcp:(dhcp_requests, dhcp_proj_lease $ dhcp_ipv4) stack he in
  let alpn_client =
    let mimic = mimic_happy_eyeballs stack he dns in
    paf_client (tcpv4v6_of_stackv4v6 stack) mimic
  in
  register "http-fetch" [ client $ alpn_client ]
```
Unikernel.ml:
```OCaml
open Lwt.Syntax

module K = struct
  open Cmdliner
  let url =
    let doc = Arg.info ~doc:"URL to fetch" [ "url" ] in
    Mirage_runtime.register_arg @@
    Arg.(value & opt string "https://mirageos.org/" doc)
end

module Client (_ : Http_mirage_client.S) = struct
  let start http =
    let url = K.url () in
    Fmt.pr "Fetching %s with http-mirage-client\n\n%!" url;
    let+ r =
      Http_mirage_client.request http url
        (fun _response len data -> Lwt.return (len + String.length data))
        0
    in
    match r with
    | Error e -> Fmt.pr "Error fetching: %a\n%!" Mimic.pp_error e
    | Ok (_response, len) ->
      Fmt.pr "Received body length: %u\n" len;
      Fmt.pr "http-mirage-client fetch done\n------------\n%!"
end
```
Sample run with no arguments on a network with DHCP that says to use DNS server 172.23.0.53:
```
console 2025-12-05T11:42:06-00:00:             |      ___|
console 2025-12-05T11:42:06-00:00:   __|  _ \  |  _ \ __ \
console 2025-12-05T11:42:06-00:00: \__ \ (   | | (   |  ) |
console 2025-12-05T11:42:06-00:00: ____/\___/ _|\___/____/
console 2025-12-05T11:42:06-00:00: Solo5: Bindings version v0.10.0
console 2025-12-05T11:42:06-00:00: Solo5: Memory map: 32 MB addressable:
console 2025-12-05T11:42:06-00:00: Solo5:   reserved @ (0x0 - 0xfffff)
console 2025-12-05T11:42:06-00:00: Solo5:       text @ (0x100000 - 0x47bfff)
console 2025-12-05T11:42:06-00:00: Solo5:     rodata @ (0x47c000 - 0x53efff)
console 2025-12-05T11:42:06-00:00: Solo5:       data @ (0x53f000 - 0x8b1fff)
console 2025-12-05T11:42:06-00:00: Solo5:       heap >= 0x8b2000 < stack < 0x2000000
console 2025-12-05T11:42:06-00:00: 2025-12-05T11:42:06-00:00: [INFO] [netif] Plugging into service with mac 00:80:41:28:4e:fe mtu 1500
console 2025-12-05T11:42:06-00:00: 2025-12-05T11:42:06-00:00: [INFO] [ethernet] Connected Ethernet interface 00:80:41:28:4e:fe
console 2025-12-05T11:42:06-00:00: 2025-12-05T11:42:06-00:00: [INFO] [dhcp_client_lwt] Lease obtained! IP: 172.23.1.129, routers: 172.23.0.1
console 2025-12-05T11:42:06-00:00: 2025-12-05T11:42:06-00:00: [INFO] [ARP] Sending gratuitous ARP for 172.23.1.129 (00:80:41:28:4e:fe)
console 2025-12-05T11:42:06-00:00: 2025-12-05T11:42:06-00:00: [INFO] [ipv6] IP6: Starting
console 2025-12-05T11:42:06-00:00: 2025-12-05T11:42:06-00:00: [INFO] [ndpc6] ND6: Unsupported ND option in RA: ty=14 len=1
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [ndpc6] IP6: Processing unknown option, MSB 5
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [ndpc6] ICMP6: Unknown packet type: ty=143
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [ipv6] IP6: Started with fe80::280:41ff:fe28:4efe
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [tcp.pcb] TCP layer connected on 172.23.1.129/16, fe80::280:41ff:fe28:4efe/64
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [udp] UDP layer connected on 172.23.1.129/16, fe80::280:41ff:fe28:4efe/64
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [tcpip-stack-direct] Dual TCP/IP stack assembled: mac=00:80:41:28:4e:fe,ip=172.23.1.129/16, fe80::280:41ff:fe28:4efe/64
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [WARNING] [dns_client_mirage] ignoring UDP nameservers 172.23.0.53:53, using TCP nameservers 172.23.0.53:53
console 2025-12-05T11:42:07-00:00: Fetching https://mirageos.org/ with http-mirage-client
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [ndpc6] IP6: Processing unknown option, MSB 5
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [ndpc6] ICMP6: Unknown packet type: ty=143
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [ndpc6] IP6: Processing unknown option, MSB 5
console 2025-12-05T11:42:07-00:00: 2025-12-05T11:42:07-00:00: [INFO] [ndpc6] ICMP6: Unknown packet type: ty=143
console 2025-12-05T11:42:08-00:00: Received body length: 19711
console 2025-12-05T11:42:08-00:00: http-mirage-client fetch done
console 2025-12-05T11:42:08-00:00: ------------
console 2025-12-05T11:42:08-00:00: Solo5: solo5_exit(0) called
```

</details>

Using the mirage-skeleton/device-usage/http-fetch/ code I need to pass `--dhcp=true` to `mirage configure` before dhcp works. It seems to me we may want to change the default so we explicitly have to disable dhcp. With current `--dhcp=true` the usage of dhcp is decided at runtime depending on the `--ipv4=` etc arguments.

hannesm:
> Using the mirage-skeleton/device-usage/http-fetch/ code I need to pass --dhcp=true to mirage configure before dhcp works. It seems to me we may want to change the default so we explicitly have to disable dhcp.

Yes, I'm completely in favour to remove this dhcp flag for configure time (or make it default to true); and have by default the decision at run-time whether to use DHCP or not (by providing the `--ipv4=..` argument (to avoid DHCP), or not providing it (so the unikernel will ask for DHCP)).

reynir:
I changed the default of --dhcp and I added runtime options to the dhcp thing (such as --ipv4, --ipv4-gateway, and the no_init --ipv6-only).

I attempted to change the generic_stackv4v6 so we can interact with the dhcp, but I struggle as  generic_stackv4 is either a socket thingie, or it's a direct thingie which may or may not do dhcp. So maybe the `dhcp_requests` value should be an argument to `ipv4_of_dhcp`?? so we can create one in any case and just ignore it if it isn't in the case of direct **and** dhcp.

reynir:
The first commit I just pushed makes the following changes to the config.ml from before necessary 
```diff
--- config.ml'	2025-12-05 14:44:30.298736766 +0100
+++ config.ml	2025-12-05 14:43:16.848612714 +0100
@@ -4,9 +4,10 @@
   main "Unikernel.Client" (alpn_client @-> job)
 
 let () =
+  let dhcp_requests = make_dhcp_requests () in
   let e = ethif default_network in
   let a = arp e in
-  let dhcp_requests, dhcp_ipv4 = ipv4_of_dhcp default_network e a in
+  let dhcp_ipv4 = ipv4_of_dhcp ~dhcp_requests default_network e a in
   let stack =
     direct_stackv4v6 (dhcp_proj_net $ dhcp_ipv4) e a (dhcp_proj_ipv4 $ dhcp_ipv4)
       (create_ipv6 (dhcp_proj_net $ dhcp_ipv4) e)
```

The second requires following changes:
```diff
--- config.ml'	2025-12-05 14:44:30.298736766 +0100
+++ config.ml	2025-12-05 15:06:09.777211090 +0100
@@ -4,15 +4,10 @@
   main "Unikernel.Client" (alpn_client @-> job)
 
 let () =
-  let e = ethif default_network in
-  let a = arp e in
-  let dhcp_requests, dhcp_ipv4 = ipv4_of_dhcp default_network e a in
-  let stack =
-    direct_stackv4v6 (dhcp_proj_net $ dhcp_ipv4) e a (dhcp_proj_ipv4 $ dhcp_ipv4)
-      (create_ipv6 (dhcp_proj_net $ dhcp_ipv4) e)
-  in
+  let dhcp_requests = make_dhcp_requests () in
+  let stack, lease = generic_stackv4v6_with_lease ~dhcp_requests default_network in
   let he = generic_happy_eyeballs stack in
-  let dns = generic_dns_client ~dhcp:(dhcp_requests, dhcp_proj_lease $ dhcp_ipv4) stack he in
+  let dns = generic_dns_client ~dhcp:(dhcp_requests, lease) stack he in
   let alpn_client =
     let mimic = mimic_happy_eyeballs stack he dns in
     paf_client (tcpv4v6_of_stackv4v6 stack) mimic
```
As you can see there is now a `Mirage.generic_stackv4v6_with_lease` which returns a pair of devices: a stack and a lease.

palainp:
Thank you @reynir for your work on that.

I'll try to give a try next week with an update to the example from mirage-nat (https://github.com/mirage/mirage-nat/blob/0f62253d76e957678a30bb39138cf7c4a142a486/example/config.ml#L28-L31) :)

lib/devices/stack.ml
```diff
@@ -105,8 +115,17 @@ let generic_stackv4v6 ?group ?(dhcp_key = Key.value @@ Key.dhcp ?group ())
...
+  ( match_impl p
+      [ (`Socket, socket_stackv4v6 ?group ()) ]
+      ~default:(fst generic_ipv4v6_stack),
+    match_impl p [ (`Socket, no_lease) ] ~default:(snd generic_ipv4v6_stack) )
+
+let generic_stackv4v6 ?group ?dhcp_key ?net_key ?ipv4_network ?ipv4_gateway
+    ?ipv6_network ?ipv6_gateway ?tcp tap =
+  generic_stackv4v6_with_lease ?group ?dhcp_requests:None ?dhcp_key ?net_key
+    ?ipv4_network ?ipv4_gateway ?ipv6_network ?ipv6_gateway ?tcp tap
+  |> fst
```
reynir, line 130: Here we don't take the optional `?dhcp_requests` and we don't return the lease. This way the interface is the same as before.

lib/mirage.ml
```diff
@@ -117,15 +117,25 @@ type 'a ip = 'a Ip.ip
...
 let ipv4 = Ip.ipv4
 let ipv6 = Ip.ipv6
 let ipv4_qubes = Ip.ipv4_qubes
 let ipv4v6 = Ip.ipv4v6
+let dhcp_ipv4 = Ip.dhcp_ipv4
+let lease = Ip.lease
 let ipv4_of_dhcp = Ip.ipv4_of_dhcp
+let dhcp_proj_net = Ip.dhcp_proj_net
+let dhcp_proj_ipv4 = Ip.dhcp_proj_ipv4
+let dhcp_proj_lease = Ip.dhcp_proj_lease
```
reynir, line 133: So a `dhcp_ipv4` is in a way three things at once: a view of a network device where DHCP packets are intercepted, an ipv4, and a lease. These three functions are projections from `dhcp_ipv4` to the three different devices. This is perhaps a bit odd, but it was the best I could think of that would work with Mirage's functor application language Functoria.

lib/mirage.ml
```diff
@@ -117,15 +117,25 @@ type 'a ip = 'a Ip.ip
...
 let create_ipv4 = Ip.create_ipv4
 let create_ipv6 = Ip.create_ipv6
 let create_ipv4v6 = Ip.create_ipv4v6
+let make_dhcp_requests = Ip.Dhcp_requests.make
+let add_dhcp_request = Ip.Dhcp_requests.add
```
reynir, line 139: These also stand out a bit from the Mirage interface. This is how we can propagate back to the dhcp device what DHCP options dependent devices need. It's using mutable state and things have to be called in the right order, but I added checks so it will complain loudly if this is not the case. I think it's not easy to write a config.ml that would violate the required order.

lib/devices/dns.ml
```diff
@@ -4,18 +4,47 @@ type dns_client = Dns_client
...
+                 ip; Fmt.str \"tcp:%%a\" Ipaddr.V4.pp ip]) \
+                 nameservers)@]@]@.in@.%s.connect @[?nameservers ?timeout:%s \
+                 ?cache_size:%s@ (%s, %s)@]"
+                nameservers lease modname timeout cache_size stackv4v6
+                happy_eyeballs
+          | _ -> Misc.connect_err "generic_dns_client" 6
+        in
+        ( package ~min:"2.0.0" ~max:"3.0.0" "charrua" :: packages,
+          Some [ dep lease ],
+          connect )
```
reynir, line 48: There is quite some code duplication here.

The code generation is also more complex because the list of ip addresses has to be turned into a list of strings "udp:0.0.0.0", "tcp:0.0.0.0". The mirage dns client doesn't support UDP at the moment, so it's a bit unnecessary to emit the "udp:0.0.0.0" strings, but I don't want to hardcode for that limitation.

lib/devices/stack.ml
```diff
@@ -54,15 +52,23 @@ let keyed_direct_stackv4v6 ?tcp ~ipv4_only ~ipv6_only network eth arp ipv4 ipv6
   $ Udp.direct_udp ip
   $ match tcp with None -> Tcp.direct_tcp ip | Some tcp -> tcp
 
-let generic_ipv4v6_stack p ?group ?ipv4_network ?ipv4_gateway ?ipv6_network
-    ?ipv6_gateway ?(arp = Arp.arp) ?tcp tap =
+let no_lease =
+  let connect _ _ _ = code ~pos:__POS__ "Lwt.return None" in
+  impl ~connect "Lwt" Ip.lease
```
reynir: This may be useful to expose in mirage.mli.

lib/devices/dns.ml
reynir: The two match cases could be unified if we use `Ip.no_lease` in the `None` case as the `extra_deps`. Then the `packages` and `connect` can be shared in both cases. The downside is then that the runtime would depend on charrua always (though not charrua-client).

hannesm: I appreciate to avoid "depend on charrua always", and if that means there's need for code duplication, so be it.

reynir: Thanks. I agree, I prefer not to pull in charrua unnecessarily.

lib/devices/dns.ml
```diff
@@ -4,18 +4,47 @@ type dns_client = Dns_client
...
+            ] ->
+              (* The nameservers argument has to be turned into a string. *)
+              code ~pos:__POS__
+                "let nameservers =@[@ match %s, %s with@ | None, None -> None@ \
+                 | Some ns, _ -> Some ns@ | None, Some lease ->@[@ let \
+                 nameservers = Dhcp_wire.collect_dns_servers lease in@.Some \
+                 (List.concat_map (fun ip -> [Fmt.str \"udp:%%a\" Ipaddr.V4.pp \
+                 ip; Fmt.str \"tcp:%%a\" Ipaddr.V4.pp ip]) \
+                 nameservers)@]@]@.in@.%s.connect @[?nameservers ?timeout:%s \
+                 ?cache_size:%s@ (%s, %s)@]"
```
hannesm, line 41: This looks fine. Note: the nameserver provided on command-line (boot parameter) takes precedence over the DHCP option. Also, the DHCP option only contains IPv4 addresses.

A question is what happens if the badly-configured DHCP server emits a Dns_servers option with the empty list?

lib/devices/ip.ml
```diff
@@ -7,11 +7,32 @@ type 'a ip = IP
...
+  type t = { mutable requests : IntSet.t; mutable consumed : bool }
+
+  let make () = { requests = IntSet.empty; consumed = false }
+
+  let add t req =
+    if t.consumed then invalid_arg "DHCP option code added too late";
+    t.requests <- IntSet.add req t.requests
+
+  let consume t =
+    assert (not t.consumed);
```
hannesm, line 34: can we log or do invalid_arg instead of assert here? there may be people disabling assertions, which will lead to strange behaviour silently.

lib/devices/dns.ml
```diff
@@ -4,18 +4,47 @@ type dns_client = Dns_client
...
+                nameservers lease modname timeout cache_size stackv4v6
+                happy_eyeballs
+          | _ -> Misc.connect_err "generic_dns_client" 6
+        in
+        ( package ~min:"2.0.0" ~max:"4.0.0" "charrua" :: packages,
```
hannesm, line 46: what are the bounds? below there's 3.0.0 as minimum. Though I don't care much, these bounds are merged anyways...

lib/devices/dns.ml
reynir: So `Dhcp_wire.collect_dns_servers` will return the empty list both if the server didn't send the dns servers option and if the dns servers option is an empy list (which is disallowed by the RFC, but I don't remember if the parser enforces this). It seems we don't check if the list is empty which we should and in that case return `None`. Nice catch.

hannesm:
These changes look fine. Can we get this to work nicely in CI? I guess #1629 is needed. But also, is https://github.com/mirage/mirage-skeleton/pull/418 needed (if yes, please rebase that, and add to the initial comment something link "requires https://github.com/mirage/mirage-skeleton/pull/418" -- then the "Mirage CI" will pick it up).

lib/devices/dns.ml
reynir: This code works fine with 2.0.0 although it would be odd to use it without the 3.0.0 bits that produce the dhcp options. So in my view 2.0.0 is correct even if a bit odd.

lib/devices/ip.ml
reynir: This function is not exported and is used once in this module. We can as well remove the assert.

lib/devices/dns.ml
reynir: Addressed in 602d87d32b6fbd7e2bc8070fc2459dcc2107b0f7 (the diff is unfortunately pretty unreadable due to ocamlformat...)

reynir:
> These changes look fine. Can we get this to work nicely in CI? I guess #1629 is needed. But also, is [mirage/mirage-skeleton#418](https://github.com/mirage/mirage-skeleton/pull/418) needed (if yes, please rebase that, and add to the initial comment something link "requires [mirage/mirage-skeleton#418](https://github.com/mirage/mirage-skeleton/pull/418)" -- then the "Mirage CI" will pick it up).

We need to not use lwt 6.0.0 with the currently released dune versions (not sure it's possible to express that). From memory, strictly speaking https://github.com/mirage/mirage-skeleton/pull/418 is not needed, but it exercises the new APIs. It's been rebased (the conflict was a purely syntactic change between unnecessary parens).

hannesm:
`We need to not use lwt 6.0.0 with the currently released dune versions` yes, that's what #1629 tries to achieve. I guess a dune release may fix this, so we'll have some trouble expressing that once it is released (such as 'avoid lwt 6 if dune < 3.Y').

lib/devices/ip.ml
```diff
@@ -7,11 +7,33 @@ type 'a ip = IP
...
+module Dhcp_requests = struct
+  module IntSet = Set.Make (Int)
+
+  type t = { mutable requests : IntSet.t; mutable consumed : bool }
+
+  let make () = { requests = IntSet.empty; consumed = false }
+
+  let add t req =
+    if t.consumed then invalid_arg "DHCP option code added too late";
+    if req <= 1 || req >= 255 then invalid_arg "Invalid DHCP option code";
```
reynir: I added a check here so we get an error if a bad option code was used. Option codes 0 and 255 are padding and end respectively and I don't think it makes sense to request those.

palainp: On the snippet req is compared to 1 and 255. So maybe you mean `req < 1` or `req <= 0` ?
EDIT: code 1 is subnet mask

reynir: Indeed! That was a mistake. I fixed it and put you as a co-author, and I also added a comment documenting why we do that check. Thanks!

hannesm (approved):
I reviewed this code again, and it looks fine. It finally allows us to use DHCP and DHCP options in MirageOS unikernels. This is a huge step forward.

hannesm:
My plan is to merge and release this early next week. Feedback and reviews are very welcome.

PR #1629, opened 2026-01-27
mirage-runtime: avoid lwt 6.0.0 (due to the dune bug)

hannesm:
This should allow the CI to pass.


reynir (approved):
Thanks!

hannesm:
OCaml-CI failures are expected on 5.4 due to different output of the compiler. There's a different attempt to fix it in #1627 -- but that doesn't work as far as I can tell.

reynir:
- There are two failures for openbsd in dune also observed in https://github.com/mirage/charrua/pull/144
- The Windows errors seem unrelated (`ssh opam@localhost`??)
- Finally, FreeBSD builds are stuck which I've observed today in other PRs as well.
This is fine to merge IMO.

PR #1630, opened 2026-02-17
[tcp] introduce tcp-max-open-connections runtime argument

Firobe:
Companion PR to https://github.com/mirage/mirage-tcpip/pull/535. This introduces the `--tcp-max-open-connections` to customize the maximum amount of listening TCP connections, which defaults to 1000.

**TODO** before considering this PR for merging, if and when https://github.com/mirage/mirage-tcpip/pull/535 is merged and released, raise upper bound for `mirage-tcpip`

See also https://github.com/mirage/mirage-tcpip/issues/534

### DIFF
--- a/lib/devices/runtime_arg.ml
+++ b/lib/devices/runtime_arg.ml
@@ -51,6 +51,10 @@ let interface ?group ?docs default =
   runtime_network_key ~pos:__POS__ "interface %a%a%S" pp_group group pp_docs
     docs default
 
+let tcp_max_open_connections ?group ?docs default =
+  runtime_network_key ~pos:__POS__ "tcp_max_open_connections %a%a%d" pp_group
+    group pp_docs docs default
+
 module V4 = struct
   open Ipaddr.V4
 
--- a/lib/devices/runtime_arg.mli
+++ b/lib/devices/runtime_arg.mli
@@ -33,6 +33,10 @@ val v : 'a arg -> Functoria.Runtime_arg.t
 val interface : ?group:string -> ?docs:string -> string -> string runtime_arg
 (** A network interface. *)
 
+val tcp_max_open_connections :
+  ?group:string -> ?docs:string -> int -> int runtime_arg
+(** A limit on the number of open connections for the TCP layer *)
+
 (** Ipv4 Arguments. *)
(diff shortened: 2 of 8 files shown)


hannesm:
What is the evidence for `1000`? Would you mind to share benchmarks with different CPUs and memory configurations that show some hints why `1000` is a good default?

Firobe:
> What is the evidence for `1000`? Would you mind to share benchmarks with different CPUs and memory configurations that show some hints why `1000` is a good default?

It was initially chosen by @edwintorok (IIUC) with the criterion that an HVT unikernel with 512 MB of RAM should survive a SYN flood. But it's also clear it is arbitrary and quite hard to tune. It makes some sense with 512 MB being the default memory alocated by solo5.

If we agree on the target, I'm happy to run some benchmarks (maybe using @edwintorok's attack?) to adjust that number based on some evidence.

edwintorok:
> > What is the evidence for `1000`? Would you mind to share benchmarks with different CPUs and memory configurations that show some hints why `1000` is a good default?
> 
> It was initially chosen by @edwintorok (IIUC) with the criterion that an HVT unikernel with 512 MB of RAM should survive a SYN flood. 

Linux by default supports 1024 open file descriptors, so that seemed like a good place to start (you can raise that limit with `ulimit`).
Also IIRC a single unikernel is limited to a single core (but perhaps Unikraft can change that).

So it seems that the way to scale unikernels up is not by giving a single unikernel the ability to handle a lot of connections, but rather by spawning many unikernels and load balancing connections across them.

If you want to make your load balancer a unikernel too, then you might need to give it the ability to handle more connections.

Having said that, there might be other desirable targets here (10 000, 100 000), and we should do some tests to see what is the highest value that still gives adequate protection for a typical unikernel (or one with the max amount 4GiB of RAM).
IIRC 1000 seems to have defended against my attack well enough at the time, and I haven't tried measuring higher values.

> an HVT unikernel with 512 MB of RAM

I was surprised to discover that there is an ~4GiB limit, even on x86-64, so even if your host would have more RAM, you can't give more to a single unikernel. So testing with both 512MiB and 4GiB might be useful.

PR #1631, opened 2026-03-02
Make opam-monorepo invocations a bit more verbose

reynir:
See https://github.com/tarides/opam-monorepo/issues/420 for the motivation.

In this implementation it is not configurable.


Firobe:
You'll just need to update the cram test results (see the CI)

Firobe:
@reynir I think we can close this?

reynir:
Opam-monorepo can take an environment variable `OPAM_MONOREPO_VERBOSITORY=info` to achieve this. See https://github.com/tarides/opam-monorepo/pull/422. Note that it's not yet released.

PR #1635, opened 2026-04-21
generic_ipv4v6_stack: pass ?group to ipv4_of_dhcp

reynir:
Otherwise we get clashing runtime arguments if we have multiple stacks.

Introduced in #1628.


reynir:
The Mirage CI failure is due to unikraft version constraints not going well with OCaml 5.4.

ocaml-ci has the usual 5.4 compiler error messages and an opensuse error that sounds similar to https://github.com/ocaml/infrastructure/issues/184

Otherwise I think CI is good.

PR #1636, opened 2026-04-28
Use utcp

hannesm:
This adds an optional bool argument whether to use uTCP implementation or the mirage-tcpip TCP implementation.

On another thought, the question is whether to make the value by default true? Any opinion against making it the default?

### DIFF
--- a/lib/devices/key.ml
+++ b/lib/devices/key.ml
@@ -135,17 +135,24 @@ let block ?group () =
 
 (** {3 Stack keys} *)
 
-let dhcp ?group () =
-  let default = true in
-  let doc = Fmt.str "Enable dhcp for %a (default %B)." pp_group group default in
-  configure_key ~doc ?group ~default Cmdliner.Arg.bool "dhcp"
+let configure_flag ?(group = "") ~doc name =
+  let prefix = if group = "" then group else group ^ "-" in
+  let doc =
+    Cmdliner.Arg.info ~docs:unikernel_section
+      ~docv:(String.uppercase_ascii name)
+      ~doc
+      [ prefix ^ name ]
+  in
+  let key = Key.Arg.flag doc in
+  Key.create (prefix ^ name) key
+
+let no_dhcp ?group () =
+  let doc = Fmt.str "Enable dhcp for %a." pp_group group in
+  configure_flag ~doc ?group "no-dhcp"
 
 let utcp ?group () =
-  let default = false in
-  let doc =
-    Fmt.str "Enable µTCP stack for %a (default %B)." pp_group group default
-  in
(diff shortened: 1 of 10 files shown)


hannesm:
But what the default should be can be decided in a separate PR.

palainp:
Hello @hannesm, I fully agree with adding this option; it makes integrating uTCP much easier :)

As for the default value, I wonder if `true` is the best choice, as this could increase the size of unikernels using TCP and other L2/L3 protocols (which still require mirage-tcpip). But perhaps I’m wrong, and I’m certainly not entirely against true as default value; I’ll just need to check and eventually add a use_utcp=false, for example, for qubes-mirage-firewall ;)

hannesm:
@palainp thanks for your comment.

From how I understand OCaml, since qubes-mirage-firewall doesn't use TCP, and there's no dependency on tcpip.tcp (the ocamlfind sublibrary), the TCP stack isn't linked into the binary (we should actually test whether this is true ;)

Now, the possible change to use utcp by default shouldn't affect the qubes-mirage-firewall at all. I'll test whether this is true kater when I'm at a computer.

hannesm:
@palainp so setting `use_utcp` to false or true doesn't impact the binary of qubes-mirage-firewall. It indeed has a strict dependency on `tcpip.tcp` due to mirage-nat using that code (it looks like only the packet parsing and constructing -- so maybe this could be in a separate library and mirage-nat would be smaller).

As mentioned earlier, this PR does not propose to use utcp by default.

palainp (approved):
@hannesm thank you for your time and investigation. LGTM :)

Regarding CI: Ocaml 5.3 is not supported anymore with latest ocaml-solo5, so hvt is expected to fail, not sure how to fix the qemu issue for Ocaml 5.4 (monorepo is unhappy :/ and it doesn't seem related to the PR), there are ocaml-ci issues with all 5.4 tests, it doesn't seem related to this PR

hannesm:
I pushed another commit to remove `?tcp` from the stacks: now we can use `?use_utcp` directly, and do not need the ability to pass a tcp stack around via `?tcp`.

hannesm:
I did some cleanups:
- (a) we can use "flag" for DHCP/utcp :) (now, since flags are by default false, there's a `--utcp` and a `--no-dhcp` flag)
- (b) we do not pass the `dhcp_key` nor the `net_key` through the functions, but we just construct it

So we end up with a slightly easier API. I'm happy to hear whether this is welcome or not. It may break existing unikernels, please let me know if your unikernel breaks and we'll find a solution.

hannesm:
I also verified with the dns-secondary unikernel that the flags indeed work as expected and advertised.

PR #1638, opened 2026-05-07
Mirage_runtime_network: use "-" instead of "_" in boot arguments

hannesm:
dns-servers (was dns_servers),
dns-timeout (was dns_timeout),
dns-cache-size (was dns_cache_size),
he-aaaa-timeout (was he_aaaa_timeout),
he-connect-delay (was he_connect_delay),
he-connect-timeout (was he_connect_timeout),
he-resolve-timeout (was he_resolve_timeout),
he-resolve-retries (was he_resolve_retries),
he-timer-interval (was he_timer_interval)

//cc @reynir 

Since this is breaking, I also removed the deprecated Mirage_runtime.register binding.

### DIFF
--- a/lib_runtime/mirage_runtime.ml
+++ b/lib_runtime/mirage_runtime.ml
@@ -182,7 +182,6 @@ let with_argv =
       ]
 
 let runtime_args = Functoria_runtime.runtime_args
-let register = Functoria_runtime.register_arg
 let register_arg = Functoria_runtime.register_arg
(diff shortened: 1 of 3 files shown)


reynir (approved):
I do use `--dns_servers` and `--dns_timeout` for at least one unikernel. While it would require me to change the boot arguments it should fail early and I think even suggest "did you mean --dns-servers". Thanks

PR #1639, opened 2026-05-08
doc: --no-dhcp *disables* dhcp

reynir:
The docstring for `--no-dhcp` was kept from the `--dhcp=...` option in #1636 even though the semantics are inverted.


hannesm:
thanks

PR #1640, opened 2026-06-09
tests/functoria: fix differences between OCaml 4 and 5

edwintorok:
The test was failing on OCaml 5.4 and passing on 4.14:
```diff
-  Hint: Did you mean start?
+  Hint:   Did you mean Unikernel_make__4.start?

-  File "errors/test/main.ml", line 34, characters 3-26:
+  File "errors/test/main.ml", line 34, characters 3-20:
```

Remove `Unikernel_make__4.`, whitespace differences, and line/column numbers from the error message to make the test pass on both.

### DIFF
--- a/test/functoria/e2e/errors.t
+++ b/test/functoria/e2e/errors.t
@@ -3,11 +3,12 @@ of a device's connect function:
 
   $ ./test.exe configure -f errors/in_device.ml
   test.exe: [WARNING] Skipping version check, since our_version ("1.0~test") fails to parse: only digits and . allowed in version
-  $ dune build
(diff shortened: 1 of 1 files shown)


edwintorok:
ocaml-CI tries running this on windows, should probably update `available`, because I don't think that'd be supported

Firobe (approved):

edwintorok:
5.3 failures: I'm told this is an issue with `fstat`.
5.4 failure: this is an issue between a 5.4.0 version constraint and the CI trying to use 5.4.1 (I don't think the CI supports minor versions, it uses a container name ending in 54)

hannesm:
Thanks.

PR #1642, opened 2026-07-02
allow solo5 0.12+

hannesm:


palainp (approved):
Thank you @hannesm ! Even without green CI, I don't see it should fail, so I approve :)

PR #1644, opened 2026-07-07
Avoid RSS memory leaks: call malloc_trim

edwintorok:
Before:
    ```
    $ ulimit -v 440000;OCAMLRUNPARAM=d=1 dist/memory)
    [...]
    2026-07-07T08:28:26+01:00: [INFO] [application] Done
    2026-07-07T08:28:26+01:00: [INFO] [application] Allocating a total of 36372480 words
    2026-07-07T08:28:26+01:00: [INFO] [application] Allocating 141802 * 254 = 36017708 words
    2026-07-07T08:28:26+01:00: [INFO] [application] Done
    2026-07-07T08:28:26+01:00: [INFO] [application] Allocating a total of 36372480 words
    Fatal error: caml_skiplist_insert: out of memory
    
    $ /bin/time dist/memory
    5.34user 0.20system 0:05.55elapsed 99%CPU (0avgtext+0avgdata 583236maxresident)k
    0inputs+0outputs (0major+145230minor)pagefaults 0swaps
    ```
    
    After:
    ```
    time dist/memory
    4.92user 0.94system 0:05.87elapsed 99%CPU (0avgtext+0avgdata 295536maxresident)k
    0inputs+0outputs (0major+446311minor)pagefaults 0swaps
    ```

Draft PR, because I also need to open a PR with the testcase.

### DIFF
--- a/lib_runtime/mirage_runtime.ml
+++ b/lib_runtime/mirage_runtime.ml
@@ -189,7 +189,7 @@ let help_version = Functoria_runtime.help_version
 external malloc_trim : int -> bool = "stub_malloc_trim_noalloc" [@@noalloc]
 
 let trim_on_major_cycle () =
-  let (_ : bool) = malloc_trim (1 lsl 22) in ()
+  let (_ : bool) = malloc_trim (1 lsl 22) in
+  ()
 
-let (_ : Gc.alarm) =
-  Gc.create_alarm trim_on_major_cycle
+let (malloc_trim_alarm : Gc.alarm) = Gc.create_alarm trim_on_major_cycle
--- a/lib_runtime/mirage_runtime.ml
+++ b/lib_runtime/mirage_runtime.ml
@@ -185,3 +185,5 @@ let runtime_args = Functoria_runtime.runtime_args
 let register_arg = Functoria_runtime.register_arg
 let argument_error = Functoria_runtime.argument_error
 let help_version = Functoria_runtime.help_version
+
+external malloc_trim : int -> bool = "stub_malloc_trim_noalloc" [@@noalloc]
--- a/lib_runtime/mirage_runtime.ml
+++ b/lib_runtime/mirage_runtime.ml
@@ -187,3 +187,9 @@ let argument_error = Functoria_runtime.argument_error
 let help_version = Functoria_runtime.help_version
 
 external malloc_trim : int -> bool = "stub_malloc_trim_noalloc" [@@noalloc]
+
+let trim_on_major_cycle () =
+  let (_ : bool) = malloc_trim (1 lsl 22) in ()
+
+let (_ : Gc.alarm) =
+  Gc.create_alarm trim_on_major_cycle
--- a/lib_runtime/dune
+++ b/lib_runtime/dune
@@ -2,6 +2,9 @@
  (name mirage_runtime)
  (public_name mirage-runtime)
  (modules mirage_runtime)
+ (foreign_stubs
+  (language c)
+  (names stub_mirage_runtime))
  (libraries mirage-runtime.functoria lwt logs))
 
 (library
--- a/lib_runtime/mirage_runtime.mli
+++ b/lib_runtime/mirage_runtime.mli
@@ -146,3 +146,5 @@ val runtime_args : unit -> unit Cmdliner.Term.t list
 
 val set_name : string -> unit
 (** Set the name of the unikernel, called at load time for the default name. *)
+
+val malloc_trim: int -> bool
--- a/lib_runtime/mirage_runtime.mli
+++ b/lib_runtime/mirage_runtime.mli
@@ -147,4 +147,5 @@ val runtime_args : unit -> unit Cmdliner.Term.t list
 val set_name : string -> unit
 (** Set the name of the unikernel, called at load time for the default name. *)
 
-val malloc_trim: int -> bool
+val malloc_trim : int -> bool
+val malloc_trim_alarm : Gc.alarm
--- a/lib_runtime/stub_mirage_runtime.c
+++ b/lib_runtime/stub_mirage_runtime.c
@@ -1,0 +1,15 @@
+#define CAML_NAME_SPACE
+#include "caml/mlvalues.h"
+
+#ifdef __GLIBC__
+#include <malloc.h>
+#else
+
+static int malloc_trim(size_t pad) { return 0; }
+#endif
+
+CAMLprim value stub_malloc_trim_noalloc(value const val_pad) {
+  long const pad = Long_val(val_pad);
+  return Val_bool(pad >= 0 && malloc_trim(pad));
+}
+
--- a/test/mirage/memory/dune
+++ b/test/mirage/memory/dune
@@ -1,0 +1,7 @@
+(test
+ (name test_malloc_trim)
+ ; malloc_trim is glibc specific, so restrict to Linux
+ ; /proc/self/smaps_rollup is Linux specific too
+ (enabled_if
+  (= %{system} linux))
+ (libraries mirage-runtime))
--- a/test/mirage/memory/test_malloc_trim.ml
+++ b/test/mirage/memory/test_malloc_trim.ml
@@ -1,0 +1,94 @@
+let find_rss str =
+  try Scanf.sscanf str "Rss: %d kB" Option.some
+  with Scanf.Scan_failure _ -> None
+
+let find_rss_slow ch =
+  let (_skip_first_line : string) = input_line ch in
+  Seq.of_dispenser (fun () -> In_channel.input_line ch) |> Seq.find_map find_rss
+
+(** [get_rss_slow ()] returns the Resident Set Size of the current process.
+
+    See [proc_pid_statm(5)]: the RSS values reported in [/proc/self/stat],
+    [/proc/self/statm], and [/proc/self/status] are inaccurate. Accurate values
+    are in [/proc/self/smaps] and [/proc/self/smaps_rollup] although these are
+    slower to query.
+
+    This function is used in unit tests, so accuracy is preferred.
+
+    There is no POSIX function to retrieve current RSS usage, [getrusage(2)] can
+    only return MaxRSS. *)
+let get_rss_slow () =
+  Gc.compact ();
+  let r =
+    In_channel.with_open_text "/proc/self/smaps_rollup" find_rss_slow
+    |> Option.get
+  in
+  Gc.compact ();
+  r
+
+let n = 1 lsl 14
+let m = 509 (* 509 + 3 words overhead for large arrays -> 512*8 = 4096 *)
+
+let test () =
+  try
+    let rss0 = get_rss_slow () in
+    let a = Array.make 4 [||] in
+
+    a.(0) <- Array.make_matrix n m 0;
+    a.(1) <- Array.make_matrix 1 m 0;
+    a.(2) <- Array.make_matrix n m 0;
+    a.(3) <- Array.make_matrix 1 m 0;
+
+    (* RSS should've increase here *)
+    let rss1 = get_rss_slow () in
+
+    (* Now free the middle values.
+         Do not free the last value, that'd get immediately returned by glibc.
+       *)
+    a.(0) <- [||];
+    a.(2) <- [||];
+
+    (* This frees the values from the OCaml side,
+         but glibc might retain them *)
+    Gc.compact ();
+
+    let rss2 = get_rss_slow () in
+
+    (* keep it alive across the [Gc.compact] *)
+    let _alive = Sys.opaque_identity a in
+    Gc.compact ();
+
+    let rss3 = get_rss_slow () in
+    Printf.printf "RSS0: %d kB\n" rss0;
+    Printf.printf "RSS1: %d kB\n" rss1;
+    Printf.printf "RSS2: %d kB\n" rss2;
+    Printf.printf "RSS3: %d kB\n" rss3;
+    assert (rss0 < rss1);
+
+    (* not accounting for OCaml value overhead the 2 arrays
+         are at least this many bytes
+       *)
+    let expected_kb = 2 * n * m * Sys.word_size / 8 / 1024 in
+    Printf.printf "Expected decrease: %d kB\n" expected_kb;
+    let actual_kb = rss1 - rss2 in
+    Printf.printf "Actual decrease RSS1-RSS2: %d kB\n" actual_kb;
+    if actual_kb < expected_kb then
+      failwith "OCaml released memory, but libc didn't"
+  with Sys_error e ->
+    (* maybe the kernel doesn't have support compiled in *)
+    Printf.eprintf "SKIP: %s\n" e
+
+let () =
+  (* ensure module is linked and initializers run *)
+  Mirage_runtime.set_name "test_malloc_trim";
+  test ();
+  test ();
+  Printf.printf "Disabling malloc_trim workaround\n";
+  Gc.delete_alarm Mirage_runtime.malloc_trim_alarm;
+  try
+    test ();
+    (* the behaviour of glibc can change over time, so don't fail if it works
+       without malloc_trim
+     *)
+    Printf.printf "Succeeded even without malloc_trim"
+  with Failure e -> Printf.eprintf "Ignoring expected failure: %s\n" e


lib_runtime/mirage_runtime.ml
```diff
@@ -185,3 +185,11 @@ let runtime_args = Functoria_runtime.runtime_args
 let register_arg = Functoria_runtime.register_arg
 let argument_error = Functoria_runtime.argument_error
 let help_version = Functoria_runtime.help_version
+
+external malloc_trim : int -> bool = "stub_malloc_trim_noalloc" [@@noalloc]
+
+let trim_on_major_cycle () =
+  let (_ : bool) = malloc_trim (1 lsl 22) in ()
```
dinosaure: I'm not sure to understand the value `1 lsl 22`, I did the same about `mkernel` here: https://github.com/robur-coop/mkernel/blob/be761d563be06a41b9ace7e54705f5fe53a8f727/lib/mkernel.solo5.ml#L687 but with `0`.

edwintorok: I'm a bit undecided whether to use 0 or not. Leaving some memory untrimmed may benefit performance (have to measure it), and may also help with some out of memory crashes (it prevents the memory from being grabbed by `mmap` in the small value pool allocator).
Finalizers need some extra memory during garbage collection (worst case: 3 words for every 2 words in the minor heap),  which would be about 3MiB.

4MiB would match `HUGE_PAGE_SIZE` in OCaml itself, and interestingly it is also the value chosen by qemu: https://github.com/qemu/qemu/commit/5a22ab71623c0fb709d49df353bdf2ec7c445c4c

edwintorok: If we ever upstream this to the OCaml runtime itself then I think:
* non-zero malloc_trim at the end of a major cycle (exact value TBD)
* 0 malloc_trim in Gc.compact

But I can't tell the difference between a compact and a regular cycle in the Gc alarm.

dinosaure: yes, I'm really not against this PR (I did myself such things) but it's important to contextualize such change. I think @palainp can also have an opinion on this because he did some previous works about allocation and `qubes-mirage-firewall`.

dinosaure: Also, if you are interested by metrics, you can also try with [mnet/unikernels/echo](https://github.com/robur-coop/mnet/tree/main/unikernels/echo) which is a basic echo server. I already planned (only in my head) to put a real pressure on it and see how the GC reacts.

edwintorok: > Also, if you are interested by metrics, you can also try with [mnet/unikernels/echo](https://github.com/robur-coop/mnet/tree/main/unikernels/echo) which is a basic echo server. I already planned (only in my head) to put a real pressure on it and see how the GC reacts.

Good point, these tests and improvements are probably needed by both `mirage` and `mnet`, and would be good if we didn't have to duplicate code and fixes.

I have more GC stress tests and fixes coming up, currently in a `mirage-runtime.memory` subpackage, but I think I can make it independent of `mirage` and `Lwt`. They are in general useful improvements to have if you run OCaml in a memory constrained environment (not just unikernels, but perhaps containers too).

Although eventually I'd like to trickle through the improvements into the OCaml runtime, meanwhile a separate package that `mirage` and `mnet` could share might be useful.

To make reviewing and testing easier I'll open draft  PRs for `mirage` with code in a subdir, but then instead of merging it to `mirage` I could release a new opam package.

What do you think?

dinosaure: My view is that the "shared" code shouldn’t be all that important, in my opinion, and it shouldn't warrant a separate package in its own right. If we can also upstream changes to OCaml, that would already be beneficial for Mirage and mnet.

My main point is to share the expertise on this issue rather than centralising GC tuning for unikernels in a single small package: above all, this ensures that everyone can benefit from this expertise and adapt it to their own context.

I would therefore be more inclined to let you experiment on Mirage, whilst ensuring that I can follow your work and subsequently integrate it into mkernel/mnet.

Finally, the proposal regarding the echo server is primarily intended to provide you with a fairly simple, ready-to-use unikernel that incorporates a full TCP/IP stack (including utcp), which would fundamentally benefit the entire ecosystem if we could derive useful metrics from it.

palainp:
Thank you @edwintorok for that work, and thank you @dinosaure for the ping.
When I was investigating the OOM failures in qubes-mirage-firewall, I came to the conclusion that unikernels targeting Xen (and this probably applies to spt and hvt as well), as they obtain all their memory at once, don't need to use `malloc_trim()` and that it's more important to limit the usage of Cstruct in favor of Bytes/String so that the OCaml runtime can manage most of the memory (and I recognise that this may pose a problem for large arrays such as network packets, but that’s another story ;) ).

On the other hand I agree that it may be useful to use `malloc_trim()` for the Unix target.

I just have one question regarding the `#ifdef _GLIBC_` directive: I’m wondering whether the static definition of `malloc_trim()` is mandatory, as it's defined in ocaml-solo5 (https://github.com/mirage/ocaml-solo5/blob/3ec68685056ca9be2d480efa30010c2bbb24e5ac/nolibc/dlmalloc.i#L5384) but perhaps it’s necessary for the unikraft target?

edwintorok:
> Thank you @edwintorok for that work, and thank you @dinosaure for the ping. When I was investigating the OOM failures in qubes-mirage-firewall, I came to the conclusion that unikernels targeting Xen (and this probably applies to spt and hvt as well), as they obtain all their memory at once, don't need to use `malloc_trim()` and that it's more important to limit the usage of Cstruct in favor of Bytes/String so that the OCaml runtime can manage most of the memory (and I recognise that this may pose a problem for large arrays such as network packets, but that’s another story ;) ).
> 
> On the other hand I agree that it may be useful to use `malloc_trim()` for the Unix target.
> 
> I just have one question regarding the `#ifdef _GLIBC_` directive: I’m wondering whether the static definition of `malloc_trim()` is mandatory, as it's defined in ocaml-solo5 (https://github.com/mirage/ocaml-solo5/blob/3ec68685056ca9be2d480efa30010c2bbb24e5ac/nolibc/dlmalloc.i#L5384) but perhaps it’s necessary for the unikraft target?

The `#ifdef` is needed for FreeBSD. Linux isn't the only Unix target.

palainp:
> 
> The `#ifdef` is needed for FreeBSD. Linux isn't the only Unix target.

Yes, you're right, sorry for the noise :)

edwintorok:
I found a simpler way to test this (initially it was a unikernel in `mirage-skeleton`). 
The fix is Linux-specific (well `glibc` specific, but AFAICT the only platform supported by Mirage that uses `glibc` is Linux), so the test uses Linux-specific `/proc/self/smaps_rollup` to read the process's *current* RSS usage.
After a call to `Gc.compact ()` this is expected to decrease by at least the amount freed on the OCaml side
(note: this is not MaxRSS, since that wouldn't decrease).

The test runs 3 times, first 2 times with the malloc_trim workaround, where we can see that RSS2 < RSS1 after the call to compact.
The 3rd time it disables calls to `malloc_trim`, and here we can see that RSS2 doesn't decrease (depends on the exact values used in the test, sometimes it does, but only a little).

```
    RSS0: 6196 kB
    RSS1: 138052 kB
    RSS2: 6220 kB
    RSS3: 6200 kB
    Expected decrease: 130304 kB
    Actual decrease RSS1-RSS2: 131832 kB
    RSS0: 6456 kB
    RSS1: 138304 kB
    RSS2: 6476 kB
    RSS3: 6456 kB
    Expected decrease: 130304 kB
    Actual decrease RSS1-RSS2: 131828 kB
    Disabling malloc_trim workaround
    RSS0: 6456 kB
    RSS1: 138304 kB
    RSS2: 138304 kB
    RSS3: 6580 kB
    Expected decrease: 130304 kB
    Actual decrease RSS1-RSS2: 0 kB
    Ignoring expected failure: OCaml released memory, but libc didn't
```

This actually looks like a `glibc` bug, maybe https://sourceware.org/bugzilla/show_bug.cgi?id=15321 (opened 2013, patch available since 2018), but `glibc` has been long known to have memory usage issues (in another project I switched to alternative allocators like `jemalloc` due to that reason), so having the workaround would be useful.

dinosaure:
> The fix is Linux-specific

I’m not entirely sure I follow. As I understand it, what we’ve observed (though it’s difficult to say for certain at this stage) is that the GC behaves in a way that leads directly to an `Out_of_memory` error in the context of Solo5/hvt and Solo5/xen back in the days of OCaml 4.

One idea suggested by @palainp was precisely to use `malloc_trim` from `dlmalloc` so that we could "return" a contiguous block of memory from our heap to "the OS". Of course, in the case of Solo5/hvt and Solo5/xen, it wasn't really a matter of "returning" it, but this observation came from the fact that the GC was essentially just expanding the heap and was struggling (in the context of `qubes-mirage-firewall`) to reuse free memory blocks: it was here that @palainp concluded (on a hunch) that perhaps OCaml's combination of `dlmalloc` and the GC wasn't all that efficient (which ties in with your idea of using `jemalloc`).

Subsequently, with OCaml 5, where compaction is no longer automatic, the problem seems to have become more pronounced, although @palainp is currently using `malloc_trim` on qubes-mirage-firewall (see this [line](https://github.com/mirage/qubes-mirage-firewall/blob/a968e5c9c4c524d16a89b7818d8792ac422d9022/memory_pressure.ml#L19)). The problem appears to persist (however, I cannot comment on the metrics).

From my point of view, it is difficult at present to confirm that we have a real problem, to characterise it and to make it reproducible: this is the major difficulty that leads to us being able to say "there seems to be a problem" but being unable to provide any concrete metrics. However, this problem relates to Solo5/{hvt,xen} and dlmalloc, and affects the `unix` target only to a very limited extent; this target is, in itself, (as yet) only necessary for testing unikernels (and not really for deploying them).

What might be interesting is to have a reproducible way of demonstrating that we are encountering an `Out_of_memory` error when, logically, we shouldn’t be (this is where the echo server example could be useful; spamming the server should lead to the unikernel becoming unavailable, but not to an `Out_of_memory` error). Next, we need to identify why we’re getting this `Out_of_memory`, and this is where @palainp and I have considered three possible causes:
1) excessive use of `Cstruct.t`, which puts too much strain on the GC
2) `dlmalloc`, which doesn’t play well with the OCaml GC
3) a possible issue with the OCaml GC itself

I know it’s a lot of work, and I’m personally very interested in this and in sorting out the problem once and for all: I’ve spent far too long watching unikernels crash due to an `Out_of_memory` error. We could also try to organise this work together, and I’m available to do so if you’re interested (I also know that the OCaml team are interested in this issue, so we could look into joining forces!).

PS: It is worth noting that I am moving straight on to observations of real-world applications. @palainp had already created a few mini-unikernels like the one in your example, but they did not allow us to draw a connection between what we observed in those mini-unikernels and a 'real' unikernel such as `qubes-mirage-firewall`.

edwintorok:
> I know it’s a lot of work, and I’m personally very interested in this and in sorting out the problem once and for all: I’ve spent far too long watching unikernels crash due to an `Out_of_memory` error. We could also try to organise this work together, and I’m available to do so if you’re interested (I also know that the OCaml team are interested in this issue, so we could look into joining forces!).

Thanks for the offer. I'll be opening a few more PRs with testcases and fixes, please try to poke holes into the testcase and the fix. We can also collaborate on test unikernels.
 I agree that it is important to fix this, I've actually been looking at finding a solution since I joined Tarides in April, and I think I'm quite close to having a full fix.
Although every time I thought I had a solution and testcases that worked, I kept finding a new way to break it by slightly tweaking the testcase.
(I'll try to write up what I learned in a blogpost when it is all done, next time I'll try to blog as I go along).

Recently some of the testcases that I wrote passed on all the distros in the CI:
https://ocaml.ci.dev/github/edwintorok/mirage/commit/e8d7b93ce155a9b47a2b98a22587e47e4f3ebd93

Previously there was a lot of counter-intuitive behaviour: giving the testcase slightly more memory actually made it crash with Out_of_memory on some distros, I think there were 3 prerequisites to make the test/fixes reliable:

* OCaml 5.5, which fixed some GC corner cases (more on that below)
* the use of `malloc_trim` that makes `Gc.compact()` actually free memory (even if fragmented). Without it, you get the weird behaviour where giving it more memory makes it crash depending on fragmentation
* the observation that probing available memory with `mmap` is the only reliable way. You cannot rely on (platform-specific) ways to query available memory, because this is just a sum of the free memory chunks. Even if you have gigabytes "free", they could all be so fragmented that you have 0 chunks available at the requested size (a network packet).

I'm now trying to simplify the testcases even further, to make them more robust. Although the approach still feels somewhat brittle. Eventually I'd like to upstream this into the OCaml runtime, once we've proven that it works in Mirage.

I think there are 3 components required to avoid Out of memory:
* malloc_trim (this PR), although as you say, this is mostly relevant for unit tests
* avoid crashing with `caml_fatal_error` (which cannot be caught), always raise `Out_of_memory` (which can be caught by OCaml). 
  * Gc.compact always does a minor collection first, and if there is not enough memory for the minor collection to promote values you'll crash even if compact would later be able to free enough memory (it never gets a chance to run
  * Without changes to the OCaml runtime I cannot guarantee this, however in practice the likelihood can be reduced to almost 0: reserve some memory required for the minor GC to complete, and when low on free memory then free this reservation, run a GC compact, and reacquire the reservation (this can fail, although it'll fail outside of a minor GC, in regular OCaml code, and raises `Out_of_memory`)
  * probing for free memory is done using a C binding which does `mmap`/`munmap`. We cannot use `malloc` for probing, because that'd just make the problem worse (if `malloc` succeeds that can actually mean that a later `mmap` could fail, unless we call `malloc_trim()`)
  * this probing function needs to be called periodically, e.g. on every received packet, on every major cycle (or even from statmemprof callbacks, but I'd avoid that). To reduce the performance impact I look at GC counters, and number of received packets, and reserve enough memory to account for calling it less often. Currently this is about 4x minor heap size.
  * this reservation could be done more reliably in the OCaml runtime itself (e.g. reserve an extra minor heap size for each minor heap), it already has some functions to commit/decommit memory for example, so this'd only increase VIRT usage, not RSS.
* limit the memory used by TCP buffers/packets.
  * fragmentation can make this more difficult, see https://sqlite.org/malloc.html#_mathematical_guarantees_against_memory_allocation_failures. Smallest packet is probably 2 words (when IP fragmentation is used), largest is 16 MTU, so according to the formula an ideal allocator would need ~7.8x more memory than the sum of packets to ensure that no allocation fails (which means that we wouldn't be able to allow more than 100/7.8 ~ 12% of memory to be used by TCP buffers). 
  * initially I thought about modifying the TCP stack to account for how many packets it stores exactly in each buffer/cache/etc., but that approach is error-prone and hard to maintain
  * a better approach is to make the bigarray account for that itself, unfortunately the GC doesn't provide any stats about how much custom memory we've allocated. And OCaml finalizers don't work, because they get called as soon as the subarray becomes garbage, even when there are more reference counts holding the C memory. I wrote some C bindings that intergrate with Bigarray's reference counting to keep track of exactly how many packets we have live.
  * once the limit is exceeded then we can drop packets, reduce TCP buffer sizes, call compact, etc.
  * we can also reuse the probing function from the previous step to try to prevent `Out_of_memory`

This of course still doesn't prevent `Out_of_memory` from other parts of the network stack, but at least it prevents the most easy and common attacks (flooding with many connections).

> affects the `unix` target only to a very limited extent; this target is, in itself, (as yet) only necessary for testing unikernels (and not really for deploying them).

I think this fix might also be useful for the `spt` target:
```
$ nm dist/hello.spt|grep malloc_trim
0000000000197940 T malloc_trim
```

> One idea suggested by @palainp was precisely to use `malloc_trim` from `dlmalloc` so that we could "return" a contiguous block of memory from our heap to "the OS".

I haven't been able to reproduce the problems/improvements from this particular PR on the Unix backend using the `dlmalloc` from the freestanding implementation in Solo5.
I think the reason is that on Solo5 `mmap` is implemented by calling `posix_memalign`, which calls `malloc` (eventually), so everything goes through the same allocator.
On Unix memory that is held on to by `malloc` is not available to `mmap` (which leads to some interesting failures in the minor GC, where it crashes with `caml_fatal_error`).

Also because there is no real mmap, `malloc_trim` on `dlmalloc` is not quite as effective as on Linux: it can only free memory from the contiguous region at the top, and it cannot free  the fragmented free regions inside (without mmap it doesn't have any other system call available that could do that):
See https://github.com/mirage/ocaml-solo5/blob/main/nolibc/dlmalloc.i#L4333.

Having said that  calling `malloc_trim` on Solo5 or Unikraft could still be useful for releasing memory from the top, just need to find a testcase that clearly demonstrates it has an effect, and then I can add the appropriate ifdefs to enable `malloc_trim` there too. I think for it to have an effect there'd have to be something else in Solo5/Unikraft that'd try to allocate memory in another way than calling `malloc`.


> it's more important to limit the usage of Cstruct in favor of Bytes/String so that the OCaml runtime can manage most of the memory (and I recognise that this may pose a problem for large arrays such as network packets, but that’s another story ;) ).

Yes, limiting memory usage is also needed for fixing https://github.com/mirage/mirage-tcpip/issues/534

> , with OCaml 5, where compaction is no longer automatic, the problem seems to have become more pronounce

Do you only see `Out_of_memory`, or do you also see  `Fatal error: `? (I usually see a mix of the two)

There were also a lot of bugs in earlier OCaml 5.x versions in how the GC kept track of bigarrays.
I was writing some testcases, and testing various workarounds/fixes, but they remained quite brittle until OCaml 5.5.
See this testcase which passes on 5.5, but not 5.4: https://github.com/ocaml/ocaml/pull/14818. Earlier 5.x versions had a bug in how they accounted for the amount of work needed for sweeping the small sizeclass pools, resulting in an ~90x total memory to live memory ratio (i.e. only 1/90th of the memory was live, the rest was garbage, but the GC didn't do enough incremental work to keep up).

> The problem appears to persist (however, I cannot comment on the metrics).

Which OCaml versions were you using?

> What might be interesting is to have a reproducible way of demonstrating that we are encountering an Out_of_memory error when, logically, we shouldn’t be (this is where the echo server example could be useful; spamming the server should lead to the unikernel becoming unavailable, but not to an Out_of_memory error).

Yes, I'll eventually get to the echo server (if all goes well I should get to it next week). I'm trying to take this step by step, one layer at a time.
FIrst I was trying to stabilize protecting against an ICMP flood (i.e. the ethernet and IP layers).

> excessive use of Cstruct.t, which puts too much strain on the GC

There are some very promising improvements in the handling of off-heap allocations wrt to GC pacing by Damien: https://github.com/ocaml/ocaml/pull/14796#issuecomment-4924501898

Using bytes instead of Cstruct might avoid some of that, however it'd also introduce its own challenges:
* with custom values you have the reference counting,  near 0-cost subarray views, etc.
* with Bytes you lose that, and in particular it'll be harder to track exactly how much memory the pending network packets in buffers/caches are using

One possible solution here would be to have a bytes implementation of cstruct, which tracks the amount of memory used itself (no C stubs needed for that, can be done with OCaml finalizers).

> dlmalloc, which doesn’t play well with the OCaml GC

Our `dlmalloc` might be quite old, although appears to be the latest (https://gee.cs.oswego.edu/pub/misc/malloc.c), the only change is CC0->MIT-0 license.
I think `glibc` and others that use an allocator "based on dlmalloc" might've tweaked/fixed things without necessarily upstreaming back the changes.

> a possible issue with the OCaml GC itself

There used to be definitely plenty of those, although my "probing" approach above should mean that even if the GC is not entirely accurate we should be able to detect an increase in memory use and call GC.compact in time.
This doesn't rely on any of the GC's internal counters to be accurate (other than the amount of allocated words increasing).
In general (see https://github.com/damiendoligez/gc-pacing/blob/main/gc-pacing.pdf) the GC is designed to work in a steady-state (amount allocated = amount deallocated). An attacker can trigger a pattern that is very different from that, and make the GC deviate from the expected space overhead.

The worst case for the GC actually seems to be the kind of windowed processing that TCP does: if its pace doesn't keep up, then you'll have a lot more garbage than expected (and without compaction this won't be returned because the latest TCP buffer/window will hold on to the "top" of the allocation).

edwintorok:
> From my point of view, it is difficult at present to confirm that we have a real problem, to characterise it and to make it reproducible: this is the major difficulty that leads to us being able to say "there seems to be a problem" but being unable to provide any concrete metrics. 

Small update:

I wrote a small tool that can dump the value reference graph and value sizes of the OCaml heap: https://github.com/edwintorok/oheapdump.
It doesn't yet work inside a unikernel, but could be used when running in Unix mode, and tries to avoid allocating memory itself so it might eventually be useful to handle `Out_of_memory` or `caml_fatal_error`.

This doesn't yet assign names to all the values (just the sampled allocation sites), but I have some ideas on how to extend that by patching `caml_modify/caml_initialize` in the compiler to mark parents of sampled blocks in some unused value header bits, and to also capture a callstack. This'll be slower, but updating a map only needs `O(log n)` modifications, updating a list head `O(1)`, so it should be useful to see which values hold the network packets (or other big allocations) alive in the OCaml heap, at least during development.

Also discovered an issue with Statmemprof in the context of Mirage: it doesn't provide callbacks for when the underlying bigarray proxy object is freed (its reference count reaches 0). And the existing Custom callback will just track the lifetime of the toplevel bigarray, which will get GCed very soon, but there'll be a lot of subarrays pointing at it. The subarrays have size 0 from memprof's point of view, so they won't get sampled. So looking at how much memory is alive with statmemprof for a mirage unikernel can be very misleading, since it won't show any long lived custom values...
Fixing in general (for any custom value) is not possible, but should be possible to introduce some special support for bigarrays.
https://github.com/ocaml/ocaml/issues/14938

I'll be on vacation next week for a week, so lets resume the discussion after that.

Meanwhile I'll try to clean up and publish the testcases that I have for the memory usage.

PR #1645, opened 2026-07-13
Disable ocamlformat on code format string

reynir:
This makes it easier to read the code format string which ocamlformat otherwise wants to break at around 80 characters instead of how you'd otherwise format the OCaml code. This unfortunately makes the code hard to read. With this patch it is easier to read.


dinosaure:
You can add `[@@ocamlformat "disable"]` to disable it on some parts.

reynir:
I think that's what I did. Please correct me if I'm wrong.

dinosaure:
Ah yes, I did not see it at the end :+1:.

PR #1647, opened 2026-07-31
only print the 'adding unit argument' hint when configuring

### LINKED ISSUE #1608: `mirage clean` says it adds a unit argument to `start ()`
```shell
mirage-skeleton/applications/crypto$ make clean
mirage clean
adding unit argument to 'start ()' (to delay execution)
```

This is slightly confusing as it deletes files.

MavenRain:
Fixes #1608.

**Problem.**  `mirage clean` prints `adding unit argument to 'start ()' (to delay execution)`, which is confusing since clean only deletes files.  The same message also leaks into `describe`, `query` and `help`.

**Root cause.**  The message comes from `Functoria.DSL.main`, which runs when `config.ml` is evaluated at the top level, before cmdliner has parsed the sub-command.  So it fires for every sub-command even though it only describes something the configure action does.

**Fix.**  Queue the message at DSL-evaluation time (`Misc.defer_message`) and flush the queue as the first output of the configure action (the `Cli.Configure` arm in `lib/functoria/lib.ml`).  Queueing preserves order and multiplicity for configurations that call `main` several times.  The queue is a plain process-local ref;  each sub-command runs in a freshly executed `config.exe`, so messages cannot leak across actions.  The configure output is byte-identical to before:  the three existing `configure.t` goldens are untouched by this PR.

**Why not `Logs.info`?**  That would move the hint to stderr behind the default log level, effectively hiding it during a normal `mirage configure`, and would change the configure output users currently see.  Keeping it on stdout, but only under configure, preserves current behaviour exactly.  Not printing it for `query`/`describe` also keeps machine-consumed output clean.

**Tests.**  New cram tests `test/mirage/job-no-device/{clean,query,describe}.t` check that `configure` prints the hint and that `clean`, `query` and `describe` do not.  (`help` output is not pinned;  a golden for it would mostly test cmdliner's man rendering.)

I'd be happy to add a CHANGES.md entry if you would like one in-branch;  I left it out since entries appear to be written at release time.


MavenRain:
The ocaml-ci red is the `freebsd-15.1-4.14_opam-2.5` lane only;  this commit is green on every other completed lane, including `freebsd-15.1-5.5` and the other 4.14 lanes.  The failure is dune's post-rule target check, in a library this PR does not touch:

```
File "lib/devices/dune", lines 1-4, characters 0-85:
Error: Rule failed to generate the following targets:
- default/lib/mirage/devices/kv.mli
```

`kv.mli` is a checked-in source file, so this is the copy rule's output going missing underneath dune.  `main` hit the same lane with the same error class on efe22424 (Jul 4), with no PR involved:

```
Error: Error trying to read targets after a rule was run:
- default/lib/mirage/devices/devices__Http.cmx: Broken symbolic link
```

This looks like builder-side filesystem flakiness on the FreeBSD workers rather than anything in the diff.  Could someone with CI access hit rebuild on that step?  Happy to rebase to retrigger instead if that is easier.

palainp (approved):
LGTM (but tested on linux, not freebsd), mirage clean does not print the message anymore :)

